PR #5, opened 2019-05-09
Integrating Regression tests into CI

blishko:
The main functionality is running regression tests as part of automatic build.
Other changes include:
- restricting CI to run on master branch only
- throwing and catching exception when division by zero is encountered
- replaced unordered_set with set in Tableau to get consistent behaviour on Linux and Mac
- removed LIA regression tests that took too long to finish
- fixing bug when simplification information was not updated on pop


src/tsolvers/lrasolver/Tableau.h
```diff
@@ -12,6 +12,7 @@
 
 #include <unordered_set>
```
aehyvari: Are these still used somewhere?  Maybe remove them everywhere and steer clear of them to avoid similar problems?

blishko: Yes, you are right, it should not be used anywhere else, so try to remote it if it works.

PR #9, opened 2019-05-21
Reserving more memory in advance in important operation of LRA solver

blishko:
Attached is the comparison of performance on latest SMT-LIB QF_LRA. It is not always better, but overall, there is a tiny performance gain and one more problem solved.
[comparison.pdf](https://github.com/usi-verification-and-security/opensmt/files/3201830/comparison.pdf)

### DIFF
--- a/src/tsolvers/lrasolver/Polynomial.cpp
+++ b/src/tsolvers/lrasolver/Polynomial.cpp
@@ -100,7 +100,7 @@ void
 Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, std::function<void(LVRef)> informAdded,
                   std::function<void(LVRef)> informRemoved) {
     decltype(poly) merged;
-    merged.reserve(std::max(this->poly.size(),other.poly.size()));
+    merged.reserve(this->poly.size() + other.poly.size());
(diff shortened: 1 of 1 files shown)


aehyvari:
> Attached is the comparison of performance on latest SMT-LIB QF_LRA. It is not always better, but overall, there is a tiny performance gain and one more problem solved.
> [comparison.pdf](https://github.com/usi-verification-and-security/opensmt/files/3201830/comparison.pdf)

Why does this change cause random effects in the solver, do you know?  I was expecting it to be an overall improvement.  I guess the order of variables in merge changes, and therefore Simplex chooses now different pivots..

blishko:
> Why does this change cause random effects in the solver, do you know? I was expecting it to be an overall improvement. I guess the order of variables in merge changes, and therefore Simplex chooses now different pivots..

No, the work done by the solver is exactly the same. The performance difference is caused only in different handling of memory. The effect I cannot really explain. I see potential slowdown because at the end of the function we shrink_to_fit in order to keep unnecessarily large amount of memory at the end of the function and if we pre-allocate bigger chunk, it will be more often need to be shrunk. On the other hand, if we do not pre-allocate enough, that the vector will need to grow during the merge and that can cause slowdown.

PR #10, opened 2019-05-23
Small improvement by re-using memory in important loop

blishko:
After pivot operation, tableau representation needs to be updated.
Instead of allocating and deallocating helper vector in each iteration, allocate once at the before the loop and reuse.

See the attached comparison.
[comparison_reuse.pdf](https://github.com/usi-verification-and-security/opensmt/files/3211800/comparison_reuse.pdf)

### DIFF
--- a/src/tsolvers/lrasolver/Polynomial.cpp
+++ b/src/tsolvers/lrasolver/Polynomial.cpp
@@ -145,6 +145,56 @@ Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, std::func
     poly.shrink_to_fit();
 }
 
+void
+Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, std::function<void(LVRef)> informAdded,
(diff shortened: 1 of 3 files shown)


src/tsolvers/lrasolver/Polynomial.cpp
```diff
@@ -145,6 +145,56 @@ Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, std::func
...
+                storage.emplace_back(it->var, it->coeff * coeff);
+                informAdded(it->var);
+            }
+            break;
+        }
+        if (otherIt == otherEnd) {
+            storage.insert(storage.end(), myIt, myEnd);
+            break;
+        }
+        if(cmp(*myIt, *otherIt)) {
```
aehyvari: Add a space in if(cmp(...))

blishko: Good eyes! Fixed!

PR #18, opened 2019-05-31
Docs edit

aehyvari:


blishko:
It looks good. I have one big question and small comments.
Shouldn't the Simplex and Egraph solver be mentioned in more details as the techniques for solving the respective logics? Also, I guess some references to their respective papers could be added.

- T-DPLL - isn't DPLL(T) the standard term? It is this way also in the title of the referenced paper.
- "Farkas" should start with capital F in reference [4]

I will read the rules of SMT-COMP again, if anything else should be added.

aehyvari:
> It looks good. I have one big question and small comments.
> Shouldn't the Simplex and Egraph solver be mentioned in more details as the techniques for solving the respective logics? Also, I guess some references to their respective papers could be added.

I wanted to keep the discussion short and not go to too much details, but I'll add the references!  Did you have a reference for the Yices parent list technique or was it just from the code?

> 
> * T-DPLL - isn't DPLL(T) the standard term? It is this way also in the title of the referenced paper.

The authors use T-DPLL in a later journal version, so I think it's the preferred one.  It could be an idea to cite the journal paper though.. I'll do that.

> * "Farkas" should start with capital F in reference [4]

Ok, I'll fix that!

> 
> I will read the rules of SMT-COMP again, if anything else should be added.

Great, let me know!

blishko:
> I wanted to keep the discussion short and not go to too much details, but I'll add the references! Did you have a reference for the Yices parent list technique or was it just from the code?

That was just from going over their source code. I don't know if they describe it somewhere. I consider it just an implementation technique.

aehyvari:
> > * T-DPLL - isn't DPLL(T) the standard term? It is this way also in the title of the referenced paper.
> 
> The authors use T-DPLL in a later journal version, so I think it's the preferred one. It could be an idea to cite the journal paper though.. I'll do that.

Sorry, the paper already cites the journal article which uses the term DPLL(T).  The term T-DPLL is from the handbook of satisfiability, where the relevant chapter is from the same guys and more recent.  Perhaps the citation is correct this way, but I'd still use the more recent term.

blishko:
> Sorry, the paper already cites the journal article which uses the term DPLL(T). The term T-DPLL is from the handbook of satisfiability, where the relevant chapter is from the same guys and more recent. Perhaps the citation is correct this way, but I'd still use the more recent term.

OK, it was just that I was familiar with DPLL(T), but not with T-DPLL. I learned something new :)

From the rules: 

- "This should include a list of all authors of the system, their present institutional affiliations, and any appropriate acknowledgements" Should we try to add the affiliations, or just ignore this?
- "The programming language(s) and basic SMT solving approach employed should be described" I think the current description is satisfactory, no need to go into more details about EGraph and Simplex.
- "System descriptions should also include a URL for a web site for the submitted tool" Do you think the GitHub link is enough or should the link to group website be added?

mmarescotti:
I noticed you removed "lazy". is it intrinsic with T-DPLL?
From the rules I read: mention the basic SMT solving approach employed.

aehyvari:
> I noticed you removed "lazy". is it intrinsic with T-DPLL? From the rules I read: mention the basic SMT solving approach employed.

Yes, I thought that would be implied in T-DPLL.  What do you think, should it be extended?

mmarescotti:
I am just now double-checking and I found a paper from Roberto Sebastiani about it. It looks like they are equivalent terms, so no need to keep both.

blishko:
For me it looks good and we can submit it.

mmarescotti:
I agree

Issue #19, opened 2019-06-04
Duplicate literals in explanation of a conflict in Egraph solver


blishko:
An explanation of a conflict in Egraph contains duplicate literals.
The problem was introduced in b30133f5.
In method Egraph::assertLit, TWO consecutive calls to assertEq are now possible, but if the first one leads to conflict, the second one should not happen (otherwise, it can introduce additional literals to the explanation, which is undesirable).

aehyvari:
Ok, I guess that needs to be fixed.

blishko:
I think I have a fix. I will create a pull request soon.

aehyvari:
Thank you!

PR #20, opened 2019-06-04
Fixing issue #19

blishko:
Fixed possible assertion of equality to Egraph solver when its state is already unsatisfiable.
Also, the information about the state after assertion is stored as bool, no need for l_bool, since only l_False (conflict) and l_Undef (no conflict) were used.

### DIFF
--- a/src/tsolvers/egraph/Egraph.h
+++ b/src/tsolvers/egraph/Egraph.h
@@ -333,8 +333,8 @@ private:
   // Asserting literals
   //
 public:
-  lbool       addDisequality      ( PtAsgn );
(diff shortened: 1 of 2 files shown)


blishko:
I am also thinking if this
`assertEq(boolTermToERef[logic.mkNot(pt_r)], enode_store.ERef_False, PtAsgn(pt_r, l_True))`
cannot be changed to
`addFalse(logic.mkNot(pt_r))`

But it looks like the meaning would be different, wouldn't it?

aehyvari:
I don't remember now why I wrote it like that, but I think there was a reason related to using the correct not symbol (not the one from Logic but the one from Egraph).

Issue #7, opened 2019-05-15
Wrong answer in UF solver when substitutions are turned off


blishko:
OpenSMT gives wrong answer on the following example script.
When the optimizations are not disabled, the answer is correct.

```
(set-logic QF_UF)
(set-option :do-substitutions 0)

(declare-sort US 0)
(declare-fun f (Bool US) US)
(declare-fun a () US)

(declare-fun b () US)
(declare-fun c () Bool)

(assert (and
        (= a b) (not (= (f c a) (f true b) ) ) (not (= (f c a) (f false b) ) )
))
(check-sat)
```

blishko:
fixed by #25

PR #25, opened 2019-06-13
fixing issue #7

blishko:
fix for issue #7 

Explicitly declaring boolean variables from inside UF terms to the SAT solver.


blishko:
I added a public method addVar and renamed the protected method to addVar_.
I welcome any suggestions for better names.

PR #28, opened 2019-06-17
Changes to the short description for SMT-COMP 2019

aehyvari:
Hi Martin & Matteo,

I made a quick pass on the short documentation which we will send to SMT workshop proceedings.  Have a look please, and let me know if you find something strange!

### DIFF
--- a/docs/abstract.bib
+++ b/docs/abstract.bib
@@ -41,6 +41,9 @@
                2018},
   pages     = {56--75},
   year      = {2018},
+  publisher = {EasyChair},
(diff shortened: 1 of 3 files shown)


blishko:
Is it OK if you use plural ("recent examples", "our own model checkers") but then you give only a single example?

aehyvari:
> Is it OK if you use plural ("recent examples", "our own model checkers") but then you give only a single example?

At least syntactically it is ok.

aehyvari:
> Here some suggestions, what do you think?

Looks good, I updated the text.

PR #29, opened 2019-06-27
Small refactoring and cleanup in Logic

blishko:


src/logics/Logic.C
```diff
@@ -95,27 +95,32 @@ Logic::Logic(SMTConfig& c) :
...
+    vec<PTRef> args;
     sort_store.newSort(bool_id, tmp_srefs);
     sort_BOOL = sort_store["Bool"];
+    tmp_srefs.push(sort_BOOL);
+
+    sym_TRUE = newSymb(tk_true, tmp_srefs, &msg);
+    sym_store[sym_TRUE].setNoScoping();
+    sym_store.setInterpreted(sym_TRUE);
 
-    term_TRUE = mkConst(getSort_bool(), tk_true);
```
aehyvari: Why this change?  Would it be better to explicitly call mkConst and mark the symbol a constant instead of mkFun and then marking it a constant (as is done later on line 112)?

blishko: You are right. I can revert this.
The idea behind was that mkConst was calling mkVar which was calling insertTerm and there it would go to mkUninterpFunc, which did not seem right. But then I changed mkVar to call mkFun directly instead of insertTerm, so the problem was solved in a different way, but I kept the original solution.

aehyvari:
Ok, good, thank you!

src/logics/Logic.C
blishko: Reverted.

PR #30, opened 2019-07-31
Added even stronger simplification of interpolants. Refactoring

blishko:
Added an aggressive version of lifting conjunctions and disjunctions. New version ensures that no disjunction is a child of disjunction, and the same of conjunctions.
Added a more aggressive interpolation simplification based on this.
Refactoring.

### DIFF
--- a/src/api/MainSolver.C
+++ b/src/api/MainSolver.C
@@ -129,9 +129,7 @@ sstat MainSolver::simplifyFormulas(char** err_msg)
             // Optimize the dag for cnfization
             if (logic.isBooleanOperator(fla)) {
                 PTRef old = fla;
-                Map<PTRef,int,PTRefHash> PTRefToIncoming;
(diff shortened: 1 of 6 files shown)


aehyvari:
Looks good! I like the cleaning of the two phases of computing incoming edges and rewriting. So is this the difference between aggressive and classic rewriting in that classic tried to preserve shared structure and aggressive does not?

blishko:
> Looks good! I like the cleaning of the two phases of computing incoming edges and rewriting. So is this the difference between aggressive and classic rewriting in that classic tried to preserve shared structure and aggressive does not?

Yes, "classic" is the original algorithm that leaves the shared nodes untouched, while "aggressive" does not care whether the child with the same connective is shared or not and it will pull child's children up.

Please, do not close the pull request. If you agree, then approve. Then we can merge and close.

aehyvari:
Ah sorry, I was answering from my phone and screwed it up..

PR #32, opened 2019-08-20
Polarity suggestion

blishko:
This adds the possibility for the SAT solver to ask theory solver for suggested polarity once SAT solver has picked new decision variable.

New interface method is provided with implementations for LRA solver and UF solver.
The measured impact in LRA is huge, in UF is small but still positive.

### DIFF
--- a/src/tsolvers/egraph/EgraphSolver.C
+++ b/src/tsolvers/egraph/EgraphSolver.C
@@ -221,6 +221,34 @@ PTRef Egraph::getSuggestion( )
   return PTRef_Undef;
 }
 
+lbool Egraph::getPolaritySuggestion(PTRef p)
+{
+    if (!isInformed(p)) { return l_Undef; }
+    // MB: it could be uninterpreted predicate! No suggestion in that case
+    if (!logic.isEquality(p) && !logic.isDisequality(p)) { return l_Undef; }
+    bool equality = logic.isEquality(p);
+    const Pterm& term = logic.getPterm(p);
+    if(term.size() > 2) { return l_Undef; } // For now focus on 2 arguments
+    PTRef lhs = term[0];
+    PTRef rhs = term[1];
+    assert(enode_store.termToERef.has(lhs) && enode_store.termToERef.has(rhs));
+    if (!enode_store.termToERef.has(lhs) || !enode_store.termToERef.has(rhs)) { return l_Undef; }
+    ERef e_lhs = termToERef(lhs);
+    ERef e_rhs = termToERef(rhs);
+    if (getEnode(e_lhs).getRoot() == getEnode(e_rhs).getRoot()) {
+        // Already in the same equivalence class, avoid conflict
+        return equality ? l_True : l_False;
+    }
+    PtAsgn tmp(PTRef_Undef, l_Undef);
+    bool res = unmergeable(getEnode(e_lhs).getRoot(), getEnode(e_rhs).getRoot(), tmp);
+    if (res) {
+        // they are unmergable so don't assert equality
+        return equality ? l_False : l_True;
+    }
+    // no decision
+    return l_Undef;
+}
+
 //
 // Communicate conflict
 //
@@ -1319,7 +1347,7 @@ void Egraph::undoDisequality ( ERef x )
 }
 
 
-bool Egraph::unmergeable (ERef x, ERef y, PtAsgn& r)
+bool Egraph::unmergeable (ERef x, ERef y, PtAsgn& r) const
 {
     assert( x != ERef_Undef );
     assert( y != ERef_Undef );
--- a/src/tsolvers/lrasolver/LRASolver.C
+++ b/src/tsolvers/lrasolver/LRASolver.C
@@ -194,7 +194,7 @@ LRALogic&  LRASolver::getLogic() { return logic; }
 
 lbool LRASolver::getPolaritySuggestion(PTRef ptref) const {
     if (!this->isInformed(ptref)) { return l_Undef; }
-    LVRef var = this->lavarStore.getVarByLeqId(logic.getPterm(ptref).getId());
+    LVRef var = this->getVarForLeq(ptref);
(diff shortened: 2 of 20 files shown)


blishko:
@aehyvari, I created an option in SMTConfig, which enables this feature by default.
Should this option be also accessible from SMT-LIB scripts? Should a string representation be created?

src/tsolvers/lrasolver/LRASolver.C
```diff
@@ -191,6 +191,32 @@ LRASolver::~LRASolver( )
 
 LRALogic&  LRASolver::getLogic() { return logic; }
 
+lbool LRASolver::getPolaritySuggestion(PTRef ptref) const {
+    if (!this->isInformed(ptref)) { return l_Undef; }
+    LVRef var = this->lavarStore.getVarByLeqId(logic.getPterm(ptref).getId());
```
aehyvari: Do we need a method LVRef getVarByPTRef(PTRef tr) const { return getVarByLeqId(logic.getPterm(tr).getId())); }  ?

src/tsolvers/lrasolver/LRASolver.C
```diff
@@ -191,6 +191,32 @@ LRASolver::~LRASolver( )
...
+        positive = true;
+    }
+    bool negative = false;
+    auto const& negative_bound = this->boundStore[bounds.neg];
+    if ((negative_bound.getType() == bound_l && negative_bound.getValue() <= val)
+        || (negative_bound.getType() == bound_u && negative_bound.getValue() >= val)) {
+        negative = true;
+    }
+    // MB: Maybe we will not get any suggestion, if the value is <0|-1/2>, but the bounds are <0|0> and <0|-1>
+//    assert(positive || negative);
```
aehyvari: What's the invariant here?  Can assert(positive || negative) be removed?

src/tsolvers/egraph/EgraphSolver.C
```diff
@@ -221,6 +221,34 @@ PTRef Egraph::getSuggestion( )
...
+    // MB: it could be uninterpreted predicate! No suggestion in that case
+    if (!logic.isEquality(p) && !logic.isDisequality(p)) { return l_Undef; }
+    bool equality = logic.isEquality(p);
+    const Pterm& term = logic.getPterm(p);
+    if(term.size() > 2) { return l_Undef; } // For now focus on 2 arguments
+    PTRef lhs = term[0];
+    PTRef rhs = term[1];
+    assert(enode_store.termToERef.has(lhs) && enode_store.termToERef.has(rhs));
+    if (!enode_store.termToERef.has(lhs) || !enode_store.termToERef.has(rhs)) { return l_Undef; }
+    ERef e_lhs = termToERef(lhs);
```
aehyvari: Looks like the code 236 -- 239 is the same as the beginning of Egraph::unmergeable.  I think the same functionality would be available with less code by just calling bool res = unmergeable(e_lhs, e_rhs, tmp);

aehyvari:
Hi Martin, thanks!  I think that your experiments are so convincingly good
that it might just be better to make it not an option :)

src/tsolvers/lrasolver/LRASolver.C
blishko: > Do we need a method LVRef getVarByPTRef(PTRef tr) const { return getVarByLeqId(logic.getPterm(tr).getId())); } ?

It looks like it could be useful. I can try to add it.

src/tsolvers/egraph/EgraphSolver.C
blishko: The change you suggest would result in a different behaviour.
With the current implementation, EGraph suggests to make things equal only if they ARE ALREADY in the same equivalence classes. (And suggests to make them not equal if they are unmergeable).

In your scenario, Egraph would suggest to make things equal if it CANNOT prove that they are unmergeable. So it would make suggestions more often.

My original idea was for the Egraph would only suggest polarity if it KNOWS that the opposite polarity would lead to conflict. But I don't mind making it more aggressive suggestions to make things equal. If you think it's better that way I will change it.

src/tsolvers/lrasolver/LRASolver.C
blishko: Originally I thought that the invariant would be positive XOR negative (exactly one will be true and the other one false).
But then I ran into a situation where a value of a variable, "x", was <0, -1/2> (-1/2 is the delta), and the constraint was x >= 0. The positive bound on this is a lower bound with value <0,0> and the negative bound is an upper bound with value <0, -1> ( x < 0, i.e, x <= 0 - Delta)
In this case the current value does not satisfy neither positive nor negative bound.
It is kind of a special case but I had to comment out the assert because of this since it is a valid situation.
I just wanted to stress that this assertion is not valid (even though I originally thought it would be) so I left it there as a comment.

Do you suggest to remove it completely?

src/tsolvers/egraph/EgraphSolver.C
aehyvari: I think you're right, just keep it the way you first coded it.

src/tsolvers/lrasolver/LRASolver.C
aehyvari: Hm ok... In that case the 'consistent' suggestion would be to set negative polarity, that is, x < <0,-1>?  But it wouldn't really be consistent in the end because x = <0,-1/2> would break x < <0,-1>.   What does our code do in that case?

I'm not sure what's the right thing to do about the assert.  Perhaps removing it and expanding the comment a little as you do above would be better.

blishko: Yes, neither choice is consistent with the assignment, so the theory solver just says "I don't know" and the decision falls back to the SAT solver.

I will remove the commented assert and add more detailed explanation.

blishko:
I have changed the comments in LRASolver::getPolaritySuggestion and refactored the code of translating PTRef to LVRef a little bit.
Check the two new commits.

PR #33, opened 2019-09-06
The lookahead partitioning fixes

aehyvari:
The pull request contains fixes to lookahead partitioning and refactoring local to the general lookahead features.


src/smtsolvers/LookaheadSMTSolver.h
```diff
@@ -239,6 +99,8 @@ class LookaheadSMTSolver : public SimpSMTSolver {
     PathBuildResult setSolverToNode(LANode *n);                                         // Set solver dl stack according to the path from root to n
     laresult expandTree(LANode *n, LANode *c1, LANode *c2);                             // Do lookahead.  On success write the new children to c1 and c2
     void deallocTree(LANode *n);                                                        // Dealloc the tree rooted at n
+    LookaheadScore *score_p;
```
blishko: I believe this pointer is never deleted now and is leaked on destruction of the solver.
How about turning this to a unique_pointer? It would clearly define the ownership and take care of the destruction automatically.

aehyvari: True, thanks!  I committed what I believe to be a fix in https://github.com/usi-verification-and-security/opensmt/pull/33/commits/79b326fdc49829865cf3344922f46a958c0c0a7d.  Let me know if you think this is the right way!

blishko: Looks good!

PR #34, opened 2019-09-19
Separates simplex from the PTRefs and provides it as a more stand-alone class that can be reinstantiated.

aehyvari:


src/tsolvers/lrasolver/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ target_sources(tsolvers
     PRIVATE "${CMAKE_CURRENT_LIST_DIR}/Delta.h"
     PRIVATE "${CMAKE_CURRENT_LIST_DIR}/LABounds.h"
     PRIVATE "${CMAKE_CURRENT_LIST_DIR}/LARefs.h"
+    PRIVATE "${CMAKE_CURRENT_LIST_DIR}/LAVarMapper.h"
```
blishko: Should we move these files regarding LARefs, LAVarMapper, LABounds, etc. to the LASolver folder now?

src/tsolvers/lasolver/Simplex.cpp
```diff
@@ -0,0 +1,490 @@
...
+
+bool Simplex::isUnbounded(LVRef v) const
+{
+    bool rval = model.isUnbounded(v);
+//    if (rval)
+//        printf("Var %s is unbounded\n", lva.printVar(v));
+    return rval;
+}
+
+std::unique_ptr<Polynomial> Simplex::addPoly(std::unique_ptr<std::vector<std::pair<LVRef,opensmt::Real>>> terms) {
```
blishko: This method seems useless.
The vector of terms passed here should never have been created.
Instead the Polynomial should be created directly.

I believe the only important thing is the call to tableau.nonbasicVar
There should be a way to do that directly.

src/tsolvers/lasolver/Simplex.cpp
```diff
@@ -0,0 +1,490 @@
...
+            return val <= bound.getValue();
+        }
+        else {
+            assert(bound.getType() == bound_l);
+            return val >= bound.getValue();
+        }
+    }
+}
+
+bool Simplex::checkSimplex(std::vector<LABoundRef> &explanation, std::vector<opensmt::Real> &explanationCoefficients) {
```
blishko: I don't like these out parameters that may or (more often) may not be used.
I think the simplex should just store the flag about the conflicting row and the solver would explicitly ask for the explanations.

src/tsolvers/lasolver/Simplex.cpp
```diff
@@ -0,0 +1,490 @@
...
+    // nvDiff represents how much we need to change nv, so that bv gets to the right value
+    auto nvDiff = (bvNewVal - model.read(bv)) / coeff;
+    // update nv's value
+    changeValueBy(nv, nvDiff);
+}
+
+//
+// Returns the bounds conflicting with the actual model.
+//
+void Simplex::getConflictingBounds( LVRef x, std::vector<LABoundRef> &explanation, std::vector<opensmt::Real> &explanationCoefficients)
```
blishko: Probably also here, it would be nice to move the out params to return value.
If the function returns something, let it return it through return value.

src/tsolvers/lrasolver/LRAModel.h
```diff
@@ -75,7 +71,23 @@ class LRAModel
 
     void printModelState();
 
-    void clear();
+    virtual void clear();
+};
+
+class TermLRAModel : public LRAModel
```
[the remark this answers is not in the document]
aehyvari: The current logic of the LASolvers keeps track of the decision history in both the asserted bounds needed by the simplex algorithm and the asserted PTRefs needed by the SMT solver.  The model is needed by the simplex's internal logic but this is now made independent of the PTRefs since the simplex queries internal to LIA solver will in general create new bounds that the SMT solver does not need.

The proposed solution makes the PTRef-aware LRAModel a sub-class of the general LRA model.  Another possibility would be to introduce a new wrapper class for LRAModel that would take care of PTRefs. Since much of the logic for LRAModel integrates to the handling of the PTRefs, this seems to be a simpler solution, although possibly a little unorthodox.

src/tsolvers/lrasolver/CMakeLists.txt
aehyvari: I think that would be good.  I was hesitant to do this straight up because of the parallel development on LIA solver, but perhaps this is not an issue in reality.

src/tsolvers/lasolver/Simplex.cpp
aehyvari: I'm not sure if I understand this one. The explanation especially in Simplex should have a limited lifetime that is specific to the query, so having it not live longer than necessary should be good.  The fact that the explanationCoefficients is not used is another issue that should be solvable by having perhaps an interpolating derived class of LASolver.

For the LRASolver I see that the explanation might need to have a longer lifetime, but I think this implementation did not change.

src/tsolvers/lasolver/Simplex.cpp
aehyvari: That might work, also for the above thing.  It would be nice to couple together the sounds and the coefficients, when that makes sense, into a single vector that gets returned.

src/tsolvers/lasolver/Simplex.cpp
aehyvari: That's probably a good idea!

src/tsolvers/lasolver/Simplex.cpp
aehyvari: Perhaps turning the thing into a return value as suggested below would fix the problem.

src/tsolvers/lrasolver/CMakeLists.txt
aehyvari: Done.

src/tsolvers/lasolver/Simplex.cpp
aehyvari: Done.

src/tsolvers/lasolver/Simplex.cpp
aehyvari: Done.

src/tsolvers/lrasolver/LRAModel.h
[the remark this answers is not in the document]
aehyvari: As discussed, I suggest to keep this the way it is for now, and return to the question once it's more clear how the model system should look like.

blishko:
Looks good!
I think the changes are good, let me just quickly run HiFrog tests and then me can merge.

PR #36, opened 2019-10-11
Ownership of LRAModel moved to Simplex

aehyvari:
A bit of refactoring.  Have a look and let me know what you think!


aehyvari:
For now it looks like the Yices-style Bland adjustment makes the results worse.  I didn't expect this.
[osmt2-LRA-2019-09-18_vs_2019-10-15.pdf](https://github.com/usi-verification-and-security/opensmt/files/3732958/osmt2-LRA-2019-09-18_vs_2019-10-15.pdf)

blishko:
I agree with disabling the Yices rule for now.
I would finish this pull request with that rule disabled and investigate further.

aehyvari:
Yes, I committed a version without that, and run the LRA instances on the
cluster with the reverted version to make sure it's not some other
weirdness.  Once that's done and all looks normal, I think we're good to
merge.

aehyvari:
The performance in the pull request is consistent with the current master.  I'll try to figure out why rebasing doesn't work and the do the thing.

[osmt2-LRA-2019-09-18_vs_2019-10-16.pdf](https://github.com/usi-verification-and-security/opensmt/files/3735018/osmt2-LRA-2019-09-18_vs_2019-10-16.pdf)

blishko:
Looking at the graph, it almost looks like you started the branch on top of some stash. It has some non-linear structure for reason I do not understand. 
Probably just do the merge.

PR #38, opened 2019-10-19
Fix for issue #37

### LINKED ISSUE #37: Crash in (get-value) command
OpenSMT crashes when it is asked for a value of a Real variable that was declared but not present in the formula.

Here is the example:

(set-logic QF_LRA)
(declare-fun x () Real)
(declare-fun y () Real)
(assert (>= x 2))
(check-sat)
(get-value (x y))

blishko:
Fixes #37


blishko:
> Hi Martin, looks ok to me! I have the feeling that this (or something similar) was already fixed once. Could you add the smtlib2 file from issue #37 as a regression test?

That's a good idea. I will add it.

Issue #12, opened 2019-05-27
Nondeterminism in simplification code


blishko:
Nondeterminism in the algorithm of breaking substitutions loop can lead to different substitutions being applied which means a different formula is send to the solver.

The nondeterminism is caused by sorting starting nodes of the algorithm by their address in memory.

See method `Logic::breakSubstLoops`

PR #49, opened 2019-11-29
Substitution refactor

aehyvari:
Closes issue #12 .  There is now a new, memory-managed implementation for finding and breaking substitution loops.   The implementation is separated into a class for cleaner code and ease of testing.  Some stuff would still be nice to have:

- Make sure that the worst-case behaviour of the loop breaking is not exponential
- Try to find substitutions that would result in minimum number of terms
- Flatten the substitution structure (currently the substitutions must be done iteratively to reach sinks)

There is no significant impact on run times compared to current master.  The deviations were not reproducible and therefore probably artefacts of cluster's run time environment (see [here for QF_UF](https://github.com/usi-verification-and-security/opensmt/files/3905538/master_vs_substitution-refactor-2019-11-29.pdf) and [here for QF_LRA](https://github.com/usi-verification-and-security/opensmt/files/3905544/master_vs_subst-refactor-2019-11-29.pdf)).


blishko:
Branch has been rebased to exclude commits from other branches, which has already been merged to master.

src/logics/SubstLoopBreaker.h
```diff
@@ -0,0 +1,192 @@
...
+enum class NStatus { inStack, complete, unseen };
+
+template <class T>
+struct Ref {
+    uint32_t x;
+    Ref() : x(INT32_MAX-1) {}
+    Ref(uint32_t i) : x(i) {}
+    inline friend bool operator== (const Ref& a1, const Ref& a2)   { return a1.x == a2.x; }
+    inline friend bool operator!= (const Ref& a1, const Ref& a2)   { return a1.x != a2.x; }
+    inline friend bool operator< (const Ref& a1, const Ref& a2)    { return a1.x > a2.x;  }
```
blishko: This is most probably not a problem, I am just curious, why is the operator flipped?
Why `a1 < a2` iff `a1.x > a2.x`?

src/logics/SubstLoopBreaker.h
```diff
@@ -0,0 +1,192 @@
...
+private:
+
+    const Logic& logic;
+    Map<PTRef,bool,PTRefHash> seen;
+    TargetVarListAllocator tvla;
+    SubstNodeAllocator sna;
+
+public:
+    SubstLoopBreaker(const Logic& l) : logic(l), tvla(1024), sna(tvla, logic, 1024) {}
+    Map<PTRef,PtAsgn,PTRefHash> operator() (const vec<Map<PTRef,PtAsgn,PTRefHash>::Pair*>&& substs);
```
blishko: I don't think the parameter should be `const &&`, const rvalue does not really make sense, since you cannot move from it, if it is const.
The same is true for `constructSubstitutionGraph`
In the implementation I don't think you need to move this parameters anywhere, so I think this should be just const lvalue reference `const &`, and a bunch of `std::move`s can be removed.

src/logics/SubstLoopBreaker.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+//        printGraphAndLoops(startNodes, loops);
+        vec<SNRef> orphans = breakLoops(loops);
+        for (int i = 0; i < orphans.size(); i++)
+            startNodes.push(orphans[i]);
+        sort(startNodes);
+        uniq(startNodes);
+//        printGraphAndLoops(startNodes, loops);
+    }
+    return constructLooplessSubstitution(std::move(substs));
```
blishko: Applying std::move twice to the same variable is very suspicious. If you move from it once, that it is in unspecified state and should not be accessed anymore.
As I write in another comment, I believe this parameter should be just classic `const &` and the moves should be removed.

src/logics/SubstLoopBreaker.h
aehyvari: Hehe I think it was just for fun :)

src/logics/SubstLoopBreaker.h
aehyvari: Hm yes... My idea was to have the map consumed by the function since the whole point of the process is to replace an old map with a new map.  But I could buy the argument that I'm needlessly making the implementation more specialised and it would be better to keep it general.  In any case, also if my initial idea would be good the argument shouldn't be declared const.

src/logics/SubstLoopBreaker.h
blishko: OK :)
It's the same for PTRef.
I wonder what would happen if we would flip it to the more natural implementation.
But maybe it's better not to try :)

src/logics/SubstLoopBreaker.h
blishko: OK, I get the original idea, but I don't think the current implementation is achieving that goal.
I don't know what would be the best way to do that. But I think that in any case `const &&` should be replaced, because it is confusing and the double move from the other comment needs to be addressed.

aehyvari:
Yes. I had an idea on the evening but didn't have time to work it out yet.
I'll do something today and let's discuss if it got better.

src/logics/SubstLoopBreaker.h
aehyvari: This issue should now have been addressed.

src/logics/SubstLoopBreaker.cpp
aehyvari: This is fixed now.

src/logics/SubstLoopBreaker.cpp
```diff
@@ -248,7 +250,7 @@ Map<PTRef,PtAsgn,PTRefHash> SubstLoopBreaker::operator() (const vec<Map<PTRef,Pt
         uniq(startNodes);
 //        printGraphAndLoops(startNodes, loops);
     }
-    return constructLooplessSubstitution(std::move(substs));
+    return constructLooplessSubstitution(std::move(substs)); // Why don't we need the move contructor here, given that Map is a template?
```
blishko: Why do you think being or not being a template makes a difference?

src/minisat/mtl/Vec.h
```diff
@@ -42,8 +42,8 @@ class vec {
     int cap;
 
     // Don't allow copying (error prone):
-    vec<T>&  operator = (vec<T>& other) { assert(0); return *this; }
-             vec        (vec<T>& other) { assert(0); }
+    vec<T>&  operator = (vec<T>& other)  = delete;
```
blishko: Align with the next line :) (Extra space?)

blishko:
I am fine with merging this.
I think this time we need to do a proper merge, given that in the meantime we added #53 to the master.

src/logics/SubstLoopBreaker.cpp
aehyvari: I forgot that comment there, I'll remove it.  I actually later figured out that when map comes from that function it already has been cast to the movable type by std::move.

But I was expecting the copy elusion to work in that function (which it does not) without std::move.  I figured out that for copy elusion there needs to be a non-template constructor, and I'm now thinking that that's the reason.

src/minisat/mtl/Vec.h
aehyvari: Done :)

Issue #45, opened 2019-11-18
Ghost literals


aehyvari:
An SMT formula $f$ in CNF is satisfiable if a theory-consistent partial truth assignment $p$ satisfies all clauses of $f$ (Prop. 2.2 of [1]).  However, OpenSMT currently determines that an instance satisfiable only after finding a complete theory-consistent assignment, causing at least in theory extra work for finding satisfying assignments for the rest of the variables.

Given a (propositional) truth assignment $p$ and an SMT formula $f$ in CNF, if a literal $l$ appears only in clauses $c \in f$ that are satisfied by $p$, $l$ is called a ghost literal.  Ghost literals have the property that they need not to be branched on to determine the satisfiability with respect to Prop. 2.2 above.

This issue studies ways of implementing ghost literal detection for opensmt and whether they would provide a performance boost for (in particular) satisfiable instances.

[1] Roberto Sebastiani: [Lazy Satisability Modulo Theories](https://dblp.org/rec/html/journals/jsat/Sebastiani07). JSAT 3(3-4): 141-224 (2007)

aehyvari:
The results on the initial implementation with a naive method of determining the satisfiability of a clause are inconclusive.  To me it looks like there is a small positive effect when measuring the number of decisions.  I ran some tests for QF_LRA:

- run time comparison [noghost_vs_ghost-swap.pdf](https://github.com/usi-verification-and-security/opensmt/files/3938847/noghost_vs_ghost-swap.pdf)
- decision comparison [noghost_vs_ghost-swap_decs.pdf](https://github.com/usi-verification-and-security/opensmt/files/3938857/noghost_vs_ghost-swap_decs.pdf)
- average decision time comparison [noghost_vs_ghost-swap_avg_dec_time.pdf](https://github.com/usi-verification-and-security/opensmt/files/3938862/noghost_vs_ghost-swap_avg_dec_time.pdf)

Note that there's a resolution problem with the avg dec time comparison and one dot corresponds therefore to many instances.  For a proper analysis we'd have to go and see the "weight" of the dots.

I would try to see if a better clause satisfiability check will help with the run time, and otherwise keep this code disabled for now at least for QF_LRA, possibly enabling it for presumably more costly logics such as QF_LIA.  QF_UF needs still to be tested.

blishko:
The initial version has been implemented in #50

PR #50, opened 2019-12-02
Ghost vars

aehyvari:
This addresses the issue #45 initially, by implementing a new branching function that is sensitive to variable "ghostliness". The new code is implemented in a class derived from SimpSMTSolver. Some adjustment to the rest of the code was necessary to allow overriding functions, but by default the features are disabled and the execution is equal to previous master.

At the moment the results are not unambiguously positive, but checking the satisfiability of a clause is implemented naively and it could be possible to obtain speedup if this part is implemented more efficiently.


src/smtsolvers/GhostSMTSolver.cpp
```diff
@@ -0,0 +1,250 @@
...
+void
+GhostSMTSolver::detachClause(CRef cr, bool strict)
+{
+    SimpSMTSolver::detachClause(cr, strict);
+    const Clause& c = ca[cr];
+    if (c.learnt())
+        return;
+
+    assert(false);
+    for (int i = 0; i < c.size(); i++) {
```
blishko: Why do we need code after assert(false)?
If it is not an error than there should not be assert(false), if it is error than probably something is wrong with the solver and we should just abort.

aehyvari:
I'd have some somewhere on the disk but not with me now... I'd like to have
the next pull request in there before working more on this one actually.
I'll check the one you noticed with the assert false asap!

src/smtsolvers/GhostSMTSolver.cpp
aehyvari: True, I removed the code, thanks!

PR #54, opened 2019-12-27
Cleanup and some optimizations of LRA solver

blishko:
This merge contains a cleanup of Polynomial::merge method and 2 optimizations:
1. Precise tracking of basic variables that are out of bounds - this prevents unnecessary checks when looking for suitable candidate for pivoting.
2. Implementation detail in data structure representing Tableau's column - when removing row from a column, it seems better to search for the position of the row from the end of the column rather than from the beginning. Possible explanation in the commit's message.


aehyvari:
Did the optimisation affect the runtime?

blishko:
> Did the optimisation affect the runtime?

I believe so, the attached comparison should be capturing the changes in this PR. It solves some additional instances and the overall runtime also improves.
[comparison_linear_PR_54.pdf](https://github.com/usi-verification-and-security/opensmt/files/4019090/comparison_linear_PR_54.pdf)

[comparison_PR_54.pdf](https://github.com/usi-verification-and-security/opensmt/files/4019093/comparison_PR_54.pdf)

PR #58, opened 2020-02-10
Update in LRA decomposition

blishko:
This PR contains the implementation of complete algorithm for computing decomposition of the vector of Farkas coefficients based on the basis of the kernel of the matrix of coefficients of A-local  variables.
Additionaly, it contains 2 hotfixes: Increasing the version of BISON required for compilation and fixing an incorrect use of Pterm& instead of PTRef.
I also modified the code of SimplifyConst::simplify a little bit to avoid unnecessary copying of vector of arguments.

### DIFF
--- a/src/logics/LALogic.cc
+++ b/src/logics/LALogic.cc
@@ -837,13 +837,15 @@ void SimplifyConstTimes::constSimplify(const SymRef& s, const vec<PTRef>& terms,
         terms_new.push(con);
     else
     {
-        Pterm& p = l.getPterm(plus);
         vec<PTRef> sum_args;
-        for(int i = 0; i < p.size(); ++i)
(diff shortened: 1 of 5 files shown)


src/logics/LALogic.cc
```diff
@@ -837,13 +841,15 @@ void SimplifyConstTimes::constSimplify(const SymRef& s, const vec<PTRef>& terms,
         terms_new.push(con);
     else
     {
-        Pterm& p = l.getPterm(plus);
```
aehyvari: Maybe make explicit that here con != PTRef_Undef && plus != PTRef_Undef with a comment, like moving the above comment "distribute the constant over the first sum" here?

blishko: Added assertion to make it explicit. Also, moved the comment.

Issue #59, opened 2020-03-27
Assertion violation at /src/cnfizers/Cnfizer.cc:523


rainoftime:
Hi, for this formula
~~~~
(set-logic QF_LRA)
(declare-fun _substvar_29_ () Bool)
(check-sat)
(check-sat)
(assert (or (distinct true true) _substvar_29_))
(check-sat)
~~~~
opensmt (commit e9e14b4bc1cae) throws an assertion violation
~~~~
/opensmt/src/cnfizers/Cnfizer.cc:523: void Cnfizer::retrieveClause(PTRef, vec<PTRef>&): Assertion `logic.isLit (f) || logic.isOr (f)' failed.  
~~~~

aehyvari:
Hi rainoftime, thanks for the reports!  We'll start looking into these.

PR #72, opened 2020-03-30
Fixed the messy handling of distinct predicate

aehyvari:
Fixes issue #59

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -944,6 +944,27 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
     return mkFun(eq_sym, args);
 }
 
+PTRef Logic::mkDistinct(vec<PTRef>& args) {
+    if (args.size() == 0) return getTerm_false(); // Is this correct?
+    if (args.size() == 1) return getTerm_true(); // Is this correct?
+    sort(args);
+    int i = 1;
+    int j = 0;
+    for (; i < args.size(); i++, j++) {
+        if (args[j] == args[i]) {
+            return getTerm_false();
+        }
+    }
+    SymRef diseq_sym = term_store.lookupSymbol(tk_distinct, args);
+    // The boolean distinctness is either xor or false
+    if (hasSortBool(args[0])) {
+        if (args.size() > 2)
+            return getTerm_false();
+        return mkXor(args);
+    }
+    return mkFun(diseq_sym, args);
+}
+
 PTRef Logic::mkNot(vec<PTRef>& args) {
     assert(args.size() == 1);
     return mkNot(args[0]);
@@ -1156,7 +1177,7 @@ PTRef Logic::insertTerm(SymRef sym, vec<PTRef>& terms, char** msg)
     if(isEquality(sym))
         return mkEq(terms);
     if(isDisequality(sym))
-        return mkFun(sym, terms);
+        return mkDistinct(terms);
     if(isIte(sym))
         return mkIte(terms);
     if(sym == getSym_implies())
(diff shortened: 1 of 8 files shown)


src/logics/Logic.cc
```diff
@@ -944,6 +944,27 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
     return mkFun(eq_sym, args);
 }
 
+PTRef Logic::mkDistinct(vec<PTRef>& args) {
+    if (args.size() == 0) return getTerm_false(); // Is this correct?
```
blishko: If I read SMT-LIB standard correctly, distinct is defined only for 2 or more arguments, so distinct term with 0 or 1 argument is probably an error.
Nevertheless, I would argue that both distinct over 1 and 0 arguments should be TRUE, since in both cases, all arguments are distinct (the case of 0 arguments is a claim with universal quantification over empty set, so should be automatically true).

src/logics/Logic.cc
```diff
@@ -944,6 +944,27 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
...
+    if (args.size() == 1) return getTerm_true(); // Is this correct?
+    sort(args);
+    int i = 1;
+    int j = 0;
+    for (; i < args.size(); i++, j++) {
```
blishko: i,j are local to the loop, can you move them inside?
It seems writing 
for (int i = 1, j = 0; i < args.size(); i++, j++) is perfectly valid.

src/logics/Logic.cc
```diff
@@ -944,6 +944,27 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
...
+        if (args[j] == args[i]) {
+            return getTerm_false();
+        }
+    }
+    SymRef diseq_sym = term_store.lookupSymbol(tk_distinct, args);
+    // The boolean distinctness is either xor or false
+    if (hasSortBool(args[0])) {
```
blishko: In mkEq this check checks the sort of the symbol, not the first argument.
Maybe it should be like that also here?

blishko:
Maybe we should also add the example from the issue to our regression tests?

src/logics/Logic.cc
aehyvari: > I would argue that both distinct over 1 and 0 arguments should be TRUE, since in both cases, all arguments are distinct

Yes.  I'm adding a definition to the comments:

  Given args = {a_1, ..., a_n}, distinct(args) holds iff for all a_i, a_j \in args s.t. i != j: a_i != a_j

This hopefully is consistent what you wrote there.  It sounds reasonable to me.

>  distinct term with 0 or 1 argument is probably an error.

The parser already seems to throw an error.  But maybe the API can allow this, so I agree that it should be fixed.

src/logics/Logic.cc
aehyvari: Hehe I never remember how the syntax goes...  You're right.

src/logics/Logic.cc
aehyvari: Interestingly this is a bug in mkEq (shielded by the conditions inside).  I'm going to fix it the same way it is now for distinct.

blishko: True, I have just discovered this as well. It does the unnecessary checks also for real equality.

src/logics/Logic.cc
```diff
@@ -932,25 +932,27 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
...
 PTRef Logic::mkDistinct(vec<PTRef>& args) {
-    if (args.size() == 0) return getTerm_false(); // Is this correct?
-    if (args.size() == 1) return getTerm_true(); // Is this correct?
+    if (args.size() == 0) return getTerm_true();
+    if (args.size() == 1) return getTerm_true();
     sort(args);
     int i = 1;
     int j = 0;
-    for (; i < args.size(); i++, j++) {
+    for (int i = 1, j = 0; i < args.size(); i++, j++) {
```
blishko: You forgot to remove the previous declarations of i,j :)

Issue #60, opened 2020-03-27
Assertion violation at /src/smtsolvers/Debug.cc:85


rainoftime:
Hi, for this formula
~~~~
(set-logic QF_LIA)
(declare-fun v4 () Bool)
(declare-fun i0 () Int)
(assert (< i0 71))
(check-sat)
(assert (xor (and (< i0 71) v4) v4))
(check-sat)
(check-sat)
~~~~

opensmt (commit e9e14b4) throws an assertion violation
~~~~
sat
unsat
unsatisfied clause: -6   5   ; (or (not v4) (and (not (<= 71 i0)) v4))
unsatisfied clause: 5   -6   ; (or (not v4) (and (not (<= 71 i0)) v4))
opensmt: /opensmt/src/smtsolvers/Debug.cc:85: virtual void CoreSMTSolver::verifyModel(): Assertion `!failed' failed.
Aborted (core dumped)
~~~~

blishko:
Fixed by commit 8a2369a11d6ca5eef8c6fe313c0352a00cf21108 in PR #73

PR #73, opened 2020-04-02
Recent refactoring in LRA solver and fixing some issues

blishko:
This PR contains the following changes:

- Refactoring in LRA solver regarding Delta and iterating over LA variables
- Proper cleaning of LA solver and its components
- Fix for issue #60 when re-running check-sat command on solver in UNSAT state.

### DIFF
--- a/src/smtsolvers/CoreSMTSolver.cc
+++ b/src/smtsolvers/CoreSMTSolver.cc
@@ -1897,12 +1897,7 @@ lbool CoreSMTSolver::search(int nof_conflicts, int nof_learnts)
             conflictC++;
             if (decisionLevel() == 0)
             {
-                if (splits.size() > 0)
-                {
-                    opensmt::stop = true;
-                    return l_Undef;
-                }
-                else return l_False;
+                return zeroLevelConflictHandler();
             }
             learnt_clause.clear();
             analyze(confl, learnt_clause, backtrack_level);
@@ -1957,12 +1952,7 @@ lbool CoreSMTSolver::search(int nof_conflicts, int nof_learnts)
             // Simplify the set of problem clauses:
             if (decisionLevel() == 0 && !simplify())
             {
-                if (splits.size() > 0)
-                {
-                    opensmt::stop = true;
-                    return l_Undef;
-                }
-                else return l_False;
+                return zeroLevelConflictHandler();
             }
             // Two ways of reducing the clause.  The latter one seems to be working
             // better (not running proper tests since the cluster is down...)
@@ -2020,12 +2010,7 @@ lbool CoreSMTSolver::search(int nof_conflicts, int nof_learnts)
                 else if (value(p) == l_False)
                 {
                     analyzeFinal(~p, conflict);
-                    if (splits.size() > 0)
-                    {
-                        opensmt::stop = true;
-                        return l_Undef;
-                    }
-                    else return l_False;
+                    return zeroLevelConflictHandler();
                 }
                 else
                 {
@@ -2361,6 +2346,18 @@ void CoreSMTSolver::clearSearch()
 //    if (first_model_found || splits.size() > 1) {
         theory_handler.backtrack(-1);
 //    }
+}
+
+lbool CoreSMTSolver::zeroLevelConflictHandler() {
+    if (splits.size() > 0)
+    {
+        opensmt::stop = true;
+        return l_Undef;
+    }
+    else {
+        ok = false;
+        return l_False;
+    }
 }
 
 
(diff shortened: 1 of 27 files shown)


src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -2363,6 +2348,18 @@ void CoreSMTSolver::clearSearch()
 //    }
 }
 
+lbool CoreSMTSolver::zeroLevelConflictHandler() {
```
aehyvari, line 2351: I like this change, but I hope it'll be temporary.  I'd like to have the scatter-based splitting in a derived class.

src/tsolvers/lasolver/LABounds.cc
```diff
@@ -152,21 +150,10 @@ void LABoundStore::buildBounds()
     }
 
     // make sure all variables have at least the trivial bounds
```
aehyvari, line 152: -> the undefined bound?

src/tsolvers/lrasolver/LRAModel.cc
```diff
@@ -7,8 +7,8 @@
 
 void LRAModel::init()
 {
-    for (unsigned i = 0; i < bs.nVars(); i++) {
-        LVRef ref {i};
+    LAVarStore varStore = bs.getVarStore();
```
aehyvari, line 10: Should this be a reference?

src/tsolvers/lrasolver/LRAModel.cc
```diff
@@ -90,24 +88,32 @@ void LRAModel::clear() {
...
 const LABound& LRAModel::readLBound(const LVRef &v) const { return bs[readLBoundRef(v)]; }
-LABoundRef LRAModel::readUBoundRef(LVRef v) const { return int_ubounds[getVarId(v)].last(); }
+LABoundRef LRAModel::readUBoundRef(LVRef v) const { assert(hasUBound(v)); return int_ubounds[getVarId(v)].last(); }
 const LABound& LRAModel::readUBound(const LVRef &v) const { return bs[readUBoundRef(v)]; }
 void LRAModel::pushBacktrackPoint()      { bound_limits.push(bound_trace.size()); }
 void LRAModel::popBacktrackPoint() { popBounds(); bound_limits.pop(); }; // Returns the decision if the backtrack point had a decision
 int  LRAModel::getBacktrackSize() const { return bound_limits.size(); }
 
-bool LRAModel::isEquality(LVRef v) const { return bs[int_lbounds[getVarId(v)].last()].getIdx().x+1 == bs[int_ubounds[getVarId(v)].last()].getIdx().x && !Lb(v).isInf() && !Ub(v).isInf() && Lb(v) == Ub(v); }
+//bool LRAModel::isEquality(LVRef v) const { return hasLBound(v) && hasUBound(v) && bs[int_lbounds[getVarId(v)].last()].getIdx().x+1 == bs[int_ubounds[getVarId(v)].last()].getIdx().x && !Lb(v).isInf() && !Ub(v).isInf() && Lb(v) == Ub(v); }
```
aehyvari, line 100: Is there a way of getting the equalities now?  I think this might be important in the LIA implementation.  Can be like this for now, though, given that the LIA implementation is still to materialise...

src/smtsolvers/CoreSMTSolver.cc
blishko: That would be great.

src/tsolvers/lrasolver/LRAModel.cc
blishko: Yes, definitely!

src/tsolvers/lrasolver/LRAModel.cc
blishko: I belie I have commented it out, because it was not used anywhere in current implementation. I am bringing it back.
It should return true when there is both upper bound and lower bound and they have the same value, correct? I am not sure what does this check represents `bs[int_lbounds[getVarId(v)].last()].getIdx().x+1 == bs[int_ubounds[getVarId(v)].last()].getIdx().x`

aehyvari: > It should return true when there is both upper bound and lower bound and they have the same value, correct?

Yes.

> bs[int_lbounds[getVarId(v)].last()].getIdx().x+1 == bs[int_ubounds[getVarId(v)].last()].getIdx().x

It should represent that the lower bound is the same as upper bound, and it's correctness depends on the exact way how bounds are ordered in the bound lists.  I don't remember how well it's been tested though.

blishko: OK, so, why isn't this check enough?
``` Lb(v) == Ub(v)```
Is there a situation when these two checks would return different result?

blishko:
@aehyvari , I think I addressed everything. Can you check again?

PR #101, opened 2020-04-27
Support for (get-model) command as required by model validation track of SMT-COMP 2020

blishko:
Added support for (get-model) command in the interpreter and fix a few bugs in the backend.


blishko:
@aehyvari , What is the proper way to print the output in the interpreter?

src/api/Interpret.h
```diff
@@ -82,6 +83,7 @@ class Interpret {
     vec<PTRef>      vec_ptr_empty;
 
     vec<PTRef>      assertions;
+    vec<SymRef>     user_declarations;
```
aehyvari, line 86: This won't work for non-nullary symbols.  They would really have to be PTRefs.

PR #111, opened 2020-05-27
Fix for LIA solver regarding counting bounds and adding bounds to Simplex

blishko:
This PR contains two changes:
1. Fixed counting of activated bounds on variables, which could be triggered in LIA when new inequalities are created that could be immediately deduced (but are not).

2. Improved the declaration of bounds to Simplex for LIA solver, where the bounds can be (and should be) rounded to integer values.


src/logics/LALogic.cc
```diff
@@ -23,6 +23,29 @@ bool LALogic::isNegated(PTRef tr) const {
...
+    }
+    return false;
+}
+
+bool LALogic::isLinearTerm(PTRef tr) const {
+    if (isLinearFactor(tr)) { return true; }
+    if (isNumPlus(tr)) {
+        Pterm const& term = getPterm(tr);
+        for (int i = 0; i < term.size(); ++i) {
+            if (!isLinearFactor(term[i])) { return false; }
```
aehyvari, line 42: There's no need to support nested sums because Logic doesn't allow to construct them.

blishko: I agree, that's why inside the sums, there must be LinearFactors, not LinearTerms again. LinearFactor can be only a variable multiplied by a constant (or just variable or just constant).

Issue #112, opened 2020-05-27
Memory allocation failure


5hadowblad3:
Hi, there.

I encounter a memory allocation failure with the attached input.
To reproduce it, just run
~~~~
./opensmt input.smt2
~~~~

This is the error information.
~~~~
opensmt: malloc.c:2868: mremap_chunk: Assertion `((size + offset) & (GLRO (dl_pagesize) - 1)) == 0' failed.
~~~~

[opensmt-malloc1.smt2.zip](https://github.com/usi-verification-and-security/opensmt/files/4689213/opensmt-malloc1.smt2.zip)

rainoftime:
In debug mode, it trigged an assertion violation
~~~~
opensmt: /opensmt/src/logics/Logic.cc:2040: void Logic::conjoinItes(PTRef, PTRef&): Assertion `id < size' failed.
Aborted
~~~~

blishko:
Thanks for the report, we have a fix ready and will integrate it soon.

blishko:
Fixed in commit 5b27a03, #114

PR #114, opened 2020-05-28
Fix for a broken invariant where the whole formula is ITE on the top-level

blishko:
Method Logic::conjoinItes relies on an invariant that given a formula, all of its subterms have lower ID than the root. This is true in most cases since to create a term, its subterms must have already been created. But ITEs are handled in a special way: they are represented by a variable that has a definition and this definition contains that variable as a subterm, hence has higher ID than the variable.
Thus is the root term passed to this method is an ITE variable, the invariant is broken. 
This PR handles the special case.


blishko:
Fixes #112

PR #121, opened 2020-06-09
Refactoring of TSolver

blishko:
This PR removes the `deduced` array from Theory (and consequently from theory solvers). The TSolvers now use the polarity map to internally keep track of literals they already know the polarity of and asserting such literal would not change the state of the TSolver.

Additionally, some high-level methods are introduced to the base class TSolver which should be used by the derived theory solvers. This unifies the handling of deductions and marking literals as known and lifts the burden of implementing the correct behaviour from the derived theory solvers.

### DIFF
--- a/src/logics/UFLRATheory.h
+++ b/src/logics/UFLRATheory.h
@@ -45,7 +45,6 @@ class UFLRATheory : public Theory
     { }
     virtual LRALogic& getLogic() { return lralogic; }
     virtual UFLRATHandler& getTSolverHandler() { return uflratshandler; }
-    virtual UFLRATHandler *getTSolverHandler_new(vec<DedElem> &d) { return new UFLRATHandler(config, lralogic, tmap); }
     virtual bool simplify(const vec<PFRef>&, int);
 };
 
--- a/src/logics/UFLRATheory.h
+++ b/src/logics/UFLRATheory.h
@@ -41,11 +41,11 @@ class UFLRATheory : public Theory
         : Theory(c)
         , lralogic(c)
         , tmap(lralogic)
-        , uflratshandler(c, lralogic, deductions, tmap)
+        , uflratshandler(c, lralogic, tmap)
     { }
     virtual LRALogic& getLogic() { return lralogic; }
     virtual UFLRATHandler& getTSolverHandler() { return uflratshandler; }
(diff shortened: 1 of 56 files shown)


src/logics/UFLRATheory.h
```diff
@@ -41,11 +41,11 @@ class UFLRATheory : public Theory
         : Theory(c)
         , lralogic(c)
         , tmap(lralogic)
-        , uflratshandler(c, lralogic, deductions, tmap)
+        , uflratshandler(c, lralogic, tmap)
     { }
     virtual LRALogic& getLogic() { return lralogic; }
     virtual UFLRATHandler& getTSolverHandler() { return uflratshandler; }
-    virtual UFLRATHandler *getTSolverHandler_new(vec<DedElem> &d) { return new UFLRATHandler(config, lralogic, d, tmap); }
+    virtual UFLRATHandler *getTSolverHandler_new(vec<DedElem> &d) { return new UFLRATHandler(config, lralogic, tmap); }
```
aehyvari, line 48: This can be removed.

src/tsolvers/TSolverHandler.cc
```diff
@@ -102,5 +102,17 @@ TRes TSolverHandler::check(bool complete)
     return res_final;
 }
 
+// MB: This is currently needed to replace a common array of deduced elements with solver ID
+TSolver* TSolverHandler::getSolverDeducing(PTRef ptref) const {
+    // MB: Can we use solverSchedule? Double check this if theory combination is implemented
+    for (auto* solver : tsolvers) {
+        if (solver != nullptr && solver->isValid(ptref)) {
```
aehyvari, line 109: Is there some global assertion here that states that ptref is assigned at this point?

src/logics/UFLRATheory.h
blishko: Ah, true, I missed that. I think these methods were removed for other handlers and I missed this one, I will remove it, thanks!

src/tsolvers/TSolverHandler.cc
blishko: I will check this.

src/logics/UFLRATheory.h
blishko: Removed

src/tsolvers/TSolverHandler.cc
blishko: The method has been renamed to better reflect its meaning.

PR #126, opened 2020-06-18
Model and substitutions

blishko:
This PR contains two big changes:
1. It introduces new Model object with new approach to computing and returning models of satisfiable formulas.
2. It changes how the substitutions are computed and handled. Since even in incremental mode, the main substitution code receives the whole original formula, there is no need to include any additional information from previous frames.


blishko:
Since we have split the substitution and model, I will create a separate PR dedicated to Model. The substitution part has been done already in #147

PR #131, opened 2020-06-19
Fixing soundness bug in LIA and simple refactoring in Simplex

blishko:


src/api/Interpret.cc
```diff
@@ -57,6 +57,13 @@ bool stop;
 
 uint32_t LetFrame::id_cnt = 0;
 
+namespace {
+bool isPreLogicOption(char const * const option) {
```
aehyvari, line 61: Can we move this to be part of SMTConfig?

src/api/Interpret.cc
```diff
@@ -164,6 +171,11 @@ void Interpret::setOption(ASTNode& n) {
     else // Normal option
         name = strdup(n.getValue());
 
+    if (logic && isPreLogicOption(name)) {
```
aehyvari, line 174: Would it make sense to move this into SMTConfig?  We'd need to update the logic in the config file so that the config knows whether it has been set.  This could break some binaries linking to osmt possibly.

src/tsolvers/lasolver/Simplex.cc
```diff
@@ -321,44 +309,20 @@ void Simplex::updateValues(const LVRef bv, const LVRef nv){
...
-            assert(var != x);
-            if (coeff < 0) {
-                LABoundRef br = model->readLBoundRef(var);
-                expl.push_back({br, -coeff});
-            }
-            else {
-                LABoundRef br = model->readUBoundRef(var);
-                expl.push_back({br,coeff});
-            }
+    LABoundRef br_f = conflictOnLower ? model->readLBoundRef(x) : model->readUBoundRef(x);;
```
aehyvari, line 312: Extra `;`

src/api/Interpret.cc
aehyvari: We actually have in place a method of keeping the solver's logic and config's logic in sync. This code could fit well with  PR #139.

src/tsolvers/lasolver/Simplex.cc
blishko: Fixed

src/api/Interpret.cc
blishko: Changes related interpolation interface has been removed from this PR and a separate PR will be created for that.

src/api/Interpret.cc
blishko: Changes related interpolation interface has been removed from this PR and a separate PR will be created for that.

blishko:
The original changes in the PR has been split to two parts and this PR now contains only changes to the fix of LIA soundness bug (and some refactoring in Simplex).

blishko:
@aehyvari Working on this, I was thinking if it would not be possible to unify LRASolver and LIASolver (and also maybe LRALogic and LIALogic) since it does not seem there are that many differences and it we would get support for LIRA for free.

aehyvari:
> @aehyvari Working on this, I was thinking if it would not be possible to unify LRASolver and LIASolver (and also maybe LRALogic and LIALogic) since it does not seem there are that many differences and it we would get support for LIRA for free.

This could make sense for the current implementation of LIA solving we have now.  Still, for readability, isn't there a way of doing this with minimal code duplication using inheritance?  I'd be fine with a derived class that implements QF_LIRA.

Issue #132, opened 2020-07-02
Build option for static library only


leonardoalt:
It would be nice to have a build option that only generates `libopensmt2.a`, without the shared library and the executable.
This is useful for systems that use the solver via the API.

blishko:
I agree that this should be possible. It is on my long TODO list and it has just moved closer to the top :)

andrewvaughanj:
@leonardoalt could you take a look at https://github.com/usi-verification-and-security/opensmt/pull/133 ?

blishko:
@leonardoalt Do you also want to be able to install the static library (e.g., with `make install`)?
I think the current set-up is that only the executable and the shared library are installed this way.

leonardoalt:
@blishko I'm fine without it being part of `make install`

blishko:
I think we can close this one. 
Thanks @andrewvaughanj !

The next step will be dealt with in a separate issue (#135)

PR #133, opened 2020-07-02
Providing support for building shared libraries/static libraries/target executables in isolation

andrewvaughanj:
## Problem

Issue #132 requested the ability to build "only" the static libraries.

## Resolution

This PR adds support to:

- *turning off* building static libraries (`-DBUILD_STATIC_LIBS:BOOL=OFF`)

- *turning off* building shared libraries (`-DBUILD_SHARED_LIBS:BOOL=OFF`)

- *turning off* building the final executable (`-DBUILD_EXECUTABLES:BOOL=OFF`)

By default, *all* are `ON`, which means that the current behaviour is unchanged. You *cannot* build executables *without* building static libraries -- this will error-out when running `cmake`.

### Build only static

```
cmake \
    -G Ninja \
    -DBUILD_STATIC_LIBS:BOOL=ON \
    -DBUILD_SHARED_LIBS:BOOL=OFF \
    -DBUILD_EXECUTABLES:BOOL=OFF \
    -DPACKAGE_TESTS:BOOL=OFF ..
```

### Build only shared

```
cmake \
    -G Ninja \
    -DBUILD_STATIC_LIBS:BOOL=OFF \
    -DBUILD_SHARED_LIBS:BOOL=ON \
    -DBUILD_EXECUTABLES:BOOL=OFF \
    -DPACKAGE_TESTS:BOOL=OFF ..
```

### Build executables (no shared)

Explicit:

```
cmake \
    -G Ninja \
    -DBUILD_STATIC_LIBS:BOOL=ON \
    -DBUILD_SHARED_LIBS:BOOL=OFF \
    -DBUILD_EXECUTABLES:BOOL=ON \
    -DPACKAGE_TESTS:BOOL=OFF ..
```

Alternative (static and executables default to `ON`):

```
cmake \
    -G Ninja \
    -DBUILD_SHARED_LIBS:BOOL=OFF \
    -DPACKAGE_TESTS:BOOL=OFF ..
```

Signed-off-by: Andrew V. Jones <andrew.jones@vector.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ include_directories(
 )
 
 if (BUILD_EXECUTABLES AND NOT BUILD_STATIC_LIBS)
-  message(FATAL_ERROR "Cannot build executables without also building shared libraries")
+  message(FATAL_ERROR "Cannot build executables without also building static libraries")
 endif()
 
 if (ITP_DEBUG)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ option(LADEBUG "Debug lookahead" OFF)
 option(PRINT_DECOMPOSED_STATS "Print statistics about decomposed interpolants" OFF)
 option(STATISTICS "Compute and print solver's statistics" OFF)
 option(USE_READLINE "Use readline over libedit" OFF)
+option(BUILD_STATIC_LIBS "Build static library" ON)
+option(BUILD_SHARED_LIBS "Build shared library" ON)
+option(BUILD_EXECUTABLES "Build executables" ON)
 
 if (NOT USE_READLINE)
   find_package(LibEdit REQUIRED)
@@ -70,6 +73,10 @@ include_directories(
     ${CMAKE_SOURCE_DIR}/proof
     ${GMP_INCLUDE_DIR}
 )
+
+if (BUILD_EXECUTABLES AND NOT BUILD_STATIC_LIBS)
+  message(FATAL_ERROR "Cannot build executables without also building shared libraries")
+endif()
 
 if (ITP_DEBUG)
     add_definitions(-DITP_DEBUG)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -71,6 +74,10 @@ include_directories(
     ${GMP_INCLUDE_DIR}
 )
 
+if (BUILD_EXECUTABLES AND NOT BUILD_STATIC_LIBS)
+  message(FATAL_ERROR "Cannot build executables without also building shared libraries")
```
blishko, line 78: without **static** libraries

blishko:
We should also update the README to mention these options.

CMakeLists.txt
andrewvaughanj: Done!

andrewvaughanj:
> We should also update the README to mention these options.

All documented.

Should we create an issue for documenting all of the other CMake options?

blishko:
@andrewvaughanj 
I took the liberty to fix the name of the static archive in this PR, hope you don't mind.

> Should we create an issue for documenting all of the other CMake options?

Yes, it makes sense to do so.

andrewvaughanj:
> @andrewvaughanj
> I took the liberty to fix the name of the static archive in this PR, hope you don't mind.
> 

Yep, totally cool!

> > Should we create an issue for documenting all of the other CMake options?
> 
> Yes, it makes sense to do so.

If I get some cycles, I can take a look at this.

leonardoalt:
@andrewvaughanj oh nice, thanks!!

leonardoalt:
Maybe not necessarily as part of this PR, but I would further suggest:
Does the library (either dyn or static) use `src/api/Interpret*`? That's the only source that depends on `libedit`. Removing those sources from the library build would also remove `libedit` from the static library, decreasing its size and dependencies.
Context: for a PoC we're currently removing those sources manually when building the static library with emscripten

andrewvaughanj:
There's also other "issues" with `Interpret.cc` I've spotted that should also be fixed.

Rather than removing `Interpret.cc` all together, I think it would be better to allow a user to disable `libedit`, and then `#if` out the code in `Interpret.cc`.

It might literally be "#if out `Interpret::interpInteractive`".

Would that work?

blishko:
Removing that dependencies is also something I wanted to do, but we have to discuss also with @aehyvari to decide on the best way.

leonardoalt:
@andrewvaughanj yep, that's also what I meant

blishko:
We are going to address the dependency of the library on EditLine in a separate issue (#135), we want to do some additional refactoring related to that part of the code.

PR #136, opened 2020-07-04
enabling MacOS build in Travis CI

blishko:
This PR add a MacOS build to Travis CI.
We add only a single configuration for MacOS using Debug build and Clang compiler, because the MacOS is slower than Linux build. This is caused by the need to upgrade Bison, which requires an update of the Homebrew in the VM of the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: cpp
 branches:
     only:
         - master
-
 os:
   - linux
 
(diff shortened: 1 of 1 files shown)


andrewvaughanj:
Quick query: would it be quicker to build bison manually vs. using homebrew?

blishko:
> Quick query: would it be quicker to build bison manually vs. using homebrew?

I don't know. I have not tried. Using Homebrew seems easier and it's what I found in Travis docs.
If you want to improve, you are welcome to try, but I don't think it is a big issue, unless we really want to save on network traffic and avoid updating Homebrew.

andrewvaughanj:
Yep, really not a problem (to me) -- it was just a thought given OpenSMT seemed "constrained" by Homebrew's speed.

blishko:
I may have been too dramatic in the description of the PR.

Issue #137, opened 2020-07-07
Incorrect handling of equality with more than 2 arguments


blishko:
OpenSMT gives wrong answer on the following example:

```
(set-logic QF_UF)
(declare-fun x () Bool)
(declare-fun y () Bool)
(assert (= x y (= x false)))
(check-sat)
```

The problem is that according to SMT-LIB standard, equality is a chainable operation, so an equality with 3 arguments should be interpreted as a chain of equalities with two arguments.

However, OpenSMT currently does not handle this case correctly.

blishko:
Fixed in #138

PR #138, opened 2020-07-07
Fixing the handling of equality with more than 2 arguments

blishko:
Fix for issue #137


rainoftime:
Is there any plan for extending inequality to support more than 2 arguments? Thanks!

blishko:
> Is there any plan for extending inequality to support more than 2 arguments? Thanks!

Do you mean `distinct` with multiple arguments? In which context?
It should be working in QF_UF,  but I noticed some weird behaviour in QF_LRA if more than 2 arguments are used. Is that what you mean? Can you create an issue a give a more detailed description there?

rainoftime:
Including distinct and >, >=, etc
e.g.,
~~~~
(assert (>= x y z)
~~~~

andrewvaughanj:
@rainoftime I'm curious, do you need this for something you're building, or is this more to enable "better" bug finding?

I'm trying to understand if this is a feature request for a real-world use-case of OpenSMT vs. a "it would be nice if OpenSMT supported $n$-ary operators".

rainoftime:
Neither. This is not a "feature request".
I am just curious (since some other solvers support the feature for both equality and inequality, and this PR extends the equality operation)

blishko:
I have consulted the SMT-LIB standard and these operators are defined as chainable in theories of reals and integers, so it should be supported, if only to be standard-compliant.
I will create an issue for this.

Issue #135, opened 2020-07-03
Remove the dependency of OpenSMT library on Editline/Readline


blishko:
### Issue
The library for processing command line is needed only in the executable, but not in the (static or dynamic) library.

### Task
Refactor the text so Interpret class can be moved to bin folder and excluded from the library build.

#### Note
We still want to keep the ability to parse and interpret some SMT-LIB commands in the library, it is needed for parallel version of OpenSMT in SMTS.

blishko:
Done in #151

PR #139, opened 2020-07-11
Changing the OpenSMT API

blishko:
This PR aims to solve #135 

This is solved by moving the code working with this library to the main .cc file which is part of the executable only.

In addition, some refactoring has been done to better define the relations between the main components: MainSolver now becomes responsible was maintaining all other components (SimpSMTSolver, Theory, THandler, ...), but it is created with a specific Theory.


blishko:
@aehyvari please check the changes and let me know if you think this would be OK.
There is also the question if we should do something about `MainSolver::readFormulaFromFile` method and `Interpret::parse_only` field.

blishko:
The original PR has been split to two parts. The first part that dealt with restricting the dependency on Readline/Editline library has been moved to #151.

The aim of this second part is to change the API of OpenSMT and the main classes so that single Logic object can be used with multiple MainSolvers (still in sequential program only). Also, Logic and MainSolver (apart from SMTConfig) would be the only classes visible to the user.

blishko:
There is still a problem with the method `MainSolver::readFormulaFromFile`, which not would not work.

src/options/SMTConfig.h
```diff
@@ -319,6 +320,12 @@ struct SMTConfig
   vec<char*>    info_names;
   Map<const char*,Info*,StringHash,Equal<const char*> >   infoTable;
   Map<const char*,SMTOption*,StringHash,Equal<const char*> > optionTable;
+
+  bool usedForInitialization = false; // Some options can be changed only before this config is used for initialization of MainSolver
+  bool isPreInitializationOption(const char* o_name) {
+      return strcmp(o_name, o_produce_inter) == 0 || strcmp(o_name, o_produce_proofs) == 0;
```
aehyvari, line 326: We should add:
 - o_sat_pure_lookahead
 - o_sat_lookahead_split
 - o_ghost_vars

blishko: These options have been added.

blishko:
It looks like there are also two regression tests in the splitting folder where the options for splitting are after the (set-logic) command. The question is, is it OK to pick the type of the solver first and other parameters can be set later (parameters such as split-num or split-format)? Or do these also need to be set before set-logic?

aehyvari:
> It looks like there are also two regression tests in the splitting folder where the options for splitting are after the (set-logic) command.

Ok excellent, I'm fixing them!

> The question is, is it OK to pick the type of the solver first and other parameters can be set later (parameters such as split-num or split-format)? Or do these also need to be set before set-logic?

In the code this is not a problem, and I can't see how this would in the future cause a performance overhead.  It might actually be a good thing when thinking about a parallel implementation.  Let's keep it that way.

Issue #92, opened 2020-04-09
Invalid model for QF_LRA formula (assertion violation at Debug.cc:85) 


rainoftime:
Related issue https://github.com/usi-verification-and-security/opensmt/issues/79 that was fixed in https://github.com/usi-verification-and-security/opensmt/pull/91

-----------

Hi, for the following instance
~~~~
(set-logic QF_LRA)
(declare-fun v6 () Bool)
(declare-fun v7 () Bool)
(declare-fun v13 () Bool)
(declare-fun v15 () Bool)
(assert (=> (=> v15 v6) (= v13 v7)))
(check-sat)
(check-sat)
(push 1)
(check-sat)
(pop 1)
(assert (= v15 (= v13 v7)))
(check-sat)
(push 1)
(check-sat)
(pop 1)
(assert (distinct true v15))
(push 1)
(check-sat)
(pop 1)
(push 1)
(check-sat)
(pop 1)
(push 1)
(check-sat)
(pop 1)
(push 1)
(check-sat)
(pop 1)
(check-sat)
~~~~

opensmt (commit a492c61689)  throws an assertion violation
~~~~
sat
sat
sat
sat
sat
unsat
unsat
unsat
unsat
unsatisfied clause: 3   7   ; (or v15 (or (not v15) v6))
unsatisfied clause: 11  4   ; (or (= v13 v7) (or v6 (not (= v13 v7))))
opensmt: /home/rainoftime/Work/tofuzz/opensmt/src/smtsolvers/Debug.cc:85: virtual void CoreSMTSolver::verifyModel(): Assertion `!failed' failed.
~~~~

rainoftime:
Here is another test case
~~~~
(set-logic QF_LRA)
(declare-fun r14 () Real)
(push 1)
(check-sat)
(pop 1)
(declare-fun v12 () Bool)
(assert (or (<= 582095008.0 r14) v12))
(assert (and (=> v12 (<= 582095008.0 r14)) (not (<= 582095008.0 r14))))
(push 1)
(check-sat)
(push 1)
(push 1)
(push 1)
(push 1)
(push 1)
(check-sat)
(pop 1)
(pop 1)
(pop 1)
(check-sat)
(pop 1)
(check-sat)
(pop 1)
(pop 1)
(check-sat)
~~~~

blishko:
Thanks, I see the problem.
You assert unsatisfiable formula, but you don't check-sat right away; you check-sat only after push.
[...]

PR #141, opened 2020-07-19
An attempt for better fix of handling the situation when new frames are pushed to solver already in UNSAT state

### LINKED ISSUE #140: Solution soundness bug on QF_LRA and QF_UF formulas
Hi, for the following formula
~~~~
(set-logic QF_LRA)
(assert false)
(push 1)
(check-sat)
(pop 1)
(check-sat)
(check-sat)
(check-sat)
(check-sat)
(check-sat)
(check-sat)
(assert false)
(check-sat)
(check-sat)
~~~~
opensmt 3599a01a2 yields
~~~~
unsat
sat
sat
sat
sat
sat
sat
unsat
unsat
~~~~

z3 and cvc4
~~~~
unsat
unsat
unsat
unsat
unsat
unsat
unsat
unsat
unsat
~~~~

blishko:
Fixes #140 #92


blishko:
>I was thinking of a solution where the unsatisfiability would be marked already at push, but this would not work if check is not called after every push.

That is exactly the problem with these examples, where the asserted formula is UNSAT, but you DON'T call check-sat and keep pushing.

it is possible that Jack will show us that even this solution is still not the proper one :-)

Issue #122, opened 2020-06-09
For test-bv.cc, Why the result of this "2 + 1 2" is 1 ?


Flians:
For test-bv.cc, Why the result of this "2 + 1 2" is 1 ?

```
L-T95HG8WN-1422:build flynn$ ./test-bv 2 "+" 1 2
Computing 1 (01) + 2 (10)
sat
1 (01)
L-T95HG8WN-1422:build flynn$ ./test-bv 2 '/' 00 00
Computing 00 (00) / 00 (00)
sat
3 (11)
L-T95HG8WN-1422:build flynn$ ./test-bv 2 '/' 00 01
Computing 00 (00) / 01 (01)
sat
3 (11)
L-T95HG8WN-1422:build flynn$ ./test-bv 2 '/' 01 01
Computing 01 (01) / 01 (01)
sat
3 (11)
```

Flians:
How can I resolve this problem? The file test-bv.cc is from [your examples](https://github.com/usi-verification-and-security/opensmt/blob/master/examples/test-bv.cc).

aehyvari:
Hi Rongliang,

These are currently not our priority in the development.  They are related
to the bit vector implementation in the code that is not officially
supported.  You can safely ignore these files.

Best Regards,
Antti

Flians:
Hi @aehyvari ,
Now I wanna verify three-valued logic. 
So I wanna use 2 bit to present TRUE, FALSE, and UNKNOWN.
How can I do this use opensmt?
![image](https://user-images.githubusercontent.com/16850481/84219430-198acd80-ab03-11ea-8933-d492b40fb73a.png)

andrewvaughanj:
For context: https://github.com/Z3Prover/z3/issues/4497

Flians:
@andrewvaughanj 
Feel that you are involved in the development of all smt projects.
Awesome!
Thank you!

Issue #123, opened 2020-06-09
Wrong propositional model when substitutions are enabled


blishko:
For the following propositional formula which is satisfiable, OpenSMT produces wrong model. The computed model is correct when substitutions are disabled.

```
(set-option :produce-model 1)
(set-logic QF_UF)
(declare-fun b () Bool)
(declare-fun c () Bool)
(assert (= b c ))
(assert (not b))
(check-sat)
(get-model)
``` 

Computed model is:
```
(define-fun b () Bool true)
(define-fun c () Bool true)
```

blishko:
This is most likely also the cause of the issue #122

blishko:
Fixed in #142

PR #142, opened 2020-07-21
Fixing problem with model computation in the presence of substitutions (boolean variable to False)

blishko:
This PR addresses the issue when formula has been simplified using a substitution `v -> False` for some boolean variable `v` but this substitution has not been remembered, so the model computed in case of satisfiability was not aware of this substitution.

Fixes issues #122 and #123 

Fixes the example of evaluating bit vector operations


blishko:
It seems the Travis check is now passing even if the output in regression test is different than expected.
This is most likely related to a change I did to enable Mac OS build.
I will try to fix it.

blishko:
I have put the main part of Travis script to external shell script. This enables the use of `set -e` to stop on first error. It was having some weird effect in MacOS build when it was a direct command in .yml.
Using in directly in .yml is also discouraged by Travis: https://docs.travis-ci.com/user/common-build-problems/#my-build-fails-unexpectedly

PR #147, opened 2020-07-22
Refactoring of substitutions (part 1)

blishko:
The goal of this PR is to simplify the code related to preprocessing of the input formula, especially the code related to substitutions.

Currently there is a a significant amount of code devoted to keeping various information about substitutions discovered at different frames (in the context of incremental solving).
However, at the same time, the input formula to this process is the whole formula up to the current frame (as asserted to the MainSolver). This means the substitution code has access to the whole formula up to the current frame and does not need the bookkeeping from previous frames.

In this part of the refactoring I am assuming that the latter is still true, i.e., the preprocessing will see the whole formula up to the current frame. Under this assumption the code for computing substitutions can be self-contained, i.e., there is no need for any bookkeeping about previous computations (in previous frames). Thus the code can be simplified.


blishko:
@aehyvari Let me know what you think about this idea. As the next step I would like to turn this around and bring some bookkeeping back so that actually the frames would be self-contained and in the incremental mode, the solver would only process the new parts of the formula and would not have to consider the part of the formula from unchanged frames again.

blishko:
I have also noticed I have left some TODOs in `LALogic::arithmeticElimination` I will try to address them.

Issue #148, opened 2020-07-22
Regression tests for splitting are brittle


aehyvari:
Splitting code in OpenSMT takes an SMT instance and produces constraints that enable cube-and-conquer style solving in OpenSMT.  It is important to have regression tests for this area since the code is not tested by standard regression tests.

The correctness criterion of a result goes beyond that of checking whether the satisfiability determined by the solver equals that of the instance, since the idea is not to solve the instance but instead produce the cubes.  As part of the splitting output, the test cases produce a list of literals that correspond to the cubes.

In current master the splitting benchmarks are implemented similar to the rest of the regression tests.  As part of the splitting output, the test cases produce a list of literals that correspond to the cubes.  The issue is that the literals are very sensitive to small changes in the SMT solver.  This results in changing output and spurious errors in regressions.

There should be a separate system for checking the regressions for splitting code based on some more semantic flavour of the correctness of splitting.

PR #149, opened 2020-07-22
Separate regression system for splitting code

aehyvari:
This PR addresses issue #148 by introducing a separate regression testing system for the splitting code.

The tests have three possible outcomes: `unsat`, `sat`, and `unknown`, and the expected output is hardcoded in the test script.  The interesting case is when expected result is `unknown`: If the solver returns `unknown`, the output is tested by constructing a tree of literals corresponding to a split, i.e., a set of cubes.  The splits are tested for the following properties:
 - each cube must be consistent
 - there must not be duplicate literals in a cube
 - the tree must be binary
 - if a node has two children, one must be the positive and the other the negative literal
 - the tree must be balanced
 - the number of descendants of a node (not including the node) must be
    - 2^(D-d), where D is the depth of the tree and d is the depth of the node for an inner node
    - 0 for a leaf node.

The last condition is testing more the consistency of the script itself.

In addition an `unknown` result must produce a split, and non-`unknown` result must not produce a split


ci/run_travis_commands.sh
```diff
@@ -9,6 +9,7 @@ make install
 if [[ ${CMAKE_BUILD_TYPE} == Debug ]]; then
     cd ../regression && ./run-test-notiming.sh ../build/src/bin/opensmt;
     cd ../regression_itp && ./run-tests.sh ../build/src/bin/opensmt;
+    cd ../regression_splitting && ./bin/run_tests ../build/src/bin/opensmt;
```
blishko, line 12: Missing extension: `./bin/run_tests.sh`

aehyvari: Thanks!

regression_splitting/bin/check_result.py
```diff
@@ -0,0 +1,96 @@
...
+def getLinesFromOutput(file):
+    lines = map(lambda x: x.strip(), open(file).readlines())
+    lines = filter(lambda x: re.search("; [-]?[0-9][0-9]*", x), lines)
+    lines = map(lambda x: x.split(), map(lambda x: x[2:], lines))
+    lines = [list(map(lambda x: int(x), l)) for l in lines]
+    for line in lines:
+        abses = set(map(abs, line))
+        if len(abses) != len(line):
+            die("inconsistent path")
+        if len(set(line)) != len(line):
```
aehyvari, line 20: This check seems to fail always if the above check fails, so is redundant.

regression_splitting/bin/check_result.py
```diff
@@ -0,0 +1,96 @@
...
+    if len(sys.argv) != 3:
+        print("Usage: %s <output-file> <expected>" % sys.argv[0])
+        sys.exit(1)
+
+    lines = getLinesFromOutput(sys.argv[1])
+    expected = sys.argv[2]
+
+    if len(lines) == 0 and expected == 'unknown':
+        die("Expected a split")
+        sys.exit(1)
```
aehyvari, line 79: Not reached.

regression_splitting/bin/check_result.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if expected != 'unknown':
+        s = open(sys.argv[1]).read()
+        mo = re.search('^(sat|unsat)', s)
+        if not mo:
+            die("Expected %s but found no result" % expected)
+        if mo.group(1) != expected:
+            die("Expected %s but got %s" % (expected, mo.group(1)))
+        sys.exit(0)
+
```
aehyvari, line 91: Maybe add else and move sys.exit(0) to the end.

PR #150, opened 2020-07-23
Better handling of base-frame conflict in incremental solving

### LINKED ISSUE #144: Invalid model for QF_LRA formula (Debug.cc:85)
~~~~
(set-logic QF_LRA)
(declare-fun v3 () Bool)
(declare-fun v8 () Bool)
(declare-fun v9 () Bool)
(declare-fun v11 () Bool)
(check-sat)
(assert v11)
(assert (distinct (= v8 v3) v9))
(check-sat)
(declare-fun r14 () Real)
(assert (distinct (>= r14 0.0) v9))
(assert (distinct (>= r14 0.0) (= v8 v3)))
(push 1)
(check-sat)
(pop 1)
(check-sat)
~~~~

opensmt 56fc954
~~~~
sat
sat
unsat
unsatisfied clause: -4 -5 ; (or (not (= v8 v3)) (not v9))
opensmt: /home/opensmt/src/smtsolvers/Debug.cc:85: virtual void CoreSMTSolver::verifyModel(): Assertion `!failed' failed.
Aborted

~~~~

blishko:
If a conflict is caused by the base-frame part of the formula, but more frames are present, the solver would ignore conflict clause in level 0 and this would lead to incorrect behaviour and possibly incorrect answer as demonstrated in #144. 

The proposed solution is to remember if the conflict has been caused by the base-frame part. This translated to detecting if any assumption is part of the final conflict in the SAT solver or not.
If such situation is detected the solver remembers that it is in a permanent UNSAT state.

This fixes #144 and related issues.


aehyvari:
I experimented with an approach where you can query from the solver which active assumption level the conflict occurred once a conflict is detected.

 - The first commit implements this functionality in the simple form, and results in slightly more general code in MainSolver.  
 - The second commit is my attempt at analysing more closely the `conflict` vector in `analyze_final`: `conflict` should contain only the assumption variables that are involved in conflict.  The code checks what's the highest active assumption literal in `conflict` and returns this height.  The number corresponds to the depth in formula frames where the conflict occurred.

I think it's possible that after the analysis, the conflict might not contain the literal at the current decision level any more, but I'm not 100% sure.  If this is the case, the proposed code in my latter commit should give speed-up in incrementality.

PR #162, opened 2020-08-14
Fixing undefined behaviour in absVal

### LINKED ISSUE #159: src/common/FastRational.h:66:24: runtime error
Hi, for the following formula
[66.txt](https://github.com/usi-verification-and-security/opensmt/files/5021047/66.txt)

opensmt 62ee735ed45
~~~~
/home/opensmt/src/common/FastRational.h:66:24: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
~~~~

Reported by UBSAN

blishko:
Fixes #159 
For INT32_MIN in type int32_t the negation operation cannot be represented and `- INT32_MIN` is undefined behaviour if the type of the variable is int32_t.
For the discussion and proposed solution see [this post](https://stackoverflow.com/questions/12231560/correct-way-to-take-absolute-value-of-int-min).

Additionally, the constructor of FastRational that was not normalizing the numerator and denumerator has been removed since could break the invariant of the type.

### DIFF
--- a/src/common/FastRational.h
+++ b/src/common/FastRational.h
@@ -63,11 +63,15 @@ inline State& operator |= (State& lhs, State rhs)
 }
 
 inline uword absVal(word x) {
-    return x>=0 ? x : -x;
+    // Taking just (- WORD_MIN) is undefined behaviour, changed according to https://stackoverflow.com/questions/12231560/correct-way-to-take-absolute-value-of-int-min
+    return x < 0 ? -((uword)(x))
+                 : +((uword)(x));
 }
 
 inline ulword absVal(lword x) {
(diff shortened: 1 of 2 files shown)


aehyvari (approved):

blishko:
@aehyvari, I have a small fix and some changes in FastRationals, should I add them here or create a separate pull request?
Have you had a look at the use of try_fit_word in methods like `gcd`, `lcm`, `divexact` and `fastrat_fdiv_q`? I think it should not be called there.

blishko:
Additional changes in FastRationals will be in separate PR, closing this one.

PR #163, opened 2020-08-27
Ites implemented as switches.  Supported in logic.

aehyvari:
This PR contains two major, related features.

- Logic supports expressing explicitly ITE-terms
- ITE reasoning is implemented presumably more efficiently by stacking together nested ITEs and constructing "switches" in the propositional level that decide the equivalence of an ITE term with a normal term.

The ITE managing is extracted to a separate class from Logic and used as a post-processor for the PTRef structure before the instance is inserted to the solver.

ITE terms may be presented to the theory solvers, and they will be treated as variables of the correct sort in all supported theories.

The switches are constructed from a DAG representation and is therefore linear in the original ite structures.


src/logics/LALogic.h
```diff
@@ -46,6 +46,7 @@ class LALogic: public Logic
     static const char*  s_sort_num;
 
     Map<PTRef,bool,PTRefHash> la_split_inequalities;
+    Map<PTRef,bool,PTRefHash> constIteTerms;
```
aehyvari, line 49: I had a vague idea that storing this information would allow some optimisation, but nothing of the sort is implemented.  Should be removed?

src/api/CMakeLists.txt
```diff
@@ -22,12 +22,14 @@ add_library(api OBJECT "")
 set(PRIVATE_SOURCES_TO_ADD
 	"${CMAKE_CURRENT_SOURCE_DIR}/Interpret.cc"
 	"${CMAKE_CURRENT_SOURCE_DIR}/MainSolver.cc"
+    "${CMAKE_CURRENT_SOURCE_DIR}/IteManager.cc"
```
aehyvari, line 25: Fix indentation

src/api/CMakeLists.txt
```diff
@@ -22,12 +22,14 @@ add_library(api OBJECT "")
...
 )
 
 set(PUBLIC_SOURCES_TO_ADD
 	"${CMAKE_CURRENT_SOURCE_DIR}/Interpret.h"
 	"${CMAKE_CURRENT_SOURCE_DIR}/MainSolver.h"
 	"${CMAKE_CURRENT_SOURCE_DIR}/DimacsParser.h"
+    "${CMAKE_CURRENT_SOURCE_DIR}/IteManager.h"
 	"${CMAKE_CURRENT_SOURCE_DIR}/smt2tokens.h"
```
aehyvari, line 33: Fix indentation

regression/QF_LIA/prp-0-12_simplified_3.smt2
```diff
@@ -0,0 +1,6 @@
+(set-logic QF_LIA)
+(declare-fun v0 () Int)
+(declare-fun v2 () Int)
+(assert (= (ite (= v2 0) (ite (= v0 0) 1 0) (ite (= (ite (= v0 0) 1 0) 0) 1 0)) 0))
+(check-sat)
+(exit)
```
blishko, line 6: Could you add also a similar regression test with expected answer UNSAT?

aehyvari: Done.

src/logics/LALogic.cc
```diff
@@ -520,12 +552,13 @@ PTRef LALogic::mkNumTimes(const vec<PTRef>& tmp_args, char** msg)
     PTRef tr = mkFun(s_new, args_new);
     // Either a real term or, if we constructed a multiplication of a
     // constant and a sum, a real sum.
-    if (isNumTerm(tr) || isNumPlus(tr) || isUF(tr))
+    if (isNumTerm(tr) || isNumPlus(tr) || isUF(tr) || isIte(tr))
```
blishko, line 555: I think the ITE case should be handled inside the `isNumTerm` method.
The `isNumTerm` already returns true for terms of the form `c * ite` for `c` constant, so `ite` case should be handled the same, as it is basically `1 * ite`. This is related to my previous comment for the `isNumTerm` method.

src/logics/LALogic.cc
```diff
@@ -537,7 +570,7 @@ PTRef LALogic::mkNumDiv(const vec<PTRef>& args, char** msg)
     }
     simp.simplify(get_sym_Num_DIV(), args, s_new, args_new, msg);
     if (isNumDiv(s_new)) {
-        assert((isNumTerm(args_new[0]) || isNumPlus(args_new[0])) && isConstant(args_new[1]));
+        assert((isNumTerm(args_new[0]) || isNumPlus(args_new[0]) || isIte(args_new[0])) && isConstant(args_new[1]));
```
blishko, line 573: Same as before, `ITE` case should be handled in `isNumTerm`

src/logics/LALogic.cc
```diff
@@ -601,13 +634,15 @@ PTRef LALogic::mkNumLeq(PTRef lhs, PTRef rhs)
             args[1] = nonconst_args.size() == 1 ? nonconst_args[0] : mkFun(get_sym_Num_PLUS(), nonconst_args);
             assert(mkNumPlus(nonconst_args) == args[1]);
         }
-    } else if (isNumVar(sum_tmp) || isNumTimes(sum_tmp)) {
+    } else if (isNumVar(sum_tmp) || isNumTimes(sum_tmp) || isIte(sum_tmp)) {
```
blishko, line 637: Combine the cases and use `isNumVarOrIte`?

src/logics/LALogic.h
blishko: Yes, it seems you are not using it now, so I would remove it, unless you have an idea how to use this information.

src/logics/LALogic.cc
```diff
@@ -702,7 +752,7 @@ void SimplifyConst::simplify(const SymRef& s, const vec<PTRef>& args, SymRef& s_
     vec<int> const_idx;
     for (int i = 0; i < args.size(); i++) {
         assert(!l.isNumNeg(args[i])); // MB: No minus nodes, the check can be simplified
-        if (l.isConstant(args[i]))
+        if (l.isConstant(args[i]) && (not l.isIte(args[i])))
```
blishko, line 755: Can it happen that a term is constant and ITE at the same time?

src/logics/LALogic.h
```diff
@@ -57,8 +58,11 @@ class LALogic: public Logic
     virtual PTRef     mkConst         (const opensmt::Number& c);
     virtual PTRef     mkConst         (const char* num);
     virtual PTRef     mkNumVar        (const char* name);
+    PTRef   mkIte                     (vec<PTRef>&) override;
+    PTRef   mkIte                     (PTRef cond, PTRef true_tr, PTRef false_tr) override { vec<PTRef> tmp{cond, true_tr, false_tr}; return mkIte(tmp); }
```
blishko, line 62: This second override is doing the same thing as the base class version. Consider removing it?

blishko (commented):
I am writing here my first batch of comments, I will check the rest of the code later.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -269,13 +269,13 @@ LVRef LASolver::exprToLVar(PTRef expr) {
         }
     }
 
-    if (logic.isNumVar(expr) || logic.isNumTimes(expr)) {
+    if (logic.isNumVar(expr) || logic.isNumTimes(expr) || logic.isIte(expr)) {
```
blishko, line 272: use isNumVarOrIte

src/tsolvers/liasolver/LIASolver.cc
```diff
@@ -35,7 +35,7 @@ void LIASolver::clearSolver()
 
 void LIASolver::notifyVar(LVRef v)
 {
-    assert(logic.isNumVar(getVarPTRef(v)));
+    assert(logic.isNumVar(getVarPTRef(v)) || logic.isIte(getVarPTRef(v)));
```
blishko, line 38: use isNumVarOrIte

test/unit/test_Ites.cc
```diff
@@ -0,0 +1,188 @@
+//
+// Created by Antti on 04.08.20.
+//
+#include <gtest/gtest.h>
+#include <Logic.h>
+#include <LRALogic.h>
+#include "../../src/itehandler/IteToSwitch.h"
```
blishko, line 7: Using just <IteToSwitch.h> should work.

test/unit/test_Ites.cc
```diff
@@ -0,0 +1,188 @@
...
+    PTRef x = logic.mkVar(boolsort, "x");
+    PTRef y = logic.mkVar(boolsort, "y");
+    PTRef cond = logic.mkEq(x, y);
+
+    vec<PTRef> args;
+    args.push(cond);
+    args.push(x);
+    args.push(y);
+    PTRef ite = logic.mkIte(args);
+    std::cout << logic.pp(ite) << endl;
```
blishko, line 77: It would be good if these tests would contain some assert, something to test that the terms are created correctly.

src/logics/Logic.cc
```diff
@@ -597,6 +586,9 @@ void Logic::simplifyTree(PTRef tr, PTRef& root_out)
 #endif
             for (int j = 0; j < t.size(); j++) {
                 PTRef cr = t[j];
+                if (isNonBoolIte(cr)) {
```
blishko, line 589: Is this the expected behaviour? What happens in LALogic when there is an equality in the condition of a non-boolean ITE? It is not a problem that such equality would not be split to conjunction of inequalities?

blishko (commented):
Here is a second batch of comments.

src/itehandler/IteToSwitch.cc
```diff
@@ -0,0 +1,283 @@
...
+        setFalseChild(false_node);
+    }
+}
+
+ite::NodeRef ite::Dag::newNode(PTRef val) {
+    assert(not nodes.has(val));
+    nodeRefs.push(na.alloc(val, PTRef_Undef, NodeRef_Undef, NodeRef_Undef));
+    nodes.insert(val, nodeRefs.last());
+    return nodeRefs.last();
+}
```
blishko, line 26: It seems this could just call newNode(val, PTRef_Undef, NodeRef_Undef, NodeRef_Undef), it would avoid code duplication.

src/itehandler/IteToSwitch.cc
```diff
@@ -0,0 +1,283 @@
...
+
+        bool unprocessed_children = false;
+
+        for (lbool childPolarity : { l_False, l_True }) {
+            NodeRef child = childPolarity == l_False ? el.getFalseChild() : el.getTrueChild();
+            if (child != NodeRef_Undef) {
+                if (parents.find(child) == parents.end()) {
+                    parents[child] = ParentSet();
+                }
+                parents[child].insert({el_r, childPolarity});
```
blishko, line 60: The check 
```
if (parents.find(child) == parents.end()) {
     parents[child] = ParentSet();
}
```
is not necessary. The operator[] on std::map takes care of that, since it always try to insert the key with default-constructed value.

src/itehandler/IteToSwitch.cc
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+    return dag;
+}
+
+void IteToSwitch::constructSwitches() {
+    vec<PTRef> ites = iteDag.getTopLevelItes();
+
+    for (auto tl_ite : ites) {
+        ite::Dag subDag = iteDag.getReachableSubGraph(logic, tl_ite);
```
blishko, line 268: This line is redundant. The local variable subDag is not used and the same creation of subgraph is repeated in the makeSwitch method.

src/itehandler/IteToSwitch.h
```diff
@@ -0,0 +1,177 @@
...
+
+    struct NodeRef {
+        uint32_t x;
+        inline friend bool operator== (NodeRef a1, NodeRef a2)   { return a1.x == a2.x; }
+        inline friend bool operator!= (NodeRef a1, NodeRef a2)   { return a1.x != a2.x; }
+        inline friend bool operator< (NodeRef a1, NodeRef a2)    { return a1.x < a2.x;  }
+    };
+    struct NodeRef_lt {
+        bool operator() (const NodeRef &n1, const NodeRef &n2) const { return n1 < n2; }
+    };
```
blishko, line 28: I think the struct NodeRef_lt is not necessary, it is doing the same thing as operator< of NodeRef.

src/itehandler/IteToSwitch.h
```diff
@@ -0,0 +1,177 @@
...
+    static struct NodeRef NodeRef_Undef = {INT32_MAX};
```
blishko, line 29: Should this be `UINT32_MAX`?

src/itehandler/IteToSwitch.h
```diff
@@ -0,0 +1,177 @@
...
+
+    struct NodeRefHash {
+        uint32_t operator () (const PTRef& s) const {
```
blishko, line 32: Why is `NodeRefHash` taking `PTRef` as an argument?

blishko (changes requested):

src/logics/LALogic.cc
```diff
@@ -184,6 +189,30 @@ lbool LALogic::retrieveSubstitutions(const vec<PtAsgn>& facts, Map<PTRef,PtAsgn,
     }
     return arithmeticElimination(top_level_arith, substs);
 }
+
+uint32_t LessThan_deepPTRef::getMin(PTRef tr) const {
```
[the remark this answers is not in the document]
aehyvari: How about getVarIdFromProduct?

src/logics/LALogic.cc
```diff
@@ -221,11 +250,12 @@ bool LALogic::isBuiltinFunction(const SymRef sr) const
...
 bool LALogic::isNumTerm(PTRef tr) const
 {
+    auto isVarOrUFOrIte = [this](PTRef tr) { return isNumVar(tr) || isUF(tr) || isIte(tr); };
     const Pterm& t = getPterm(tr);
     if (t.size() == 2 && isNumTimes(tr))
-        return ((isNumVar(t[0]) || isUF(t[0])) && isConstant(t[1])) || ((isNumVar(t[1]) || isUF(t[1])) && isConstant(t[0]));
+        return (isVarOrUFOrIte(t[0]) && isConstant(t[1])) || (isVarOrUFOrIte(t[1]) && isConstant(t[0]));
     else if (t.size() == 0)
-        return isNumVar(tr) || isConstant(tr);
+        return isNumVar(tr) || isConstant(tr) || isIte(tr);
```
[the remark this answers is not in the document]
aehyvari: That makes sense.

src/logics/LALogic.cc
```diff
@@ -307,7 +337,9 @@ bool LALogic::isNumGeq(PTRef tr) const { return isNumGeq(getPterm(tr).symb()); }
 bool LALogic::isNumGt(SymRef sr) const { return sr == get_sym_Num_GT(); }
 bool LALogic::isNumGt(PTRef tr) const { return isNumGt(getPterm(tr).symb()); }
 bool LALogic::isNumVar(SymRef sr) const { return isVar(sr) && sym_store[sr].rsort() == get_sort_NUM(); }
+bool LALogic::isNumVarOrIte(SymRef sr) const { return isVar(sr) || isIte(sr); }
```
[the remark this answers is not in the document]
aehyvari: yes.

src/logics/LALogic.cc
aehyvari: Agreed.  I fixed isNumTerm so that it should now correctly identify also ites (commit coming  soonish).

src/logics/LALogic.cc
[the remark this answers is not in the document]
blishko: Sounds much better.

src/logics/LALogic.cc
```diff
@@ -184,6 +189,30 @@ lbool LALogic::retrieveSubstitutions(const vec<PtAsgn>& facts, Map<PTRef,PtAsgn,
...
 }
+
+uint32_t LessThan_deepPTRef::getMin(PTRef tr) const {
+    PTRef c_t;
+    PTRef v_t;
+    l.splitTermToVarAndConst(tr, v_t, c_t);
+    return v_t.x;
+}
+
+bool LessThan_deepPTRef::operator()(PTRef& x_, PTRef& y_) {
```
blishko, line 200: I think this should be either `const` references, or maybe even better directly `PTRef`s, as copies, since it is basically just a 32-bit number, which should be better that passing a reference, which is (typically) 64-bit pointer.

src/logics/Logic.cc
aehyvari: This splitting will be done on the output of the ITE manager system, on the constructed switches.

src/itehandler/IteToSwitch.h
aehyvari: Seems to be dead code anyway.

test/unit/test_Ites.cc
```diff
@@ -131,10 +139,11 @@ TEST_F(IteManagerTest, test_IteTimesConst) {
     PTRef c1 = logic.mkConst("1");
     PTRef c2 = logic.mkConst("2");
     PTRef ite = logic.mkIte(cond, c1, c2);
-    PTRef prod = logic.mkNumTimes(ite, c2);
-    PTRef eq = logic.mkEq(x, prod);
-    IteToSwitch iteManager(logic, eq);
-    printTopLevelSwitches(iteManager);
+    try {
```
blishko, line 142: A better way to do this check is `EXPECT_NO_THROW`

test/unit/test_LIALogicMkTerms.h
```diff
@@ -0,0 +1,14 @@
...
+#ifndef OPENSMT_TEST_LIALOGICMKTERMS_H
+#define OPENSMT_TEST_LIALOGICMKTERMS_H
+
+
+class test_LIALogicMkTerms {
+
+};
+
+
+#endif //OPENSMT_TEST_LIALOGICMKTERMS_H
```
blishko, line 14: This header file is not used and not needed. Auto-generated?

test/unit/test_Ites.cc
```diff
@@ -4,7 +4,8 @@
 #include <gtest/gtest.h>
 #include <Logic.h>
 #include <LRALogic.h>
-#include "../../src/itehandler/IteToSwitch.h"
+#include <IteToSwitch.h>
+#include <MainSolver.h>
```
blishko, line 8: Including MainSolver does not seem to be necessary. I don't see any use of MainSolver in these tests.

blishko:
I think it looks good now, we just need to run the experiments to compare the two versions of handling boolean ITEs.

src/pterms/Pterm.h
```diff
@@ -96,6 +96,11 @@ class Pterm {
     Pterm operator= (Pterm&) = delete;
     Pterm operator= (Pterm&&) = delete;
 
+//    PTRef* begin() { return &args[0]; }
```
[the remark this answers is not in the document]
aehyvari: I guess it's a thing for another PR.

src/logics/LALogic.cc
```diff
@@ -775,18 +804,18 @@ void SimplifyConstTimes::constSimplify(const SymRef& s, const vec<PTRef>& terms,
         }
         if (!l.isNumOne(terms[i]))
         {
-            if(l.isNumPlus(terms[i]))
+            if (l.isNumPlus(terms[i]))
                 plus = terms[i];
-            else if(l.isConstant(terms[i]))
+            else if (l.isConstant(terms[i]) and not l.isIte(terms[i]))
```
blishko, line 809: We saw this before somewhere, no? It cannot happen that a PTRef is a constant and an ITE at the same time, no? So the second part of the condition is redundant.

src/logics/Logic.cc
blishko, line 589: Now that we have established that both boolean and non-boolean ITEs are treated the same way, this check seems suspicious. I think the processing should not stop also at boolean ITEs, also they are treated as variables now.

src/logics/Logic.cc
```diff
@@ -918,7 +870,7 @@ PTRef Logic::mkEq(vec<PTRef>& args) {
         return mkAnd(binaryEqualities);
     }
     assert(args.size() == 2);
-    if (isConstant(args[0]) && isConstant(args[1]))
+    if (isConstant(args[0]) && isConstant(args[1]) && (not isIte(args[0])) && (not isIte(args[1])))
```
blishko, line 873: Again the situation of constant vs ITE. So, can it happen that ITE is a constant or not?

src/logics/Logic.h
```diff
@@ -274,8 +260,8 @@ class Logic {
     PTRef       mkImpl        (PTRef _a, PTRef _b);
     PTRef       mkNot         (PTRef);
     PTRef       mkNot         (vec<PTRef>&);
-    PTRef       mkIte         (vec<PTRef>&);
-    PTRef       mkIte         (PTRef c, PTRef t, PTRef e);// { vec<PTRef> tmp; tmp.push(c); tmp.push(t); tmp.push(e); return mkIte(tmp); }
+    virtual PTRef mkIte         (vec<PTRef>&);
+    virtual PTRef mkIte         (PTRef c, PTRef t, PTRef e);// { vec<PTRef> tmp; tmp.push(c); tmp.push(t); tmp.push(e); return mkIte(tmp); }
```
blishko, line 264: It does not seem that these methods need to be virtual. No logic is overriding them and I don't see a reason why they should, now.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -269,8 +269,10 @@ void Egraph::declareTermRecursively(PTRef tr) {
     if (declared.find(tr) != declared.end()) { return; } // already declared
     const Pterm& term = logic.getPterm(tr);
     // declare first the childen and then the current term
-    for (int i = 0; i < term.size(); ++i) {
-        declareTermRecursively(term[i]);
+    if (not logic.isNonBoolIte(tr)) {
```
blishko, line 272: Also here, it seems like boolean and non-boolean ITE should now be treated the same.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -330,20 +332,24 @@ void Egraph::constructTerm(PTRef tr) {
...
-
-    else {
-        sym = enode_store.addSymb(tm.symb());
-        cdr = ERef_Nil;
-        for (int j = tm.size() - 1; j >= 0; j--) {
-            assert(enode_store.termToERef.has(tm[j])); // The child was not inserted
-            ERef car = enode_store.termToERef[tm[j]];
+    } else {
+
+        if (not logic.isNonBoolIte(tr)) {
```
blishko, line 337: Also here, it seems like boolean and non-boolean ITE should be treated the same.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -330,20 +332,24 @@ void Egraph::constructTerm(PTRef tr) {
...
-            assert (enode_store[cdr].getRoot() == cdr);
+                ERef prev_cdr = cdr;
+                assert (enode_store[car].getRoot() == car);
+                assert (enode_store[cdr].getRoot() == cdr);
 #endif
-            cdr = enode_store.addList(car, cdr);
+                cdr = enode_store.addList(car, cdr);
+            }
+        } else {
+            sym = enode_store.addIteSymb(tr);
```
blishko, line 351: I am just curious, why a separate (anonymous) symbol node is needed for each ITE term? Why a single symbol node cannot be re-used?

blishko (commented):
Just some final observations on the changes, overall I think it is good to go.

src/logics/Logic.cc
aehyvari: Yes.  Just to be sure, do we agree it's the other way round, so the processing should stop at all ITEs.

blishko: Yes, you are correct, I don't know why I wrote "not" there.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: The logic is that the ITEs are treated as variables, but an ITE with arguments $\bf{x}$ must be different from an ITE with arguments $\bf{y}$.  Therefore they must be different variables.   If a symbol node is reused, they become the same variable.

The symbol name, however, does not need to be different, since it's enough that the ERef of the symbol differs.

I'm not sure if this latter is confusing, and I could see it having some debugging benefits if each symbol would also be different, but since it's not necessary I didn't do it so far.

blishko: I think I see it. I did not realize before that you are ignoring the arguments of the ITE. Now it makes sense, thanks!

blishko (approved):
Looks good!

PR #164, opened 2020-09-02
Small refactoring of some operations in FastRationals

blishko:
This PR fixes a bug in FastRational::get_den() and unifies the style of methods `gdc, lcm, divexact, fastrat_fdiv_q` with the rest of operations over FastRationals

### DIFF
--- a/src/common/FastRational.h
+++ b/src/common/FastRational.h
@@ -196,11 +196,10 @@ private:
     friend inline void substractionAssign   ( FastRational &, const FastRational & );
     friend inline void multiplicationAssign( FastRational &, const FastRational & );
     friend inline void divisionAssign      ( FastRational &, const FastRational & );
-    friend FastRational gcd(FastRational&, FastRational&);
-    friend FastRational lcm(FastRational&, FastRational&);
-    friend FastRational fastrat_fdiv_q(FastRational& n, FastRational& d);
-    friend FastRational fastrat_fdiv_q(FastRational&& n, FastRational&& d);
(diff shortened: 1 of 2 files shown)


src/common/FastRational.h
```diff
@@ -386,15 +390,9 @@ class FastRational
         return r;
     }
 };
-// Divide n by d, forming a quotient q.
-// Rounds q down towards -infinity, and r will have the same sign as d.
-FastRational fastrat_fdiv_q(FastRational&& n, FastRational&& d);
-FastRational fastrat_fdiv_q(FastRational& n, FastRational& d);
-FastRational fastrat_round_to_int(const FastRational& n);
```
aehyvari, line 393: Maybe we should add
```FastRational fastrat_fdiv_q(const FastRational &n, const FastRational &d);```
here.  I'm a little confused about how the scoping rules work at the moment, looking at the changes needed in the unit tests.  Maybe the right choice would be to introduce a namespace for this stuff?

blishko: Right, I was a little confused about this myself. I should look again at the rules about friend functions.

blishko: This article explains it quite well: https://blog.panicsoftware.com/friends-and-where-to-find-them/
This is the first time I hear about Argument-Dependend Lookup, but apparently that's why the function can be called when I make the argument FastRational explicitly, but cannot be called otherwise (after I have removed the second declaration from outside of the class). I think the best thing to do is what you suggest, add the declaration again outside of the class in the header file.

aehyvari (approved):

PR #166, opened 2020-09-03
Tiny cleanup

blishko:
I am suggesting some small cleanup in MainSolver and Symbol.
This has been inspired by me accidentally calling MainSolver::solve method instead of MainSolver::check and getting weird results.
@aehyvari Let me know what you think about this.


aehyvari (commented):
I think it looks pretty good.  I added some const specifiers where they seemed to make sense and will commit the changes here.

The reloc field in Symbol would get useful if we ever implement deletion for symbols, but maybe we'll put it back there when that gets relevant.

blishko:
Thanks for adding the const qualifiers, it looks good.

aehyvari (approved):

PR #168, opened 2020-09-11
Some cleanup in comparing and sorting terms

### LINKED ISSUE #105: Unify sorting operations in term creation
Unify the use of LessThan_PTRef and operator < for PTRefs.
Make sure it does not break anything.

blishko:
Resolves #105 
The main change is using more natural implementation of operator< on PTRefs, which in turn renders the class LessThan_PTRef redundant.
Other changes are some related cleanup, including taking PTRefs argument in functions by value instead of by reference, since PTRefs are basically just 32-bit numbers, so potentially they can better fit in registers than 64-bit references.


blishko:
Superseded by #576.

PR #171, opened 2020-09-22
Cleanup in CMake files

blishko:
This PR contains the following changes:

- Executable is created directly in the build directory
- Libraries are created in the directory `lib` in the build directory
- The readline/editline library is no longer linked to the opensmt library, only to the executable
- Include directories of readline are passed to the opensmt targer
- The name of the library is synchronized with the name of the executable.
- For the shared library the library is created with full version name and simple versionless symlink is created as well.


CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 3.3)
-project(opensmt VERSION 2.0.0 LANGUAGES CXX)
+cmake_policy(VERSION 3.12)
```
aehyvari, line 2: At least in the cluster the cmake complains something without this line.

aehyvari (commented):
We should check the regression test scripts and the scripts for running ddsmt.

CMakeLists.txt
blishko: I think we have to be more careful and setting options per policy, and not globally.
On machine with older version of CMake it gives this error:
```
CMake Error at CMakeLists.txt:2 (cmake_policy):
 An attempt was made to set the policy version of CMake to "3.12" which is
 greater than this version of CMake.  This is not allowed because the
 greater version may have new policies not known to this CMake.  You may
 need a newer CMake version to build this project.
```

aehyvari: Ok.

src/api/CMakeLists.txt
```diff
@@ -40,29 +40,26 @@ PUBLIC ${PUBLIC_SOURCES_TO_ADD}
...
 set(OBJECTS_TO_ADD $<TARGET_OBJECTS:api> ${OBJECTS_TO_ADD})
 
-if (NOT USE_READLINE)
-  set(line_library ${LibEdit_LIBRARIES})
-else()
-  set(line_library ${Readline_LIBRARY})
-endif()
-
 if (BUILD_SHARED_LIBS)
   add_library(api_shared SHARED ${OBJECTS_TO_ADD})
```
aehyvari, line 44: Can we exclude Interpret.{cc,h} here?

blishko: As agreed, we move that change to separate PR

aehyvari (approved):

CMakeLists.txt
```diff
@@ -75,7 +76,7 @@ include_directories(
     ${CMAKE_SOURCE_DIR}/proof
     ${CMAKE_SOURCE_DIR}/models
     ${CMAKE_SOURCE_DIR}/itehandler
-    ${GMP_INCLUDE_DIR}
+    ${GMP_INCLUDE_DIRS}
```
aehyvari, line 79: Why is this in plural?  It causes problems on my mac in plural but works in singular.

blishko: I read somewhere, and [this](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html) CMake documentation confirms it, that the plural version should be the output of find_package, while the singular version is the input (a hint where to look). But maybe FindGMP module does not follow this convention.

blishko: I have pushed a fix, can you try it?

aehyvari: Works now both in cluster and my mac.

blishko:
Can we close this one and merge?

aehyvari:
Yes.

PR #177, opened 2020-09-25
Normalizing LIA inequalities with bound tightening already in Logic

blishko:
This is my attempt at creating strengthened bounds for LIA already during the term creation.
Part of it is also a refactoring of the normalization of sums in LRA Logic.

### DIFF
--- a/src/logics/LALogic.cc
+++ b/src/logics/LALogic.cc
@@ -5,6 +5,7 @@
 #include "LA.h"
 #include "LALogic.h"
 #include "FastRational.h"
+#include "OsmtInternalException.h"
 
 #include <memory>
 
@@ -134,11 +135,11 @@ PTRef LALogic::normalizeMul(PTRef mul)
     PTRef v = PTRef_Undef;
     PTRef c = PTRef_Undef;
     splitTermToVarAndConst(mul, v, c);
-    opensmt::Number r = getNumConst(c); //PS. shall I add opensmt::Integer j = getNumConst(c)
-    if (r < 0) //PS. OR (l < 0)
+    if (getNumConst(c).sign() < 0) {
         return mkNumNeg(v);
-    else
+    } else {
         return v;
+    }
 }
 lbool LALogic::arithmeticElimination(const vec<PTRef> & top_level_arith, Map<PTRef, PtAsgn, PTRefHash> & substitutions)
 {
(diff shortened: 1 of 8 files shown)


blishko:
@aehyvari Can you check the code? I will try to run experiments to see what effect this has in LRA/LIA benchmarks.

blishko:
Comparison with master shows almost no change in LRA and positive change in LIA (within 600 seconds timeout, number of solved benchmarks rises to 4789 from 4714.

src/logics/LIALogic.cc
```diff
@@ -162,4 +162,90 @@ const SymRef LIALogic::get_sym_Num_EQ () const {return sym_Int_EQ;}
...
+            constant = s[i];
+        } else {
+            assert(isLinearFactor(s[i]));
+            varFactors.push(s[i]);
+        }
+    }
+    if (constant == PTRef_Undef) { constant = getTerm_NumZero(); }
+    auto constantValue = getNumConst(constant);
+    termSort(varFactors);
+    vec<PTRef> vars; vars.capacity(varFactors.size());
```
aehyvari, line 183: vec\<PTRef\> vars(varFactors.size());

src/logics/LIALogic.cc
```diff
@@ -162,4 +162,90 @@ const SymRef LIALogic::get_sym_Num_EQ () const {return sym_Int_EQ;}
...
+    std::vector<opensmt::Number> coeffs; coeffs.reserve(varFactors.size());
+    for (PTRef factor : varFactors) {
+        PTRef var;
+        PTRef coeff;
+        splitTermToVarAndConst(factor, var, coeff);
+        assert(isNumVarOrIte(var) and isNumConst(coeff));
+        vars.push(var);
+        coeffs.push_back(getNumConst(coeff));
+    }
```
aehyvari, line 192: Maybe introduce a lambda for this as well? (as below)

aehyvari (changes requested):
It looks good, and the performance improvement is great news.  Does this change have implications for LIASolver?  We could have some unit test out of principle, I'll commit a suggestion.  Feel free to remove it if you disagree of course.

test/unit/test_LIAStrengthening.cc
```diff
@@ -0,0 +1,27 @@
...
+
+TEST_F(LIAStrengthening, test_LIAStrengthening) {
+    PTRef c2 = logic.mkConst(2);
+    PTRef c3 = logic.mkConst(3);
+    PTRef x = logic.mkNumVar("x");
+    PTRef y = logic.mkNumVar("y");
+    PTRef sum = logic.mkNumTimes(c2, logic.mkNumPlus(x, y));
+    PTRef ineq = logic.mkNumGeq(sum, c3);
+    const Pterm& ineq_tm = logic.getPterm(ineq);
+    ASSERT_EQ(ineq_tm[0], logic.mkConst("2"));
```
aehyvari, line 26: This makes me think that would it be any useful outside of this context to have methods `PTRef LALogic::getConstant(PTRef)` and  `PTRef LALogic::getSum(PTRef)` defined for a term for which `LALogic::isNumLeq` is true?

blishko: That sounds useful.

src/logics/LIALogic.cc
blishko: I am not sure what you mean. Maybe you can explain it to me in person.

src/logics/LIALogic.cc
blishko: If I do that, then it will default-construct the given number of elements and then in the loop I cannot use `push` but I have to re-assign the elements, correct? Is that better then reserving just the capacity and then pushing the elements?

aehyvari: I think it's good like you proposed.  Ignore that one.

test/unit/test_LIAStrengthening.cc
blishko: I have tried to add some helper methods for that, let me know what you think.

src/logics/LALogic.cc
```diff
@@ -995,6 +972,42 @@ LALogic::printTerm_(PTRef tr, bool ext, bool safe) const
...
+    PTRef leadingFactor = varFactors[0];
+    // normalize the sum according to the leading factor
+    PTRef var, coeff;
+    splitTermToVarAndConst(leadingFactor, var, coeff);
+    opensmt::Number normalizationCoeff = abs(getNumConst(coeff));
+    // varFactors come from a normalized sum, no need to call normalization code again
+    PTRef normalizedSum = varFactors.size() == 1 ? varFactors[0] : insertTermHash(get_sym_Num_PLUS(), varFactors);
+    if (normalizationCoeff != 1) {
+        // normalize the whole sum
+         normalizedSum = mkNumTimes(normalizedSum, mkConst(normalizationCoeff.inverse()));
```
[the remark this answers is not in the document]
blishko: Done

aehyvari (approved):
I think it looks good now.

PR #180, opened 2020-09-29
std::vector to minisat vec conversion

aehyvari:
This PR moves the code used in the Logic API `std::vector` =>  minisat `vec` conversion  directly to minisat `vec`'s constructor.

### DIFF
--- a/src/logics/LALogic.h
+++ b/src/logics/LALogic.h
@@ -49,8 +49,8 @@ protected:
 public:
     LALogic();
     ~LALogic() { for(int i = 0; i < numbers.size(); ++i) {delete numbers[i];}}
-    virtual bool     isBuiltinFunction(SymRef sr) const override;
-    virtual PTRef    insertTerm       (SymRef sym, vec<PTRef>& terms, char** msg) override;
+    bool             isBuiltinFunction(SymRef sr) const override;
+    PTRef            insertTerm       (SymRef sym, vec<PTRef>& terms) override;
     virtual SRef     getSort_num      () const;
     virtual PTRef    mkConst          (const char* name, const char **msg) override;
     virtual PTRef    mkConst          (SRef s, const char* name) override;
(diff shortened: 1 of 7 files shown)


src/logics/Logic.h
```diff
@@ -444,6 +442,7 @@ class Logic {
     PTRef       resolveTerm        (const char* s, vec<PTRef>& args, char** msg);
     // XXX There's a need for non msg giving version
     virtual PTRef       insertTerm         (SymRef sym, vec<PTRef>& terms, char** msg);
+    virtual PTRef       insertTerm         (SymRef sym, vec<PTRef>&& terms) { return insertTerm(sym, terms, nullptr); }
```
blishko, line 445: It looks like insertTerm is not using the msg parameter (in the actual implementation), so maybe we can the parameter from the first signature?

src/logics/LALogic.h
```diff
@@ -49,8 +49,8 @@ class LALogic: public Logic
 public:
     LALogic();
     ~LALogic() { for(int i = 0; i < numbers.size(); ++i) {delete numbers[i];}}
-    virtual bool     isBuiltinFunction(SymRef sr) const override;
-    virtual PTRef    insertTerm       (SymRef sym, vec<PTRef>& terms, char** msg) override;
+    bool             isBuiltinFunction(SymRef sr) const override;
+    PTRef            insertTerm       (SymRef sym, vec<PTRef>& terms) override;
```
blishko, line 53: I would keep the virtual keyword here, since it is present also for other overriden methods (below).

blishko (approved):
Looks good!

src/logics/LALogic.h
aehyvari: Or we could remove it from below as well, when it's redundant?

blishko: Or that.
But this is one of the places where I consider something redundant to be useful. But maybe I am just too used to virtual functions being marked as `virtual` :)

blishko:
I have added my small change since this PR is cleaning the method interfaces in Logic.

PR #183, opened 2020-10-07
Refactoring of interpolator for LRA

blishko:


src/tsolvers/lrasolver/FarkasInterpolator.cc
```diff
@@ -630,10 +570,144 @@ PTRef LRA_Interpolator::getInterpolant(icolor_t color, PartitionManager &pmanage
...
+bool FarkasInterpolator::isALocal(PTRef var) const {
     return isAstrict(pmanager.getIPartitions(var), mask);
 }
 
-bool LRA_Interpolator::isBLocal(PTRef var, PartitionManager &pmanager) const {
+bool FarkasInterpolator::isBLocal(PTRef var) const {
     return isBstrict(pmanager.getIPartitions(var), mask);
 }
+
+PTRef FarkasInterpolator::getInterpolant(FarkasItpOptions const & options) {
```
aehyvari, line 581: Is there a way of giving this function a bit more structure?  Separate the different ITP algorithm implementations to different functions, with those having a lot of code in common maybe further split into other functions?  It's a bit painful to read as it is now.

src/tsolvers/lrasolver/FarkasInterpolator.h
```diff
@@ -0,0 +1,122 @@
...
+    static FarkasItpOptions useFarkasAlgorithm() { return FarkasItpOptions(ItpAlg::STRONG); }
+    static FarkasItpOptions useDualFarkasAlgorithm() { return FarkasItpOptions(ItpAlg::WEAK); }
+    static FarkasItpOptions useFlexibleFarkasAlgorithm(FastRational factor) { return FarkasItpOptions(ItpAlg::FACTOR, std::move(factor)); }
+    static FarkasItpOptions useDecomposingFarkasAlgorithm() { return FarkasItpOptions(ItpAlg::DECOMPOSING); }
+    static FarkasItpOptions useDualDecomposingFarkasAlgorithm() { return FarkasItpOptions(ItpAlg::DECOMPOSING_DUAL); }
+
+};
+
+class FarkasInterpolator {
+public:
```
aehyvari, line 67: Should this be a child class of `src/smtsolvers/TheoryInterpolator`?

src/tsolvers/lrasolver/FarkasInterpolator.h
```diff
@@ -0,0 +1,122 @@
...
+
+    PTRef getInterpolant(FarkasItpOptions const &);
+
+    static DecomposedStatistics stats;
+
+private:
+
+    PTRef getDecomposedInterpolant(icolor_t color);
+
+    bool isLocalFor(icolor_t color, PTRef var) const{
```
aehyvari, line 86: Maybe could be refactored to a parent class?

src/tsolvers/lrasolver/LRASolver.cc
```diff
@@ -96,225 +96,28 @@ lbool LRASolver::getPolaritySuggestion(PTRef ptref) const {
...
-                delta_flag_dual=true;
-            }
-            else
-            {
-                interpolant_dual.addExprWithCoeff(LAExpression(logic, explanation[i].tr, false), -explanationCoefficients[i]);
-            }
-        }
-    }
+    }();
+    return interpolator.getInterpolant(options);
```
aehyvari, line 120: This function has a similar check to the above lambda.  Maybe with refactoring as I commented above on splitting the long function, or alternatively with a factory approach that returns an object with a derived type, one could avoid repeating the code?

aehyvari (commented):
This looks quite good.  I wrote some high-level comments as a basis for discussions, I'm not  sure how clear they are.

src/common/VerificationUtils.h
```diff
@@ -17,6 +17,11 @@ class VerificationUtils {
     bool implies(PTRef, PTRef); // Check the result with an external solver
 
     bool verifyInterpolant(PTRef partA, PTRef partB, PTRef itp);
+
+    bool verifyInterpolantInternal(PTRef partA, PTRef partB, PTRef itp);
+
+private:
+    bool checkSubsetCondition(PTRef p1, PTRef p2);
```
aehyvari, line 24: This is in general a good idea.  Maybe we should change the others to have the word `external` in them?

aehyvari (approved):
I think this looks like a pretty good step to the right direction.

src/tsolvers/lrasolver/FarkasInterpolator.h
aehyvari: Can be done at some later point.

src/common/VerificationUtils.h
blishko: Good idea, I will do that.

PR #184, opened 2020-10-11
updating CMake files to use modern patterns and to provide package files

### LINKED ISSUE #165: Add support for installing CMake file.
To make it easier to include OpenSMT as a library in other CMake projects using a simple `find_library(opensmt)` command.

blishko:
Resolves #165


blishko:
It would be good to decide on the spelling of opensmt CMake targets. The current suggetion is `opensmt`, shall we change it to `OpenSMT`?

blishko:
The CI is passing now, so the new package works for OpenSMT examples. I will test if it can be used also from HiFrog and Sally.

CMakeLists.txt
```diff
@@ -77,7 +77,6 @@ include_directories(
     ${CMAKE_SOURCE_DIR}/proof
     ${CMAKE_SOURCE_DIR}/models
     ${CMAKE_SOURCE_DIR}/itehandler
-    ${GMP_INCLUDE_DIRS}
```
aehyvari, line 80: A build in my OS X with `GMP` installed from homebrew fails with the current commit, but passes if I add here ${GMP_INCLUDE_DIR}.  I wonder if the same would happen in some systems with `libreadline` or `libedit`.  Would it be ok to add ${GMP_INCLUDE_DIR} here at least?

aehyvari (changes requested):
Is it because of having the target now name opensmt-bin?  I'm ok with the name OpenSMT as well, but I don't mind just opensmt.

CMakeLists.txt
blishko: Hmm, I would like NOT to do that. One of the points for these changes is that exactly things like that should not be necessary because it should be done automatically.
Where do you get a compilation error?

aehyvari: I see, I was just reading the documentation of cmake and it looks like that would not be the right way to go.  Anyway, to answer your question, it's my home laptop, where I'm running Mac OS X 10.15.2 and the cmake build system finds (correctly) `/usr/local/include` as the GMP_INCLUDE_DIR.

blishko: Hmm, at my laptop it was working and I have GMP installed in the same directory. But I think I have older version of Mac OS.
What does the compilation error say?

aehyvari: I managed to make this compile by adding `include_directories(${GMP_INCLUDE_DIR})` to `cmake_modules/FindGMP.cmake:32`.  Do you know if this is the correct way?

cmake_modules/FindGMP.cmake
```diff
@@ -3,17 +3,31 @@
...
+        IMPORTED_LOCATION "${GMPXX_LIBRARY}"
+        #INTERFACE_INCLUDE_DIRECTORIES "${GMP_INCLUDE_DIR}"
+    )
 
-mark_as_advanced(GMP_INCLUDE_DIRS GMP_LIBRARIES GMPXX_LIBRARIES)
+    add_library(GMP::GMP INTERFACE IMPORTED)
+    set_property(TARGET GMP::GMP APPEND PROPERTY
+        INTERFACE_LINK_LIBRARIES GMP::GMP_C GMP::GMP_CXX
+        )
+    # the above command is for compatibillity with old versions of CMAKE, the version below is simpler, but works only from CMake 3.11
```
aehyvari, line 31: Referencing the above comment, add `include_directories(${GMP_INCLUDE_DIR})` somewhere here?

CMakeLists.txt
aehyvari: Here's the error.

```
Anttis-MacBook-Pro:build anttihyvarinen$ make
[  0%] Building CXX object src/proof/CMakeFiles/proof.dir/PGBuild.cc.o
In file included from /Users/anttihyvarinen/src/opensmt/src/proof/PGBuild.cc:20:
In file included from /Users/anttihyvarinen/src/opensmt/src/proof/PG.h:23:
/Users/anttihyvarinen/src/opensmt/src/common/Global.h:36:10: fatal error:
      'gmp.h' file not found
#include <gmp.h>
         ^~~~~~~
1 error generated.
make[2]: *** [src/proof/CMakeFiles/proof.dir/PGBuild.cc.o] Error 1
make[1]: *** [src/proof/CMakeFiles/proof.dir/all] Error 2
make: *** [all] Error 2
```

blishko: What you suggests might be the right solution. 
Let me have a look at the whole problem again and if I don't find other solution soon, we can go with this one.

Do you think you could send me a log of running `make VERBOSE=1`?

cmake_modules/FindGMP.cmake
blishko: I think I have an idea why the compilation might break in your case, although it is interesting that I was able to build OpenSMT in the cluster, with non-standard GMP location with no problems.
I think we should re-design how the main library is created. When I started I used this approach that each folder is compiled separately to a bunch of object files collected in an object library and later all these object files are linked together. Now I don't like this approach very much.
I am not sure if we can identify separate modules (targets) in OpenSMT that (at least in theory) would form standalone units without cyclic dependencies. Or we could keep just one main target and collect only source files from directories instead of object files. But I think I ran into a problem of how to create both static and shared library in that case.

Maybe we can wait with this PR a little bit, I will try to figure out something.

blishko: I believe I have solved the problem to my liking. You can try compiling it again. Unfortunately, the CI has not run for the new commits, maybe after I rebase. Hopefully I have not missed anything.

blishko:
> Is it because of having the target now name opensmt-bin? I'm ok with the name OpenSMT as well, but I don't mind just opensmt.

I am now more inclined to use the name `OpenSMT`.

The problem is that I need to create a CMake target for the executable and for the library and these need to be distinct.
Since the library target is the one that can be included in other projects, I thought it would be easier to name that one `opensmt` (so users can include it like `opensmt::opensmt` rather than `opensmt::libopensmt` or something similar) and give the executable target a separate name, since we can specify what should be the name of the actual executable file in the end.

blishko:
@aehyvari I have added the POLICY for using <package>_ROOT variables to the CMakeLists of the examples (and removed setting the module path which should not be necessary anymore). Can you check again the compilation?

aehyvari:
Looks all fine, compiles with the GMP_ROOT variable set.

CMakeLists.txt
aehyvari: As discussed, this is now resolved.

cmake_modules/FindGMP.cmake
aehyvari: So far everything looks good to me.

src/api/CMakeLists.txt
```diff
@@ -40,25 +37,60 @@ PUBLIC ${PUBLIC_SOURCES_TO_ADD}
...
+  include(CMakePackageConfigHelpers)
+  write_basic_package_version_file( "${CMAKE_CURRENT_BINARY_DIR}/OpenSMTConfigVersion.cmake"
+    VERSION ${OPENSMT_VERSION}
+    COMPATIBILITY SameMajorVersion # SameMinorVersion available only from CMake 3.11 
+    )
+
+configure_package_config_file(OpenSMTConfig.cmake.in
+    "${CMAKE_CURRENT_BINARY_DIR}/OpenSMTConfig.cmake"
+    INSTALL_DESTINATION lib/cmake/opensmt
+    #PATH_VARS INCLUDE_INSTALL_DIR INSTALL_DIR
```
aehyvari, line 71: Consider removing unless this is useful once we drop support for legacy cmake.

aehyvari (commented):
It looks very good.  I have some not critical comments there.  Do you think we could rewrite the test/unit/CMakeLists.txt in the same spirit as the one for examples/CMakeLists.txt so we wouldn't have to explicitly list all the tests?  The number grows all the time..

aehyvari (approved):

src/api/CMakeLists.txt
blishko: I don't remember, most likely it can be removed.

cmake_modules/FindGMP.cmake
```diff
@@ -3,17 +3,36 @@
...
+if(GMP_FOUND AND NOT TARGET GMP::GMP)
+    add_library(GMP::GMP_C UNKNOWN IMPORTED)
+    set_target_properties(GMP::GMP_C PROPERTIES
+        IMPORTED_LOCATION "${GMP_LIBRARY}"
+        INTERFACE_INCLUDE_DIRECTORIES "${GMP_INCLUDE_DIR}"
+    )
+    add_library(GMP::GMP_CXX UNKNOWN IMPORTED)
+    set_target_properties(GMP::GMP_CXX PROPERTIES
+        IMPORTED_LOCATION "${GMPXX_LIBRARY}"
+        #INTERFACE_INCLUDE_DIRECTORIES "${GMP_INCLUDE_DIR}"
```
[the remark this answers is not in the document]
blishko: In current version, no, since the include directory is already remembered in the GMP::GMP_C.

Issue #185, opened 2020-10-12
Assertion error at tsolvers/egraph/EnodeStore.cc:175


rainoftime:
Hi, for the formula below
opensmt 
e0d849599
~~~~
(set-logic QF_UF)
(declare-fun uf7 (Bool Bool Bool Bool Bool Bool Bool) Bool)
(declare-fun v7 () Bool)
(declare-fun v8 () Bool)
(assert (uf7 true true true (not (distinct v7 v8)) true true true))
(check-sat)
(exit)

opensmt xx.smt2
opensmt: /home/peisen/test/tofuzz/opensmt/src/tsolvers/egraph/EnodeStore.cc:175: vec<std::pair<PTRef, ERef> > EnodeStore::constructTerm(PTRef): Assertion `termToERef.has(tm[j])' failed.
Aborted
~~~~

blishko:
It seems this problem has been in OpenSMT for quite some time and it has not been caused by recent refactoring as I originally thought.
Here is the minimized version:
```
(set-logic QF_UF)
(declare-fun f (Bool) Bool)
(declare-fun v7 () Bool)
(declare-fun v8 () Bool)
(assert (f (= v7 v8)))
(check-sat)
(exit)
```

aehyvari:
This should be fixed now.  Thank you!

PR #186, opened 2020-10-21
Recursive and non-recursive enodes

aehyvari:
This PR fixes issue #185 where the problem seems to have been that some equalities need a recursive definition in UF whereas others can be treated as uninterpreted predicates.  For now I've convinced myself that this fix works with incrementality as well.  While fixing this I refactored EnodeStore::constructTerm.

In addition the PR contains an implementation for begin() and end() for Pterms so that they can be used in range-based for-loops.

I ran this against the commit e0d84959, there seems to be no impact on performance. 

non-incremental-QF_UF-e0d84959-2020-10-21_vs_issue185-2020-10-21


test/unit/test_LogicMkTerms.cc
```diff
@@ -118,3 +119,32 @@ TEST_F(LogicMkTermsTest, testMkAndContradiction) {
...
+        int count = 0;
+        for (auto ch : logic.getPterm(tr)) {
+            (void)ch;
+            count++;
+        }
+        ASSERT_EQ(count, 0);
+    }
+
+    for (auto tr : {conj, disj}) {
+        Pterm &t = logic.getPterm(tr);
```
blishko, line 140:
```suggestion
        Pterm const &t = logic.getPterm(tr);
```

test/unit/test_LogicMkTerms.cc
```diff
@@ -3,6 +3,7 @@
 //
 #include <gtest/gtest.h>
 #include <Logic.h>
+#include <EnodeStore.h>
```
blishko, line 6: Redundant?

src/tsolvers/egraph/EnodeStore.cc
```diff
@@ -142,58 +180,45 @@ vec<std::pair<PTRef,ERef>> EnodeStore::constructTerm(PTRef tr) {
...
+
+    ERef er = addTerm(sym, cdr, tr);
+    new_enodes.push({tr, er});
 
-        // If tr is a complex Boolean operator, do not model the full logic but cut here (the ERef
-        // will be treated as a UF constant with the anon name from the logic).  Otherwise, (tr is
-        // a pure Boolean atom or its negation), use the term ref from the logic.
-        ERef sym = logic.isBooleanOperator(tr_pure) ? addSymb(logic.getSym_anon()) : addSymb(logic.getSymRef(tr_pure));
+    if (logic.hasSortBool(tr)) {
+        // Add both the pure and the negated terms
```
blishko, line 210: Now in this `if` you are only adding the negated term, right? The pure term has already been added at this point, no?

src/pterms/Pterm.h
```diff
@@ -125,7 +125,9 @@ class Pterm {
     void     setId(int i)   { id.x = i; }
 
     void     shrink(int s)   { header.size -= s; }
-    void     copyTo(Pterm& to);
+
+    const PTRef* begin() const { return args; }
+    const PTRef* end() const { return args + size(); }
```
blishko, line 130: This is very useful! 
I would just like to add somewhere a big warning that using the range-based for-loop is safe only when no new term can be created in the body of the loop. Otherwise we could be surprised by some unexpected crashes.

blishko (commented):
I am not sure I understand what you mean by recursive definition here. When exactly a term does or does not need a node in ENodeStore?
But overall, it looks good to me! I just wrote a few comments.

test/unit/test_LogicMkTerms.cc
aehyvari: Yes.

src/pterms/Pterm.h
aehyvari: Yes, that's true, and I suppose we both have been there...  It is true also for the traditional for-loop.  Would you have any ideas how to implement such a warning?  Were you just thinking of a comment, or is there something else that could be done?

blishko: I meant just a comment, I don't know about anything else that we could do to enforce this.

aehyvari:
> I am not sure I understand what you mean by recursive definition here. When exactly a term does or does not need a node in ENodeStore?

A very good question.   The conditions under which a term needs an `enode` are given in `bool EnodeStore::needsEnode(PTRef tr) const`.  However, once it's decided that the term needs an `enode`, there's still the question whether the children of the term should be represented in this `enode`, or should the `enode` be just an anonymous symbol.  The conditions for this are given in `bool EnodeStore::needsRecursiveDefinition(PTRef tr) const`.

src/tsolvers/egraph/EnodeStore.cc
aehyvari: Yes that's true.  I forgot to update the comment.

aehyvari:
There are also some examples in `test/unit/test_EnodeStore.cc` that try to cover the cases where enodes are needed.

src/pterms/Pterm.h
aehyvari: It would be nice to have a feature in c++ where you can declare a variable to be const within a given scope, like you can effectively do for functions.

blishko:
Thanks for the explanations! I have also tried a couple of examples, and I was not able to break anything, so I think it's good!

blishko (approved):
Thanks, looks good!

PR #189, opened 2020-10-24
Improved bookkeeping on UF appearances of boolean terms

aehyvari:
This PR implements two fixes in handling of Boolean atoms inside uninterpreted functions.

### Bookkeeping

The PR improves the bookkeeping code for appearances of Boolean atoms inside uninterpreted functions.  In current master it is possible that a Boolean term appears (directly) in a UF term at term construction time, but the substitution code that rewrites terms based on equalities at the top level of a formula removes these appearances.  This information is not correctly updated to the UF theory solver, breaking an invariant of a code and resulting in a crash.

The proposed fix is to keep track of whether a boolean term appears currently, has appeared but no longer appears, or has never appeared, in UF terms.

#### Notes

The terms appearing in UF are recorded currently also in the vector [`vec<PTRef> Logic::propFormulasAppearingInUF`](https://github.com/usi-verification-and-security/opensmt/blob/6617993836751d1f1aaf7e2e238513e611d2eb62/src/logics/Logic.h#L169).  The code guarantees that the vector doesn't contain duplicates.  This vector is used for introducing the required atoms to the SAT solver in Cnfizer [here](https://github.com/usi-verification-and-security/opensmt/blob/6617993836751d1f1aaf7e2e238513e611d2eb62/src/cnfizers/Cnfizer.cc#L214).  It should be investigated if this is necessary also for the atoms that were re-written (I assume yes).  At the moment the code ensures these are inserted. 

### Polarity Conflicts

The PR implements the handling of cases where the UF solver determines that the terms p and (not p) would need to be in the same equivalence class.  This special case is not correctly dealt with in the current master.  The PR does this by generalising the ForbidList's reason to a struct that contains information whether the forbid is due explicitly to (1) a negated equality or a disequality (marked `Expl::Type::std`), to the elements being two distinct constants (`Expl::Type::cons`), or to elements being each others' boolean negation (`Expl::Type::pol`).

#### Notes

While looking at the code it felt like the size of the `Expl` structure could be made smaller, from the current 3 words to a single one.  This extends to making `PtAsgn` smaller.  Right now the increased size of `Expl` doesn't cause a problem as can be seen from the attached performance plot against current master.

non-incremental-QF_UF-master-2020-10-29_vs_polarity_conflicts-2020-10-28


aehyvari:
In current form the PR fixes the [second example](https://github.com/usi-verification-and-security/opensmt/issues/190#issuecomment-716104209) in Issue #190.

aehyvari:
The last commit fixes also the first example in Issue #190

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -249,27 +249,26 @@ void Egraph::declareTermRecursively(PTRef tr) {
...
- * - In general the arguments of the term have UF representation.
- * - However, if the term itself is a Boolean operator, its arguments
- *   will not have enodes as this case will be handled separately by
- *   the SAT solver
  *
  * For the Boolean terms (that appear as arguments in UF), we add also
  * their negations.  This means that we need the negation uninterpreted
- * function.
+ * function.  In this Boolean case we add that a term is not equal to
+ * its negation, with the reason true
```
blishko, line 266:
```suggestion
 * its negation, with the reason true.
```

src/tsolvers/egraph/Egraph.h
```diff
@@ -334,22 +334,24 @@ class Egraph : public TSolver {
...
 
-    bool    assertEq        ( PTRef, PTRef, PtAsgn );               // Asserts an equality
-    bool    assertEq        ( ERef, ERef, PtAsgn );                 // Called by the above
-    bool    assertNEq       ( PTRef, PTRef, PtAsgn );               // Asserts a negated equality
-    bool    assertNEq       ( ERef, ERef, PtAsgn );                 // Called by the above
-    bool    assertDist      ( PTRef, PtAsgn );                      // Asserts a distinction
+    bool    assertEq        (PTRef, PTRef, PtAsgn);               // Asserts an equality
+    bool    assertEq        (ERef, ERef, PtAsgn);                 // Called by the above
+    bool    assertNEq       (PTRef, PTRef, const Expl &r);               // Asserts a negated equality
+    bool    assertNEq       (ERef, ERef, const Expl &r);                 // Called by the above
```
blishko, line 344: Indentation of the comments.

src/tsolvers/egraph/Enode.h
```diff
@@ -277,6 +277,20 @@ class EnodeAllocator : public RegionAllocator<uint32_t>
...
+/**
+ * Expl is a generalization of an explanation inequality that covers explicitly
+ * the standard inequality consisting of a negated equality or inequality,
+ * difference of two constants, and a boolean atom and its negation being distinct.
+ */
+struct Expl {
+    enum class Type { std, cons, pol, undef } type;
+    PtAsgn pta;
+    PTRef pol_term;
+    Expl() : type(Type::undef), pta(PtAsgn_Undef), pol_term(PTRef_Undef) {}
```
blishko, line 290: @aehyvari , you mentioned in the description that you think this struct should be smaller? I am not sure how to make it one-word size, but it seems to me that `pta` member is valid only for type `std` and `pol_term` is valid only for `pol`, right? Maybe they can be put into a `union`?

blishko (commented):
Looks quite good to me! I have just some questions and small observations.

blishko:
So, there are two separate issues here, right?
One is that the booleans in UF can disappear during simplifications and this somehow caused a problem in incremental setting?

And the second issue was that the conflict when `p` and `not p` are discovered to be in the same equivalence class was not detected before?

By the way, very professional description of the PR, I should work on mines :)

src/logics/Logic.cc
```diff
@@ -1356,8 +1355,10 @@ bool Logic::varsubstitute(PTRef root, const Map<PTRef, PtAsgn, PTRefHash> & subs
 #ifdef SIMPLIFY_DEBUG
                     printf("  %s -> %s\n", printTerm(t[i]), printTerm(gen_sub[t[i]]));
 #endif
+                    if (hasSub and appearsInUF(t[i])) {
+                        decreaseUFAppearanceCount(t[i]);
+                    }
```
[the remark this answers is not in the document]
aehyvari: > Is it correct to always decrement here? What if such term is being substituted in a non-UF context? Or that cannot happen?

I think it can happen.  In my quick tests I could verify this.  I'll add a check there that `isUF(tr)`, I hope that solves the issue.

> And if the count is decremented for the child here, shouldn't it be incremented for the term that is being substituted in its place?

This should already happen when the new term is constructed with the standard means, if I get what you mean.  I'll maybe have to come up with an example to verify this.

src/tsolvers/egraph/Enode.h
aehyvari: Yes.  I was thinking of packing the information in `Expl::Type` to the bits of `PtAsgn`, and packing `PtAsgn` so that two bits are reserved for the polarity (`l_True`, `l_False`, `l_Undef`).  But since there is no effect on performance in my plot, I'm planning to postpone this to a different PR.

aehyvari:
> So, there are two separate issues here, right?
> One is that the booleans in UF can disappear during simplifications and this somehow caused a problem in incremental setting?
> 
> And the second issue was that the conflict when `p` and `not p` are discovered to be in the same equivalence class was not detected before? 

Yes, exactly.

> 
> By the way, very professional description of the PR, I should work on mines :)

Hehe, I think they're good :)

src/tsolvers/egraph/Enode.h
blishko: OK, makes sense.

src/logics/Logic.cc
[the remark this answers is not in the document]
blishko: > This should already happen when the new term is constructed with the standard means, if I get what you mean. 

Yes, you are right. That's fine then.

> I think it can happen. In my quick tests I could verify this. I'll add a check there that isUF(tr), I hope that solves the issue.

Maybe this is not necessary, since the substitution should actually remove the term from everywhere, so maybe this is a not a problem. But I probably still don't understand what the original issue with these substitutions was, so don't mind me :)

blishko:
One thing that came to my mind: Recently we were trying to separate the Logic and the MainSolver, so the Logic would be re-usable for multiple solvers. However, these `booleansInUF` are still stored in Logic and this goes against the philosophy above, since they are local to the query. Maybe we should try to find a different way how to keep this information? (Maybe a topic for separate PR.)

src/logics/Logic.cc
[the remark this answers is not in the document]
aehyvari: > the substitution should actually remove the term from everywhere

I think that's not true, the substitution is not on the term that appears in UF, but on something else that results in a uf-appearance disappearing.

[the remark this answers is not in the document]
aehyvari: Now I agree with what you say above, and disagree what I wrote as a reply.  Maybe this was not the original issue, but instead the problem was somehow related to whether a term has ever appeared in UF or if it has appeared at some point but right now does not appear.  I think I'll try it that way, it might make the code a little more compact.

blishko (approved):
Any more changes planned? To me it looks good, at least from my superficial understanding of the situation.

aehyvari:
nope, that's it.

blishko:
Do you want to merge it, or shall I?

aehyvari:
I can merge. Did you accept it?

blishko:
Yes I did.

PR #192, opened 2020-11-02
Refactoring of colouring information in Theory Interpolation

blishko:
This PR is a follow-up of #183 and aims to better define how information about partitioning of terms is passed to the Theory Interpolators. 

This partitioning information can be global, derived from the partitioning of the whole SMT query, or it can be local, derived just from the partitioning of the theory conflict passed to the interpolator.
These can be different, as terms that are shared globally across partitions can be local within a single theory conflict.

This PR includes just refactoring of the current code, with no change in the functionality, but it better separates the Interpolator classes from the rest of the code, which allows for easier unit testing.

In the future we could also change how the partitioning information is computed---the global partitioning information can be delayed until the interpolation phase and does not need to be maintained fully from the beginning of the solving process.

### DIFF
--- a/src/smtsolvers/TheoryInterpolator.h
+++ b/src/smtsolvers/TheoryInterpolator.h
@@ -3,7 +3,139 @@
 
 #include "Global.h"
 #include "PtStore.h"
-#include <PartitionManager.h>
+#include "PartitionManager.h"
+#include "OsmtInternalException.h"
+
+#include <memory>
+#include <algorithm>
+
+class TermColorInfo {
+public:
+    virtual icolor_t getColorFor(PTRef term) = 0;
+    virtual ~TermColorInfo() = default;
+
+};
+
+class GlobalTermColorInfo : public TermColorInfo {
+public:
+    GlobalTermColorInfo(PartitionManager & pmanager, ipartitions_t mask) : pmanager(pmanager), mask(std::move(mask)) {}
+
+    icolor_t getColorFor(PTRef term) override {
+        auto const & termMask = pmanager.getIPartitions(term);
+        auto res = getColorForMask(termMask);
+        if (res == I_UNDEF) {
+            throw OsmtInternalException("No color detected for term");
+        }
+        return res;
+    }
+
+private:
+    PartitionManager & pmanager;
+    ipartitions_t mask;
+
+    icolor_t getColorForMask(ipartitions_t const & otherMask) {
+        bool isInA = (otherMask & mask) != 0;
+        bool isInB = (otherMask & ~mask) != 0;
+        if (isInA and not isInB) { return I_A; }
+        if (isInB and not isInA) { return I_B; }
+        if (isInA and isInB) { return I_AB; }
+        return I_UNDEF;
+    }
+};
+
+/*
+ * Stores color information for a set of terms given the colors of top-level term.
+ *
+ * Terms can be A-local, B-local, or AB-shared.
+ * Note: If a term f(x) is local, but both the function symbol and all the arguments are AB-shared,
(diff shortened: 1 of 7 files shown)


src/smtsolvers/TheoryInterpolator.h
```diff
@@ -3,7 +3,139 @@
...
+        // MB: NOTE! If P(a) is A-local, but both symbols P and a are shared, than P(a) should be shared and not A-local
+        using entry_t = std::pair<const PTRef, icolor_t>;
+        auto colorUnion = [](icolor_t f, icolor_t s) { return static_cast<opensmt::icolor_t>(f | s); };
+        std::vector<entry_t> queue;
+        for (auto entry : topLevelColors) {
+            queue.push_back(entry);
+        }
+        std::unordered_map<SymRef, icolor_t, SymRefHash> symbolColors;
+        while (not queue.empty()) {
+            auto const & entry = queue.back();
```
aehyvari, line 85: Maybe the reference is not necessary here?

blishko: You would prefer taking it by value? Or something like
```
PTRef term = queue.back().first;
icolor_t colorToAssign = queue.back().second;
```
?

aehyvari:
Yes I thought by value. Looks like the sizes are anyway 2 * 32 bits which
is the same as the address.

blishko:
OK, I can change it.

src/smtsolvers/TheoryInterpolator.h
```diff
@@ -3,7 +3,139 @@
...
+            // add symbol information
+            auto insertRes = symbolColors.insert(std::make_pair(logic.getSymRef(term), colorToAssign));
+            if (not insertRes.second) { // there was entry for this symbol already
+                auto entryIt = insertRes.first;
+                entryIt->second = colorUnion(entryIt->second, colorToAssign);
+            }
+        }
+        // Make sure complex terms have correct color assigned
+        vec<PTRef> terms;
+        for (auto const & entry : termColors) {
```
aehyvari, line 114: Consider having instead `for (auto const entry : termColors) {` ?

src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -348,10 +348,15 @@ bool UFInterpolator::colorEdgesFrom(CNode * x) {
...
-UFInterpolator::getInterpolant(const ipartitions_t &, std::map<PTRef, icolor_t> * labels, PartitionManager & ) {
+UFInterpolator::getInterpolant(const ipartitions_t &, std::map<PTRef, icolor_t> * labels, PartitionManager &) {
     assert(labels);
+    if (labels) {
+        colorInfo.reset(new LocalTermColorInfo(*labels, logic));
+        litColors = *labels;
+    } else {
+        throw OsmtInternalException("Error in UFInterpolator::getInterpolant! No labels passed");
+    }
     srand(2);
```
aehyvari, line 359: This seeding of the random number generator here seems strange.  Maybe something to consider for future.

aehyvari (approved):
I think this is fine.

src/tsolvers/egraph/UFInterpolator.cc
blishko: I agree, but it has been there before. Should I remove it?

src/smtsolvers/TheoryInterpolator.h
blishko: Replaced

src/smtsolvers/TheoryInterpolator.h
blishko: Taken by value.

PR #195, opened 2020-11-06
LIA interpolation

blishko:
This PR introduces the ability to compute interpolants also for QF_LIA queries.
The ability is somewhat limited to the current implementation of the LIA solver and will need to be revisited if new techniques are added to the LIA solver.

LIA interpolator builds on top of Farkas interpolator (for LRA), but it needs to preprocess the set of conflicting literals in order to match the strengthened constraints in the Simplex solver that led to the conflict.

Some small changes are needed in Farkas interpolator so that LIA interpolator can use it as a subcomponent.


aehyvari:
This looks good.  I'm wondering if there's a way of getting the LIA explanation construction out of the constructor, maybe by having a LIAExplanation class, which would then allow us to have FarkasInterpolator inside the class as a member and not a unique pointer.

blishko:
> I'm wondering if there's a way of getting the LIA explanation construction out of the constructor, maybe by having a LIAExplanation class, which would then allow us to have FarkasInterpolator inside the class as a member and not a unique pointer.

That's sounds like a good idea. I will try that.

blishko:
@aehyvari How about the helper class I have pushed in the latest commit? Is that something that you had in mind?

src/tsolvers/lrasolver/LRASolver.cc
```diff
@@ -101,9 +101,11 @@ lbool LRASolver::getPolaritySuggestion(PTRef ptref) const {
...
 PTRef
-LRASolver::getInterpolant( const ipartitions_t & mask , map<PTRef, icolor_t> *labels, PartitionManager &pmanager) {
+LRASolver::getInterpolant( const ipartitions_t & mask , std::map<PTRef, icolor_t> *labels, PartitionManager &pmanager) {
     assert(status == UNSAT);
-    FarkasInterpolator interpolator(logic, explanation, explanationCoefficients, labels,
+    vec<PtAsgn> explCopy;
+    explanation.copyTo(explCopy);
+    FarkasInterpolator interpolator(logic, std::move(explCopy), explanationCoefficients, labels ? *labels : std::map<PTRef, icolor_t>{},
                                     std::unique_ptr<TermColorInfo>(new GlobalTermColorInfo(pmanager, mask)));
     auto itpAlgorithm = config.getLRAInterpolationAlgorithm();
```
aehyvari, line 110: I'm curious about this mask situation here.  Should it be implicitly reusable for the user, or should the ownership be transferred to by default to the call through the API?  This might be an API-breaking change and we'd need to update the version number, but at the moment I think it would make the design cleaner.  What do you say?

aehyvari (commented):
After reading this I couldn't figure out how the interpolation works with splits.  Can you point me to the place where this happens?

blishko:
> After reading this I couldn't figure out how the interpolation works with splits. Can you point me to the place where this happens?

Good question, I am actually not sure right now :)
I will need to go over this again.

src/tsolvers/lrasolver/FarkasInterpolator.h
```diff
@@ -42,16 +41,28 @@ struct DecomposedStatistics {
 
 class FarkasInterpolator {
 public:
-    FarkasInterpolator(LALogic & logic, vec<PtAsgn> const & explanations, std::vector<opensmt::Real> const & coeffs,
-                       std::map<PTRef, icolor_t> & labels) : FarkasInterpolator(logic, explanations, coeffs, &labels,
-                                                                                nullptr) {}
+    FarkasInterpolator(LALogic & logic, vec<PtAsgn> explanations, std::vector<opensmt::Real> coeffs,
```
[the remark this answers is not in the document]
blishko: They could, but I prefer it this way.
There is no universal best way, there is quite some discussion about this in C++ world:
https://www.reddit.com/r/cpp/comments/gdf6ke/taking_sink_parameters_by_rvalue_reference/
https://stackoverflow.com/questions/7592630/is-pass-by-value-a-reasonable-default-in-c11

I guess I want to keep the option of creating this class with non-temporary objects as well, without the need to make explicit copies on caller's side.

src/tsolvers/liasolver/LIASolver.h
```diff
@@ -59,10 +59,11 @@ class LIASolver: public LASolver
     TRes check    ( bool complete) override; // Checks the satisfiability of current constraints //PS. add the implementation to LIASolver.C
     void getNewSplits(vec<PTRef>& splits) override;
 
+    PTRef getInterpolant( const ipartitions_t &, map<PTRef, icolor_t>*, PartitionManager & );
```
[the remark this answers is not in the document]
blishko: You are right. This seems to be related to the different ways how partitioning can be computed.
The signature was copied from LRASolver::getInterpolant, where it allows you to choose the way.
However, I think for LIA, only one way works, I think.
Since we don't have an agreement on a universal interpolation interface, I think I can remove the redundant parameters here.

src/tsolvers/lrasolver/LRASolver.cc
blishko: I am not sure I get your question. What should or should not be reusable?

src/tsolvers/lrasolver/FarkasInterpolator.h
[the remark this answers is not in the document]
aehyvari: Sounds reasonable.

src/tsolvers/lrasolver/LRASolver.cc
aehyvari: The question is related to the (resolved) discussion above.  We agreed to relieve the user of the burden of doing copies and, where necessary, to do them transparently inside opensmt.

src/tsolvers/liasolver/LIASolver.h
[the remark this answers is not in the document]
aehyvari: OK good, I agree.

blishko:
@aehyvari It's a good thing you asked about the splits. It turns out there was no test for that and once I have created one, the interpolation crashed :)

So, the splits are not assigned any partition currently, I will try to fix it.

aehyvari:
That's really good news in a way then, I suppose :)

blishko:
I have rebased and added a code to handle splits in the proof and interpolation that works for my one simple test :)
I will add more tests, I need to cover the cases when the spit is on A-local, B-local and AB-shared term.

blishko:
@aehyvari , I have added the tests for interpolation where splitting occurs. I think now it is ready.
Travis is green: https://travis-ci.com/github/blishko/opensmt/builds/212382852

src/proof/PGBuild.cc
```diff
@@ -187,126 +188,118 @@ void ProofGraph::buildProofGraph( int nVars )
...
-                    {
+                    else if (_ctype == clause_type::CLA_LEARNT) {
                         unsigned ssize = proof.getClause(clause).size();
-                        if( ssize >= max_learnt_size ) max_learnt_size = ssize;
+                        if (ssize >= max_learnt_size) { max_learnt_size = ssize; }
                         avg_learnt_size += ssize;
                     }
-                    else if(_ctype == clause_type::CLA_THEORY || _ctype == clause_type::CLA_ASSUMPTION)
-                    {
+                    else if(_ctype == clause_type::CLA_THEORY || _ctype == clause_type::CLA_ASSUMPTION || _ctype == clause_type::CLA_SPLIT) {
```
aehyvari, line 254: Add a space between if and '('?

src/proof/PGBuild.cc
```diff
@@ -187,126 +188,118 @@ void ProofGraph::buildProofGraph( int nVars )
...
-                    clauseToIDMap[clause]=currId;
+                    clauseToIDMap[clause] = currId;
                     //Add clause to queue
                     q.push_back(clause);
                 };
 
                 // First clause not yet processed
                 CRef clause_0 = chaincla[0];
-                while(clause_to_proof_der.at(clause_0).chain_cla.size() == 1)
+                while(clause_to_proof_der.at(clause_0).chain_cla.size() == 1) {
```
aehyvari, line 277: Space between while & (

src/proof/PGInterpolator.cc
```diff
@@ -538,14 +538,23 @@ void ProofGraph::produceSingleInterpolant ( vec<PTRef> &interpolants, const ipar
                 std::map<PTRef, icolor_t> ptref2label;
                 std::vector<Lit>& cl = n->getClause();
 
-                for(std::size_t j = 0; j < cl.size(); ++j)
-                {
+                for(std::size_t j = 0; j < cl.size(); ++j) {
```
aehyvari, line 541: Space between for & (

src/proof/PGInterpolator.cc
```diff
@@ -538,14 +538,23 @@ void ProofGraph::produceSingleInterpolant ( vec<PTRef> &interpolants, const ipar
...
+            else if (n->getType() == clause_type::CLA_SPLIT) {
+                auto const & clause = n->getClause();
+                assert(clause.size() == 2); // only binary splits at the moment
+                auto color = getVarColor(n, var(clause[0]));
+                assert(color == getVarColor(n, var(clause[1]))); // same theory variables in the atoms of the split => same color
+                assert(color == I_A || color == I_B || color == I_AB);
+                // If split on A-local (B-local) term, then return False (True). This is the same as in purely propoositional case.
+                // If split on AB-shared term, we can choose if we treat it as A-clause (resulting in False) or B-clause (resulting in True). We arbitrarily choose A now.
+                partial_interp = color == I_A ? logic_.getTerm_false() : (color == I_B ? logic_.getTerm_true() : logic_.getTerm_false());
+            }
```
aehyvari, line 557: This was the missing part, right?

src/tsolvers/lrasolver/FarkasInterpolator.h
```diff
@@ -42,16 +41,28 @@ struct DecomposedStatistics {
...
+                       std::map<PTRef, icolor_t> labels)
```
aehyvari, line 45: Is there a copy happening here?

src/tsolvers/lrasolver/FarkasInterpolator.h
```diff
@@ -79,8 +90,8 @@ class FarkasInterpolator {
 
     icolor_t getColorFor(PTRef term) const {
         // use labels
-        if(labels != nullptr && labels->find(term) != labels->end()){
-            return labels->at(term);
+        if(labels.find(term) != labels.end()){
```
aehyvari, line 93: Space between if & (

aehyvari (commented):
It looks good now.  I left some minor formatting comments.

src/proof/PGInterpolator.cc
blishko: Yes, but not only this. I added the clause type `clause_type::CLA_SPLIT`.

src/tsolvers/lrasolver/FarkasInterpolator.h
blishko: Yes, but that is intentional. We had this discussion about sink arguments taken by value vs by `&&`.
Or do you mean something else?

src/tsolvers/lrasolver/FarkasInterpolator.h
blishko: Fixed.

src/proof/PGInterpolator.cc
blishko: Fixed.

src/proof/PGBuild.cc
blishko: Fixed.

src/proof/PGBuild.cc
blishko: Fixed.

src/tsolvers/lrasolver/FarkasInterpolator.h
aehyvari: Right, this is the place.

aehyvari (approved):

Issue #197, opened 2020-11-09
FastRational.h:661:59: runtime error: signed integer overflow


rainoftime:
Hi, for the following formula,
opensmt 948bc88,
~~~~
$ opensmt xx.smt2
sat
sat
/opensmt/src/common/FastRational.h:661:59: runtime error: signed integer overflow: -5232614782500000000 - 5232614782500000000 cannot be represented in type 'long int'
sat

$cat xx.smt2
(set-logic QF_LRA)
(declare-fun r1 () Real)
(declare-fun r3 () Real)
(check-sat)
(check-sat)
(assert (< 0.8372183652 r3))
(assert (< (+ (- 0.8372183652 r1) r3) 0.0))
(check-sat)
~~~~

blishko:
Fixed by #204

PR #198, opened 2020-11-09
Simplified/unified the implementation of substraction

aehyvari:
The substraction in osmt reals implementation overflows in certain situations, as indicated in issue #197.  This PR addresses the issue by simplifying the handling of subtraction and implementing it similar to the substraction-assignment.

### DIFF
--- a/src/common/FastRational.h
+++ b/src/common/FastRational.h
@@ -674,14 +674,27 @@ inline void division(FastRational& dst, const FastRational& a, const FastRationa
     if (a.wordPartValid() && b.wordPartValid()) {
         uword common1 = gcd(absVal(a.num), absVal(b.num));
         uword common2 = gcd(a.den, b.den);
+        assert( common1 != 0 );
+        assert( common2 != 0 );
         word zn;
         uword zd;
         CHECK_WORD(zn, (lword(a.num)/common1) * (b.den/common2));
         CHECK_UWORD(zd, ulword(absVal(b.num)/common1) * (a.den/common2));
+
+        // Note: dst and a or b might be the same FastRational.
+        bool b_num_lt_0 = b.num < 0;
+        bool a_num_ge_0 = a.num >= 0;
+        bool b_num_gt_0 = b.num > 0;
+        bool a_num_le_0 = a.num <= 0;
+
         dst.num = zn;
         dst.den = zd;
-        if ((b.num < 0 && a.num >= 0) || (b.num > 0 && a.num <= 0)) dst.num = -absVal(dst.num);
-        if ((b.num > 0 && a.num >= 0) || (b.num < 0 && a.num <= 0)) dst.num = absVal(dst.num);
+
+        if ((b_num_lt_0 && a_num_ge_0) || (b_num_gt_0 && a_num_le_0)) dst.num = -absVal(dst.num);
+//        if ((b.num < 0 && a.num >= 0) || (b.num > 0 && a.num <= 0)) dst.num = -absVal(dst.num);
+        else if ((b_num_gt_0 && a_num_ge_0) || (b_num_lt_0 && a_num_le_0)) dst.num = absVal(dst.num);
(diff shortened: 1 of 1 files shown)


blishko (commented):
I like the changes, the code is simpler and the same as in `subtractionAssign`, so it's easier to reason about.

I have the following suggestions/comments:

1. I would add a unit test catching this bug in the `test_Rationals` (and maybe a couple more).
2. Now that `subtraction` and `subtractionAssign` are the same, can we have just one implementation and re-use it for both methods?
3. Should we run the SMT-LIB benchmarks just to check the effect on runtime? I have been wondering whether these special checks for the easy cases are useful and how much.

aehyvari:
> I would add a unit test catching this bug in the test_Rationals (and maybe a couple more).

I tried to do this but ran out of time with the project.  To get the thing actually to count wrong issue #197 doesn't work, as mentioned above.  I'd need some time to hand-craft an instance where this could happen.  I'll try to allocate a few hours to come up with one, but I'm not optimistic that I'll find the time easily :)

> Now that subtraction and subtractionAssign are the same, can we have just one implementation and re-use it for both methods?

I think this should be doable.  I'll commit a suggestion here soon.

> Should we run the SMT-LIB benchmarks just to check the effect on runtime? I have been wondering whether these special checks for the easy cases are useful and how much.

Yes, I'll run them now.

blishko:
> I tried to do this but ran out of time with the project. To get the thing actually to count wrong issue #197 doesn't work, as mentioned above.

When I turned on the sanitizers, I got the same error as reported in the issue. So there is an underflow happening, no? Shouldn't we be able to capture that in the test? Maybe I am missing something, but I can try to write the test that would fail in the original implementation.

aehyvari:
> When I turned on the sanitizers, I got the same error as reported in the issue. So there is an underflow happening, no? Shouldn't we be able to capture that in the test? Maybe I am missing something, but I can try to write the test that would fail in the original implementation.

The overflow happens yes, but later on in the code there's another case that catches another integer overflow.  So in the end the code expands the integers to the `gmp` representation and returns the correct answer.  To avoid the second overflow, it looked like the problem is somewhere in the area of finding two integers x, y both in (LWORD_MIN, LWORD_MIN/2) and gcd(x,y) being a large number.  How large, one would need to check more closely.  Maybe something like WORD_MIN < x / gcd(x,y) < 0?

Another option would be to just use the numbers from #197, but then the error is caught only run-time by a compiler-inserted check when these overflow checks are enabled in the compilation and not explicitly with gtest asserts.

blishko:
> The overflow happens yes, but later on in the code there's another case that catches another integer overflow. So in the end the code expands the integers to the gmp representation and returns the correct answer. To avoid the second overflow, it looked like the problem is somewhere in the area of finding two integers x, y both in (LWORD_MIN, LWORD_MIN/2) and gcd(x,y) being a large number. How large, one would need to check more closely. Maybe something like WORD_MIN < x / gcd(x,y) < 0?

Ah, OK, got it. I can try to find such a case.

aehyvari:
Here's the comparison on QF_LRA between 25b0e45f and master (at the end of September).  I think there's no consistent change. 

non-incremental-QF_LRA-master-2020-09-30_vs_substraction-overflow-2020-11-10

blishko:
Looks good!

aehyvari:
As discussed off-line, I'm trying to report here something about the effect of the arithmetic checks.  Here are the results for 68d8ca49, a commit where I removed the checks for simple cases in addition, subtraction, multiplication and division.  As a quick conclusion, there's a clearly noticeable, statistically significant slowdown when they're all disabled.

aehyvari:
Here's the figure with b01d7be, with checks removed only for addition and subtraction.  I think the slowdown starts to show off already there.

I think we might want to consider putting them back...

non-incremental-QF_LRA-master-2020-09-30_vs_b01d7be-2020-11-11

blishko:
> I think we might want to consider putting them back...

It might be necessary. I think the code was still understandable and the performance difference is not negligible.

PR #201, opened 2020-11-17
Added a gitter badge to readme

aehyvari:


blishko (approved):
We should collect as many badges as possible!

aehyvari:
> We should collect as many badges as possible!

Does it count if we make our own?

PR #204, opened 2020-12-21
Subtraction overflow

aehyvari:
This PR fixes undefined behaviour in OpenSMT rational number implementation.  In addition aims at improving the solver's overall performance in QF_LRA and QF_LIA.  The PR includes a more extensive set of unit tests aiming at a higher code coverage for the rationals implementation.

The PR was triggered by issue #197, a signed integer overflow in subtracting word-represented rationals.  The reported smtlib2 file is not included in the regression tests because the case is more reliably covered with the new unit tests, and the file did not show unsoundness or a crash under "normal" compiler options.  I do not have a concrete example where the previous version in opensmt master is unsound because of this now fixed bug.


aehyvari:
This PR should fix issue #197 and some related things that popped up while working on the fix.  I'm thinking of preserving the full chain of commits for historical reporting so that at least a part of the things that I tried on the way would be available for future reference (I have a bunch of different branches that didn't make it to this final version).   I'll add scatter plots for QF_LIA and QF_LRA comparing to the current master as soon as I have them.

aehyvari:
A few notes on the way:

 - The special case handling seems to be useful (e.g., when multiplying by 1, etc...)
 - The `*Assign` functions (`a += b` etc ...) as implemented now seem to be useful, I got a performance downgrade when I implemented as calls to the standard three-argument versions (`a  = a+b` etc...) and could not get rid of it.
 - The overflow check was too strict in some cases.  The (hopefully) exact implementation performs better.

blishko:
I am fine with preserving all the commits.
I will try to check this PR some time later today.

aehyvari:
non-incremental-QF_LIA-66f5ef34-2020-12-15_vs_7e3b6b57-2020-12-18
non-incremental-QF_LRA-66f5ef34-2020-12-01_vs_7e3b6b57-2020-12-17

aehyvari:
I'll still try to do two things on the PR:
1. Put the FastRational unit tests to the correct source file (whatever that is)
2. Make a separate performance test directory using the gtest infrastructure instead of having those as part of unit tests

src/common/FastRational.h
```diff
@@ -626,6 +676,8 @@ inline void addition(FastRational& dst, const FastRational& a, const FastRationa
             dst.den = zd;
         }
         dst.kill_mpq();
+        dst.setWordPartValid();
+////        assert(dst.isWellFormed());
```
blishko, line 680: I agree that `dst.setWordPartValid` should be called (it seems there was an implicit assumption that this would be always true).
Why is the `assert` commented out? I would keep it.

src/common/FastRational.h
```diff
@@ -572,52 +599,75 @@ inline bool FastRational::isWellFormed() const
...
+        }
     }
 }
 
 inline void addition(FastRational& dst, const FastRational& a, const FastRational& b) {
     if (a.wordPartValid() && b.wordPartValid()) {
         if (b.num == 0) {
-            dst.num = a.num;
-            dst.den = a.den;
+            dst = a;
```
blishko, line 634: I wanted to write I would keep the original, but then you get into questions like, what happens if `dst` had the `mpq` part allocated...

src/common/FastRational.h
```diff
@@ -572,52 +599,75 @@ inline bool FastRational::isWellFormed() const
...
         } else if (a.num == 0) {
-            dst.num = b.num;
-            dst.den = b.den;
+            dst = b;
+        } else if (a == -b) {
+            dst.num = 0;
+            dst.den = 1;
```
blishko, line 639: Shouldn't this behave as the other special cases? You get the same problem: what if `dst` has `mpq` part allocated?
Also I have to check, but I am suspicious about this check. What if `b.num` is WORD_MIN? This gets quite expensive. Maybe we could use `a.den == b.den and and b.num > WORD_MIN and a.num == -b.num`?
Now that I think about it, can we have a special case for `a.den == b.den`? The handling of that could be simpler, no?

src/common/FastRational.h
blishko: Now I am confused, `kill_mpq` should set the flag `wordPartValid` if this was not the case, no?

src/tsolvers/lasolver/Simplex.cc
```diff
@@ -353,8 +353,17 @@ bool Simplex::valueConsistent(LVRef v) const
...
+#endif
     for (auto & term : tableau.getRowPoly(v)){
-      sum += term.coeff * model->read(term.var);
+#ifdef TRACE_VALUE_CONSISTENT
+        cout << sum << " += " << term.coeff << " * " << model->read(term.var) << endl;
+#endif
+        sum += term.coeff * model->read(term.var);
+#ifdef TRACE_VALUE_CONSISTENT
+        cout << " => " << sum << endl;
+#endif
```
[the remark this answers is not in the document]
aehyvari: I'll remove the TRACE_VALUE_CONSISTENT.

src/common/FastRational.h
```diff
@@ -76,8 +76,15 @@ inline ulword absVal(lword x) {
...
 {
+    State state;
+    word num;
+    uword den;
+    mpq_t mpq;
+
     static const MpzUnit unit;
 
+
+
```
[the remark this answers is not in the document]
aehyvari: Removed 'em

src/common/FastRational.h
aehyvari: I believe that you're right.  I'll try to do that.

src/common/FastRational.h
aehyvari: > Shouldn't this behave as the other special cases? You get the same problem: what if dst has mpq part allocated?

You're right.  I'll fix the others.

> Also I have to check, but I am suspicious about this check. What if b.num is WORD_MIN? This gets quite expensive. Maybe we could use a.den == b.den and and b.num > WORD_MIN and a.num == -b.num?

I micro benchmarked this (the benchmarking commit will be a separate PR as discussed), but it looks like this would be a good change.  The difference on my mac is (on small numbers):

 - Real(nom, den) + Real(-nom, den) 26.2 ns vs 16.5 ns
 - Real(nom, den) + Real(nom, den) 73.9 ns vs 64.6 ns

I get very similar results on mantella.  I'll do this for subtraction as well.

> Now that I think about it, can we have a special case for a.den == b.den? The handling of that could be simpler, no?

Could this have a snowball effect?  I suggest that we keep this in mind and get back to it on a future performance iteration.

src/tsolvers/liasolver/Matrix.cc
```diff
@@ -525,6 +532,7 @@ LAMatrixStore::compute_hnf_v1(const MId U1, MId &H, int &dim1, MId &R1, MId &Ri1
     if (Ri1 != MId_Undef)
         setIdMatrix(Ri1);
 
+
```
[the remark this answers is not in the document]
aehyvari: Removed.

src/common/FastRational.h
aehyvari: Changed back to original

test/unit/test_ArithmeticExpressions.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <gtest/gtest.h>
 #include "LRALogic.h"
+#include "FastRational.h"
```
blishko, line 7: Adding this include is not necessary after the new tests has been moved to a different file.

src/common/FastRational.h
blishko: Agreed.

src/common/FastRational.h
```diff
@@ -640,32 +691,45 @@ inline void addition(FastRational& dst, const FastRational& a, const FastRationa
...
-        }
-        else if (a.den == 1) {
-            CHECK_WORD(dst.num, -lword(b.num) + lword(a.num)*b.den);
+        } else if (a.den == 1) {
+            // (a.num*b.den - b.den)/b.den is already canonicalized.
+            CHECK_WORD(dst.num, lword(a.num)*b.den - lword(b.num));
             dst.den = b.den;
-        }
-        else {
-            lword n = lword(a.num)*b.den - lword(b.num)*a.den;
```
blishko, line 661: Was there a bug here before? Is it possible that this expression overflows `lword`?
Basically, I am asking if the `CHECK_SUB_OVERFLOWS_LWORD`s are really necessary.

src/common/FastRational.h
```diff
@@ -739,19 +804,50 @@ inline void multiplication(FastRational& dst, const FastRational& a, const FastR
...
+            dst.setWordPartValid();
+            dst.kill_mpq();
+            return;
+        }
         uword common1 = gcd(absVal(a.num), absVal(b.num));
         uword common2 = gcd(a.den, b.den);
         word zn;
         uword zd;
-        CHECK_WORD(zn, (lword(a.num)/common1) * (b.den/common2));
+        CHECK_WORD(zn, ulword(absVal(a.num)/common1) * (b.den/common2));
```
blishko, line 827: Why is this using `absVal`?

src/common/FastRational.h
```diff
@@ -739,19 +804,50 @@ inline void multiplication(FastRational& dst, const FastRational& a, const FastR
...
         CHECK_UWORD(zd, ulword(absVal(b.num)/common1) * (a.den/common2));
+
+        // Note: dst and a or b might be the same FastRational.
+        bool b_num_lt_0 = b.num < 0;
+        bool a_num_ge_0 = a.num >= 0;
+        bool b_num_gt_0 = b.num > 0;
+        bool a_num_le_0 = a.num <= 0;
+
+        if ((b_num_lt_0 && a_num_ge_0) || (b_num_gt_0 && a_num_le_0)) { //CHECK_WORD(zn, (lword)(-absVal(zn)));
+            lword tmp = -(lword)(absVal(zn));
```
blishko, line 837: If we use `absVal` above, then there is no need to use it again here, no?

src/common/FastRational.h
```diff
@@ -739,19 +804,50 @@ inline void multiplication(FastRational& dst, const FastRational& a, const FastR
...
+            if (tmp < WORD_MIN || tmp > WORD_MAX) {
+                goto overflow;
+            }
+            zn = tmp;
+        }
+        else if ((b_num_gt_0 && a_num_ge_0) || (b_num_lt_0 && a_num_le_0)) CHECK_WORD(zn, (lword)(absVal(zn)));
```
blishko, line 843: Also here (`absVal`)

src/common/FastRational.h
```diff
@@ -840,12 +940,29 @@ inline void substractionAssign(FastRational& a, const FastRational& b) {
...
+        CHECK_WORD(zn, lword(common1 > 1 ? absVal(a.num)/common1 : absVal(a.num)) * (common2 > 1 ? absVal(b.num)/common2 : absVal(b.num)));
+        CHECK_UWORD(zd, ulword(common2 > 1 ? a.den/common2 : a.den) * (common1 > 1 ? b.den/common1 : b.den));
+
+        bool b_num_lt_0 = b.num < 0;
+        bool a_num_ge_0 = a.num >= 0;
+        bool b_num_gt_0 = b.num > 0;
+        bool a_num_le_0 = a.num <= 0;
+
+        if ((b_num_lt_0 && a_num_ge_0) || (b_num_gt_0 && a_num_le_0)) { //CHECK_WORD(zn, (lword)(-absVal(zn)));
+            lword tmp = -(lword)(absVal(zn));
```
blishko, line 957: Again here, the code above seems to guarantee now that `zn` will be positive, since `absVal` is used in its computation, so here it seems redundant.

src/common/FastRational.h
```diff
@@ -861,18 +978,27 @@ inline void multiplicationAssign(FastRational& a, const FastRational& b) {
...
+        CHECK_WORD(zn, ulword(absVal(a.num) / common1) * (b.den / common2));
+        CHECK_UWORD(zd, ulword(absVal(b.num) / common1) * (a.den / common2));
+
+        bool b_num_lt_0 = b.num < 0;
+        bool a_num_ge_0 = a.num >= 0;
+        bool b_num_gt_0 = b.num > 0;
+        bool a_num_le_0 = a.num <= 0;
+
+        if ((b_num_lt_0 && a_num_ge_0) || (b_num_gt_0 && a_num_le_0)) { //CHECK_WORD(zn, (lword)(-absVal(zn)));
+            lword tmp = -(lword) (absVal(zn));
```
blishko, line 994: `absVal` note again here.

blishko (commented):
Looks good!
There is one thing not clear to me, I left  comments at the appropriate places.

test/unit/test_ArithmeticExpressions.cpp
aehyvari: removed.

src/common/FastRational.h
aehyvari: I added a regression test for the case `common == 1` which overflows.  I should check if I can find one for the case where `common > 1`.

src/common/FastRational.h
aehyvari: I guess this is ultimately a question on whether the rest of the logic can be implemented more efficiently without the absVal.  I don't know.

src/common/FastRational.h
aehyvari: That seems to be true.  I'll remove absVal.

src/common/FastRational.h
aehyvari: I refactored this based on the comment.  Hopefully nothing broke.

src/common/FastRational.h
aehyvari: I agree.

src/common/FastRational.h
aehyvari: I agree.

src/common/FastRational.h
blishko: OK, I agree with your example (from the other discussion) that will overflow for `common == 1`, but I am pretty sure that it cannot overflow if `common > 1`:
At minimal `common == 2`, so `a.den / common < 2^31` and `(a.den / common) * b.num < 2^62`, similarly for `(b.den / common) * a/num < 2^62` and so their sum is `< 2^63` and it cannot overflow `lword`.
Is that correct?

aehyvari: I am not sure.  I'll need to repeat the same exercise I did for `common == 1` to convince ourselves.  While doing it for the `common == 1` there seemed to be a lot of headroom, so I'm a little surprised if `common > 1` never over/underflows, but I might be wrong.

aehyvari:
I'm running a quick performance test still on the changes you suggested above to see if there are surprises (also in the soundness area).  This will be excluding the suggestion to removal `CHECK_SUB_OVERFLOWS` in `common > 1` case so far.  I'll commit those changes if everything is ok to see if we agree up to there.

aehyvari:
QF_LRA for the fixed version:
non-incremental-QF_LRA-66f5ef34-2020-12-01_vs_d1fb5137-2020-12-23
QF_LIA for the fixed version:
non-incremental-QF_LIA-66f5ef34-2020-12-15_vs_d1fb5137-2020-12-23

src/common/FastRational.h
aehyvari: I'm quite convinced now that you're right, the `common > 1` case does not require the `CHECK_SUB_OVERFLOWS_LWORD`.  I'm going to run the benchmarks on that version (likely f7ce1e1f) and comment here probably tomorrow.

blishko:
Looks like the results are pretty similar to those before? I don't see the number of the benchmarks solved, though.
But overall, it looks fine, right? I will do one more pass, but it looks like we are ready to merge, no?

aehyvari:
Actually I figured it makes sense to commit also my final version.  I'm running the tests anyway and if there's something surprising, which I doubt, I'll just force-push it away or whatever makes sense.

src/common/FastRational.h
aehyvari: And for the record, if I remember correctly, this was the original bug reported in #197 which triggered this PR 😄

aehyvari:
> Looks like the results are pretty similar to those before? I don't see the number of the benchmarks solved, though.
> But overall, it looks fine, right? I will do one more pass, but it looks like we are ready to merge, no?

Yes, I see no surprises in the plots.

aehyvari:
The plots from f7ce171f.  I think they're fine.  

QF_LRA
non-incremental-QF_LRA-66f5ef34-2020-12-01_vs_f7ce1e1f-2020-12-23

QF_LIA
non-incremental-QF_LIA-66f5ef34-2020-12-15_vs_f7ce1e1f-2020-12-23

src/common/FastRational.h
```diff
@@ -539,21 +551,34 @@ inline int cmpabs(FastRational op1, FastRational op2)
...
+    } while (0)                                \
+
+#define CHECK_SUB_OVERFLOWS_LWORD(var, s1, s2) \
+    do {                                       \
+        if ((s1 > LWORD_MAX/2 || s2 < LWORD_MIN/2) || (s1 < LWORD_MIN/2 || s2 > LWORD_MAX/2)) { \
             goto overflow; \
         } \
-        var = s1 + s2;\
-    } while(0)
+        var = s1 - s2; \
```
aehyvari, line 579: There should be a better way of doing this check (see `CHECK_SUM_OVERFLOWS_LWORD`)?

src/common/FastRational.h
```diff
@@ -640,32 +691,50 @@ inline void addition(FastRational& dst, const FastRational& a, const FastRationa
...
     if (a.wordPartValid() && b.wordPartValid()) {
         if (b.num == 0) {
-            dst.num = a.num;
-            dst.den = a.den;
-        }
-        else if (a.num == 0) {
-            if (b.num == WORD_MIN)
-                goto overflow;
-            dst.num = -b.num; // This overflows if b = -2147483648
+            dst = a;
```
aehyvari, line 694: Use the old implementation `dst.num = a.num; dst.den = a.den;`

src/common/FastRational.h
```diff
@@ -573,11 +573,13 @@ template<uword> uword gcd(uword a, uword b);
...
-        } \
-        var = s1 - s2; \
-    } while (0)
+        if (s1 >= 0 and s2 < s1 - LWORD_MAX) { \
+            goto overflow;                     \
+        } if (s1 < 0 and s2 > (s1 + 1) + LWORD_MAX) { \
+            goto overflow;                     \
+        }                                      \
+        var = s1 - s2;                         \
+    } while (0)                                \
```
aehyvari, line 583: This seems to be working now.  I'm still running QF_LRA and QF_LIA on it just in case.

blishko: Seems correct to me as well.

src/common/FastRational.h
```diff
@@ -674,6 +745,7 @@ inline void substraction(FastRational& dst, const FastRational& a, const FastRat
             dst.den = zd;
         }
         dst.kill_mpq();
+        dst.setWordPartValid();
```
blishko, line 748: I think we agreed that this is not necessary, `kill_mpq` takes care of that. It is not present in the `addition`.

aehyvari:
Here are updated figures with the exact implementation of subtraction overflow check (7c3bbeea).

It's hard to say, but maybe the exact implementation is a little slower than f7ce171f.  In general it looks to me like a slight improvement over the current master, so maybe it's not worth the trouble to have an over-approximative implementation for a marginal, possibly random improvement.

QF_LRA
non-incremental-QF_LRA-66f5ef34-2020-12-01_vs_7c3bbeea-2020-12-24

QF_LIA
non-incremental-QF_LIA-66f5ef34-2020-12-15_vs_7c3bbeea-2020-12-24

src/common/FastRational.h
aehyvari: Right.  I spotted a few others as well and fixed them.

blishko:
I think it is hard to say, if the change is meaningful or random noise. I like the precise check, it permits to stay in the word representation for larger numbers.
I vote for the precise check.

aehyvari:
> I think it is hard to say, if the change is meaningful or random noise. I like the precise check, it permits to stay in the word representation for larger numbers.
> I vote for the precise check.

Yes, already two votes for that.  Let's go with the precise.  I committed something that should fix again the gcc errors and the extra calls to setWordPartValid().  If everything goes well then I think I'm happy with this current state.

src/common/FastRational.h
blishko: What about this one?

aehyvari: Well that's an open-ended question :)

How about:

```suggestion
    if (b.wordPartValid() && b.num == 0) {
            dst = a;
            return;
        } else if (a.wordPartValid() && b.wordPartValid() {
```

aehyvari: Let me commit that so it gets more clear.

blishko: In this suggestion, you would duplicate the check.
I like the original version.
```suggestion
            dst.num = a.num;
            dst.den = a.den;
```

aehyvari: Ok, the new commit is hopefully in line with what you suggest here.

blishko (approved):
I believe all the issues has been resolved.

PR #205, opened 2020-12-26
Fix in processing push and pop commands in the Interpret.

### LINKED ISSUE #202: False positive of push/pop stack error
Hi, for the following formula,
opensmt 948bc88,
~~~~
./opensmt xx.smt2
(error "Attempt to pop beyond the top of the stack")
sat
~~~~
However, the pop operation does not pop beyond the top of the stack, because the push is `(push 4)`
~~~~
(set-logic QF_UF)
(declare-fun v19 () Bool)
(push 4)
(pop 1)
(pop 1)
(check-sat)
~~~~

### LINKED ISSUE #203: Refutation soundness issue on QF_LRA formula
Hi, for the following formula,
opensmt 948bc88 yields `unsat`,
but z3 gives `sat`.

(seems related to the handling of push/pop)
~~~~
(set-logic QF_LRA)
(declare-fun v0 () Bool)
(declare-fun v1 () Bool)
(declare-fun v2 () Bool)
(declare-fun v3 () Bool)
(declare-fun v4 () Bool)
(declare-fun v5 () Bool)
(declare-fun v6 () Bool)
(declare-fun v7 () Bool)
(declare-fun v8 () Bool)
(declare-fun v9 () Bool)
(declare-fun r0 () Real)
(declare-fun r4 () Real)
(declare-fun r7 () Real)
(declare-fun r8 () Real)
(declare-fun r9 () Real)
(declare-fun r10 () Real)
(declare-fun r11 () Real)
(declare-fun r14 () Real)
(declare-fun r15 () Real)
(declare-fun v10 () Bool)
(assert v0)
(declare-fun v11 () Bool)
(declare-fun r17 () Real)
(declare-fun r18 () Real)
(declare-fun v13 () Bool)
(assert (< 0.175 r0))
(push 3)
(push 4)
(declare-fun r21 () Real)
(assert (not (< 0.175 r0)))
(declare-fun v21 () Bool)
(assert (=> (not (< 0.175 r0)) (< 0.175 r0)))
(assert (= (+ 3157.0 r15) (- r10)))
(push 4)
(declare-fun v22 () Bool)
(push 2)
(assert v11)
(assert (=> v21 (< r15 (+ (+ r18 0.994) r17))))
(push 1)
(pop 12)
(check-sat)
~~~~

blishko:
This PR fixes the problem in the Interpret where the numeric value how many frames should be pushed/popped was ignored previously.
This resulted in incorrect error messages or incorrect responses on valid SMT-LIB scripts, see #202, #203.

Fixes #202.
Fixes #203.


aehyvari:
This looks good.  The semantics is that

```
(push 1)
(pop 2)
```

means that one frame gets popped and for the second pop we print an error.  Alternative would be to check if the pops can be performed, and then either do all of them or fail (i.e., staying at depth 1 in the above example).  I don't mind the semantics of this pull request.  The 2.6 standard does not define what the solver should do in this case, so I guess we can go with whatever is easiest to implement.

aehyvari (approved):

PR #206, opened 2020-12-28
A more macos filesystem friendly name for our IntTypes wrapper

aehyvari:
This PR fixes a problem where on case-insensitive systems the opensmt's installed header called IntTypes.h gets confused with the system header inttypes.h causing builds to fail.  The suggested fix is to change the name of the header file.


aehyvari (commented):
In future we might want to study the feasibility of removing this file.  I'm not sure if the file would be needed for instance for a windows build?

blishko (approved):
I would probably prefer `OsmtIntTypes`, but this is also fine.

PR #207, opened 2020-12-28
Added a benchmarking system and micro benchmarks for Rationals

aehyvari:
A basis of writing micro benchmarks for opensmt.  This pull request also transfers the FastRational benchmark currently under unit tests to google's benchmark system.


CMakeLists.txt
```diff
@@ -153,6 +153,13 @@ if(PACKAGE_TESTS)
         add_subdirectory(${PROJECT_SOURCE_DIR}/test)
     endif(CMAKE_VERSION VERSION_LESS 3.11)
 endif()
+#########################################################################
+################# BENCHMARKING ##########################################
+option(PACKAGE_BENCHMARKS "Build the benchmarks" ON)
```
blishko, line 158: I am thinking that maybe this should be OFF by default?
This is more like an internal developer thing rather than something needed by the user.

blishko (approved):
Looks good!

CMakeLists.txt
aehyvari: Is now off by default.

PR #211, opened 2021-01-11
Tiny optimizations in Simplex

blishko:
This PR contains 2 small changes:

1. Adds helper methods to detect if a FastRational is positive or negative and uses these methods instead of more expensive explicit comparison with 0.
2. Pre-allocates exact space for the Simplex explanation, thus avoiding potential intermediate re-allocations.

### DIFF
--- a/src/tsolvers/lasolver/Simplex.cc
+++ b/src/tsolvers/lasolver/Simplex.cc
@@ -316,9 +316,12 @@ void Simplex::updateValues(const LVRef bv, const LVRef nv){
 Simplex::Explanation Simplex::getConflictingBounds(LVRef x, bool conflictOnLower)
 {
     LABoundRef br_f = conflictOnLower ? model->readLBoundRef(x) : model->readUBoundRef(x);
-    Explanation expl = {{br_f,1}};
+    auto const & row = tableau.getRowPoly(x);
+    Explanation expl;
+    expl.reserve(row.size() + 1);
+    expl.push_back({br_f,1});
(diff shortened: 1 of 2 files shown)


blishko:
I will measure the effect on SMT-LIB benchmark set once the cluster is free.

src/tsolvers/lasolver/Simplex.cc
```diff
@@ -316,13 +316,16 @@ void Simplex::updateValues(const LVRef bv, const LVRef nv){
 Simplex::Explanation Simplex::getConflictingBounds(LVRef x, bool conflictOnLower)
 {
     LABoundRef br_f = conflictOnLower ? model->readLBoundRef(x) : model->readUBoundRef(x);
-    Explanation expl = {{br_f,1}};
+    auto const & row = tableau.getRowPoly(x);
+    Explanation expl;
+    expl.reserve(row.size() + 1);
+    expl.push_back({br_f,1});
```
aehyvari, line 322: Why not emplace_back?

aehyvari (approved):
The changes look good to me.

src/tsolvers/lasolver/Simplex.cc
blishko: Because my compiler refused to compile that :)
I guess we would have to add a constructor to `ExplTerm` and I did not feel like doing that.
I am hoping compilers will do their magic and optimize this anyway...

blishko:
On LRA, the effect was pretty much 0%.
I have not measured LIA, only one example on my local where it did help :)

PR #212, opened 2021-01-14
Cleanup for undefined behaviour

aehyvari:
This PR removes undefined behaviour related to our use of `mtl::vec`.  It also fixes problems with indices in for-loops, indentations, and some soundness bugs in code that is currently not used.  The performance is not negatively affected, as can be seen from the scatter plots below (the y-axis commit hashes do not refer to hashes in the current PR because of rebasing, but they should be valid).

### Motivation
The goal of the PR is to allow portability for the new generation of language and compilers.

### Some observations
 - The main difference is the switch from the use of `mtl::vec<T>` to `std::vector<T>` in cases where the stored item is not trivially copyable.  `mtl::vec` uses `realloc` in case it runs out of space, and this is undefined behaviour for objects that are not trivially copyable (see the Notes section in the [c++ reference on realloc](https://en.cppreference.com/w/cpp/memory/c/realloc)).
- While going through the code I noticed some unnecessary copying in the storage implementation.  These are now fixed and might explain some speed-ups in easy instances.
- While I believe that some of the uses of `mtl::vec` are in fact safe because the code logic guarantees that realloc is never called, these cases are nevertheless reported as errors (or warnings) by some compilers (notably g++-9) so I decided to change the code.
- There are some uses of `std::pair` in the code.  `std::pair` is not trivially copyable, so I introduced `opensmt::pair` that is (given that the underlying template classes are), and is now used where necessary.
- In the process I changed MiniSat's `OccList`s to contain `std::vector<Vec>` instead of `mtl::vec<Vec>`.  This must have been anticipated by original developers because of the use of the template `Vec`.  I fixed the issue this way, and the plots below suggest that there is no performance overhead.
 
### plots

- QF_LIA
non-incremental-QF_LIA-b0153674-2021-01-11_vs_6ee66e86-2021-01-12
- QF_LRA
non-incremental-QF_LRA-b0153674-2021-01-11_vs_a134aa1a-2021-01-11
- QF_UF
non-incremental-QF_UF-b0153674-2021-01-11_vs_a134aa1a-2021-01-11

### DIFF
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -258,7 +258,7 @@ std::unique_ptr<InterpolationContext> MainSolver::getInterpolationContext() {
                                                                           getSMTSolver().getProof(), pmanager, getSMTSolver().nVars()));
 }
 
-void MainSolver::addToConj(std::vector<vec<PtAsgn> >& in, vec<PTRef>& out) const
+void MainSolver::addToConj(const std::vector<vec<PtAsgn> >& in, vec<PTRef>& out) const
 {
     for (const auto & constr : in) {
         vec<PTRef> disj_vec;
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -306,15 +306,16 @@ bool MainSolver::writeSolverState_smtlib2(const char* file, char** msg) const
 
 bool MainSolver::writeSolverSplits_smtlib2(const char* file, char** msg) const
 {
-    vec<SplitData>& splits = ts.solver.splits;
-    for (int i = 0; i < splits.size(); i++) {
+    std::vector<SplitData>& splits = ts.solver.splits;
+    int i = 0;
+    for (const auto & split : splits) {
         vec<PTRef> conj_vec;
         std::vector<vec<PtAsgn> > constraints;
-        splits[i].constraintsToPTRefs(constraints, thandler);
+        split.constraintsToPTRefs(constraints, thandler);
         addToConj(constraints, conj_vec);
 
         std::vector<vec<PtAsgn> > learnts;
-        splits[i].learntsToPTRefs(learnts, thandler);
+        split.learntsToPTRefs(learnts, thandler);
         addToConj(learnts, conj_vec);
 
         if (config.smt_split_format_length() == spformat_full)
@@ -323,7 +324,7 @@ bool MainSolver::writeSolverSplits_smtlib2(const char* file, char** msg) const
         PTRef problem = logic.mkAnd(conj_vec);
 
         char* name;
-        int written = asprintf(&name, "%s-%02d.smt2", file, i);
+        int written = asprintf(&name, "%s-%02d.smt2", file, i ++);
         assert(written >= 0);
         (void)written;
         std::ofstream file;
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -232,13 +232,13 @@ ValPair MainSolver::getValue(PTRef tr) const
     }
 }
 
-void MainSolver::getValues(const vec<PTRef>& trs, vec<ValPair>& vals) const
+void MainSolver::getValues(const vec<PTRef>& trs, std::vector<ValPair>& vals) const
 {
     vals.clear();
     for (int i = 0; i < trs.size(); i++)
     {
         PTRef tr = trs[i];
-        vals.push(getValue(tr));
+        vals.emplace_back(getValue(tr));
     }
 }
 
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -258,7 +258,7 @@ std::unique_ptr<InterpolationContext> MainSolver::getInterpolationContext() {
                                                                           getSMTSolver().getProof(), pmanager, getSMTSolver().nVars()));
 }
 
-void MainSolver::addToConj(vec<vec<PtAsgn> >& in, vec<PTRef>& out) const
+void MainSolver::addToConj(std::vector<vec<PtAsgn> >& in, vec<PTRef>& out) const
 {
     for (int i = 0; i < in.size(); i++) {
         vec<PtAsgn>& constr = in[i];
@@ -309,11 +309,11 @@ bool MainSolver::writeSolverSplits_smtlib2(const char* file, char** msg) const
     vec<SplitData>& splits = ts.solver.splits;
     for (int i = 0; i < splits.size(); i++) {
         vec<PTRef> conj_vec;
-        vec<vec<PtAsgn> > constraints;
+        std::vector<vec<PtAsgn> > constraints;
         splits[i].constraintsToPTRefs(constraints, thandler);
         addToConj(constraints, conj_vec);
 
-        vec<vec<PtAsgn> > learnts;
+        std::vector<vec<PtAsgn> > learnts;
         splits[i].learntsToPTRefs(learnts, thandler);
         addToConj(learnts, conj_vec);
 
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -260,11 +260,10 @@ std::unique_ptr<InterpolationContext> MainSolver::getInterpolationContext() {
 
 void MainSolver::addToConj(std::vector<vec<PtAsgn> >& in, vec<PTRef>& out) const
 {
-    for (int i = 0; i < in.size(); i++) {
-        vec<PtAsgn>& constr = in[i];
+    for (const auto & constr : in) {
         vec<PTRef> disj_vec;
-        for (int k = 0; k < constr.size(); k++)
-            disj_vec.push(constr[k].sgn == l_True ? constr[k].tr : logic.mkNot(constr[k].tr));
+        for (PtAsgn pta : constr)
+            disj_vec.push(pta.sgn == l_True ? pta.tr : logic.mkNot(pta.tr));
         out.push(logic.mkOr(disj_vec));
     }
 }
--- a/src/common/FastRational.cc
+++ b/src/common/FastRational.cc
@@ -177,7 +177,7 @@ FastRational divexact(FastRational const & n, FastRational const & d) {
 FastRational get_multiplicand(const std::vector<FastRational>& reals)
 {
     std::vector<FastRational> dens;
-    for (auto & r : reals) {
+    for (const auto & r : reals) {
         if (!r.isInteger()) {
             dens.push_back(r.get_den());
         }
@@ -188,7 +188,8 @@ FastRational get_multiplicand(const std::vector<FastRational>& reals)
     while (dens.size() > 0) {
         // Unique denominators
         std::sort(dens.begin(), dens.end());
-        std::unique(dens.begin(), dens.end());
+        auto last = std::unique(dens.begin(), dens.end());
+        dens.erase(last, dens.end());
 #ifdef PRINTALOT
         char *buf = (char*) malloc(1);
         buf[0] = '\0';
--- a/src/common/FastRational.cc
+++ b/src/common/FastRational.cc
@@ -6,7 +6,7 @@ Copyright (c) 2008, 2009 Centre national de la recherche scientifique (CNRS)
  */
 #include "FastRational.h"
 #include <sstream>
-#include <Sort.h>
+#include <algorithm>
 
 FastRational::FastRational( const char * s, const int base )
 {
@@ -174,19 +174,21 @@ FastRational divexact(FastRational const & n, FastRational const & d) {
 }
 
 // Given as input the sequence Reals, return the smallest number m such that for each r in Reals, r*m is an integer
-FastRational get_multiplicand(const vec<FastRational>& reals)
+FastRational get_multiplicand(const std::vector<FastRational>& reals)
 {
-    vec<FastRational> dens;
-    for (int i = 0; i < reals.size(); i++)
-        if (!reals[i].isInteger())
-            dens.push_c(reals[i].get_den());
+    std::vector<FastRational> dens;
+    for (auto & r : reals) {
+        if (!r.isInteger()) {
+            dens.push_back(r.get_den());
+        }
+    }
 
     // Iterate until dens is not empty
     FastRational mult = 1; // Collect here the number I need to multiply the polynomial to get rid of all denominators
     while (dens.size() > 0) {
         // Unique denominators
-        sort(dens);
-        uniq(dens);
+        std::sort(dens.begin(), dens.end());
+        std::unique(dens.begin(), dens.end());
 #ifdef PRINTALOT
         char *buf = (char*) malloc(1);
         buf[0] = '\0';
@@ -210,13 +212,12 @@ FastRational get_multiplicand(const vec<FastRational>& reals)
             int k = 0;
             FastRational m = lcm(dens[dens.size()-1], dens[dens.size()-2]);
             mult *= m;
-            int orig_sz = dens.size();
-            for (int j = 0; j < dens.size()-2; j++) {
+            for (size_t j = 0; j < dens.size()-2; j++) {
                 FastRational n = (m/dens[j]).get_den();
                 if (n != 1)
                     dens[k++] = n;
             }
-            dens.shrink(orig_sz-k);
+            dens.resize(k);
         }
     }
 #ifdef PRINTALOT
--- a/src/common/Global.h
+++ b/src/common/Global.h
@@ -279,6 +279,7 @@ namespace opensmt {
         return true;
     }
 
+    // std::pair is not trivially copyable.  This version seems to allow on current compilers faster implementations.
     template <class T, class U> struct pair { T first; U second; };
 
 #define Pair(T) pair< T, T >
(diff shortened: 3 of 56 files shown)


src/minisat/mtl/Vec.h
```diff
@@ -42,9 +41,9 @@ class vec {
     int sz;
     int cap;
 
-    // Don't allow copying (error prone):
-    vec<T>&  operator = (vec<T>& other) = delete;
-             vec        (vec<T>& other) = delete;
+    vec<T>&  operator = (const vec<T>& other) {
+                static_assert(std::is_copy_constructible<T>::value, "mtl::vec elements need to be copy constructable");
```
aehyvari, line 45: Change to std::is_trivially_copyable

src/minisat/mtl/Vec.h
```diff
@@ -184,51 +118,27 @@ vec<T>::vec(const std::initializer_list<T> &iList) : data(NULL), sz(0), cap(0) {
 
 template<class T>
 void vec<T>::capacity(int min_cap) {
+    static_assert(std::is_copy_constructible<T>::value, "mtl::vec elements need to be copy constructible");
```
aehyvari, line 121: Change to std::is_trivially_copyable

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -2146,12 +2146,10 @@ bool CoreSMTSolver::createSplit_scatter(bool last)
         // space
         constraints_negated.push(~l);
     }
-    for (int i = 0; i < split_assumptions.size()-1; i++)
-    {
+    for (const auto & split_assumption : split_assumptions) {
```
blishko, line 2149: The previous version was ignoring the last element of `split_assumptions`, the new version includes it. Which one is correct?

blishko, line 2149: The previous version was ignoring the last element of `split_assumptions`, the new version includes it. Which one is correct?

src/smtsolvers/CoreSMTSolver.h
```diff
@@ -182,12 +171,8 @@ inline char* SplitData::splitToString()
     }
 */
     // The constraints
-    for (int i = 0; i < constraints.size(); i++)
-    {
-        vec<Lit>& c = constraints[i];
-        for (int j = 0; j < c.size(); j++)
-        {
-            Lit l = c[j];
+    for (vec<Lit>& c : constraints) {
```
blishko, line 174: Can be `const`?

src/smtsolvers/GhostSMTSolver.cc
```diff
@@ -188,11 +188,9 @@ GhostSMTSolver::pickBranchLit() {
 }
 
 void GhostSMTSolver::relocAll() {
-    for (int i = 0; i < thLitToClauses.size(); i++) {
-        vec<CRef> &appearances = thLitToClauses[i];
-        for (int j = 0; j < appearances.size(); j++) {
-            CRef cr = appearances[j];
-            appearances[j] = ca[cr].relocation();
+    for (auto & appearances : thLitToClauses) {
+        for (CRef & cr : appearances) {
```
blishko, line 192: `auto const & appearances` and `CRef cr`?

src/smtsolvers/LAScore.h
```diff
@@ -16,7 +16,8 @@ template<class EV>
 class LABestLitBuf {
 private:
     int size;
-    vec<std::pair < EV, Lit> > buf;
+    struct EVLitPair { EV first; Lit second; };
```
blishko, line 19: Can we decide on one thing from `std::pair` vs `opensmt::pair` vs custom pair `struct`?

src/tsolvers/liasolver/Matrix.h
```diff
@@ -26,10 +26,10 @@ class LAVec
     opensmt::Real args[0]; // Either the elements or the relocation reference
 
 public:
-    LAVec(const vec<opensmt::Real>& ps, const opensmt::Real& den) : den(den) {
+    LAVec(std::vector<opensmt::Real>&& ps, const opensmt::Real& den) : den(den) {
         header.size = ps.size();
         header.reloced = 0;
-        for (int i = 0; i < ps.size(); i++) {
+        for (int i = 0; i < static_cast<int>(ps.size()); i++) {
```
blishko, line 32: `int i` -> `std::size_t i` and there is no need for the cast.

test/unit/test_Matrix.cpp
```diff
@@ -538,7 +539,7 @@ TEST(lcm_test, lcm_list)
     FastRational p2(3, 7);
     FastRational p3(5, 11);
 
-    vec<FastRational> l = {p1, p2, p3};
+    vector<FastRational> l = {p1, p2, p3};
```
blishko, line 542: `std::vector`

test/unit/test_Rationals.cpp
```diff
@@ -132,16 +132,6 @@ TEST(Rationals_test, test_overwrite)
...
-TEST(Rationals_test, test_unique)
-{
-    vec<Real> v = {1, 1, 2, 3, 3, 4};
-    vec<Real> w = {1, 2, 3, 4};
-    uniq(v);
-    ASSERT_EQ(v.size(), w.size());
-    for (int i = 0; i < v.size(); i++)
-        ASSERT_EQ(v[i], w[i]);
-}
-
```
blishko, line 144: Should we replace this test with something else?

blishko (commented):
Overall looks good, I am in favour of the changes.
I left a small comment and my main issue is on handling pairs.
In some context I am in favour of custom `struct`s, just to give proper names to the components.
But I fail to see the need for the `opensmt::pair`.

src/common/Global.h
```diff
@@ -279,6 +279,7 @@ namespace opensmt {
         return true;
     }
 
+    template <class T, class U> struct pair { T first; U second; };
```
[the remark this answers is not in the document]
blishko: Maybe add the explanation from the PR to the code?

src/smtsolvers/LAScore.h
blishko: Give more meaningful names to the components or just use `opensmt::pair`.

aehyvari:
> In some context I am in favour of custom structs, just to give proper names to the components.
> But I fail to see the need for the opensmt::pair.
> 
> Edit: Only now I have read your explanation. Interesting!

Yes, sorry, I guess it was a little hidden...  But also on a higher level your point remains valid, the use of pairs is inconsistent.  Here I didn't follow a principle other than a "greedy strategy for minimum changes", which is easily shown to be suboptimal :)

I see now two alternatives for consistency-first approach.  We could try to use consistently either:
1. `opensmt::pair` and see if that goes through, or
2. custom structs for pairs.

If we go with the first one, then the variable name should make it clear what are the `first` and `second` of the structs.  Based on quick googling around we don't seem to be the only ones who've had to ditch the exclusive use of `std::pair` for this reason (see, for instance, https://www.reddit.com/r/cpp/comments/ar4ghs/stdpair_disappointing_performance/)

src/api/MainSolver.cc
```diff
@@ -258,13 +258,12 @@ std::unique_ptr<InterpolationContext> MainSolver::getInterpolationContext() {
                                                                           getSMTSolver().getProof(), pmanager, getSMTSolver().nVars()));
 }
 
-void MainSolver::addToConj(vec<vec<PtAsgn> >& in, vec<PTRef>& out) const
+void MainSolver::addToConj(std::vector<vec<PtAsgn> >& in, vec<PTRef>& out) const
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/common/FastRational.cc
```diff
@@ -174,19 +174,21 @@ FastRational divexact(FastRational const & n, FastRational const & d) {
...
 // Given as input the sequence Reals, return the smallest number m such that for each r in Reals, r*m is an integer
-FastRational get_multiplicand(const vec<FastRational>& reals)
+FastRational get_multiplicand(const std::vector<FastRational>& reals)
 {
-    vec<FastRational> dens;
-    for (int i = 0; i < reals.size(); i++)
-        if (!reals[i].isInteger())
-            dens.push_c(reals[i].get_den());
+    std::vector<FastRational> dens;
+    for (auto & r : reals) {
```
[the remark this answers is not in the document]
aehyvari: I guess it has to be, but I'll make it explicit.

Did I get you right?
```for (const auto & r : reals) {```

File: src/common/FastRational.cc [hunk 13]
```diff
@@ -174,19 +174,21 @@ FastRational divexact(FastRational const & n, FastRational const & d) {
...
+    }
 
     // Iterate until dens is not empty
     FastRational mult = 1; // Collect here the number I need to multiply the polynomial to get rid of all denominators
     while (dens.size() > 0) {
         // Unique denominators
-        sort(dens);
-        uniq(dens);
+        std::sort(dens.begin(), dens.end());
+        std::unique(dens.begin(), dens.end());
```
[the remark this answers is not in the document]
aehyvari: Good, thanks!  That's a bug.

src/common/Global.h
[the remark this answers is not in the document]
aehyvari: Added.

src/minisat/mtl/Map.h
```diff
@@ -264,13 +264,13 @@ class Map {
 template<class K, class D, class H, class E = Equal<K> >
 class VecMap {
  public:
-    struct Pair { K key; vec<D> data; };
+    struct Pair { static_assert(std::is_trivially_copyable<D>::value, "elements of mtl::VecMap vectors need to be trivially copyable"); K key; vec<D> data; };
```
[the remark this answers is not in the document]
aehyvari: Sounds reasonable.  I moved it to the class, this seems to work.

src/minisat/mtl/Vec.h
```diff
@@ -42,9 +41,9 @@ class vec {
     int sz;
     int cap;
 
-    // Don't allow copying (error prone):
-    vec<T>&  operator = (vec<T>& other) = delete;
-             vec        (vec<T>& other) = delete;
+    vec<T>&  operator = (const vec<T>& other) {
+                static_assert(std::is_trivially_copyable<T>::value, "mtl::vec elements need to be trivially copyable");
```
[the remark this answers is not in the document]
aehyvari: It looks like some compilers (clang) can compile the template code if the copy assignment is never called from the code.  I thought this is a feature, so I left it to the place where it is actually required.  But I didn't do a lot of research on this.  At the moment the code does not require this (as far as I know), since g++-9 was complaining about the copy assignment even in cases where it was never called.

src/proof/PGBuild.cc
```diff
@@ -314,8 +314,8 @@ void ProofGraph::buildProofGraph( int nVars )
                             else
                             { c2 = last_seen_id; c1 = id__i; }
                             // Look for pair <ant1,ant2>
-                            pair<int, int> ant_pair (c1,c2);
-                            map< pair<int,int>, int >::iterator it = ants_map.find( ant_pair );
+                            std::pair<int, int> ant_pair (c1,c2);
+                            map< std::pair<int,int>, int >::iterator it = ants_map.find( ant_pair );
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/proof/PGHelp.cc
```diff
@@ -300,7 +300,7 @@ void ProofGraph::getComplexityInterpolant( PTRef int_e )
             if (logic_.isAtom(e_curr))
             {
                 // Complexity of atom is 0
-                complexity_map.insert( pair< PTRef, unsigned long >( e_curr, 0 ) );
+                complexity_map.insert( std::pair< PTRef, unsigned long >( e_curr, 0 ) );
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/pterms/PtStore.cc
```diff
@@ -144,19 +144,13 @@ void PtStore::addToCtermMap(SymRef& k, PTRef tr) {
...
 PTRef PtStore::getFromCtermMap(SymRef& k) { return cterm_map[k]; }
 
-bool PtStore::hasBoolKey(PTLKey& k) { return bool_map.has(k); }
-void PtStore::addToBoolMap(PTLKey& k, PTRef tr) {
-    bool_map.insert(k, tr);
-//        bool_keys.push(k);
-}
-PTRef PtStore::getFromBoolMap(PTLKey& k) { return bool_map[k]; }
+bool PtStore::hasBoolKey(const PTLKey& k) { return bool_map.find(k) != bool_map.end(); }
+void PtStore::addToBoolMap(PTLKey &&k, PTRef tr) { bool_map[std::move(k)] = tr; }
```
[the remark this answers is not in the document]
aehyvari: Done.

src/logics/Logic.h
```diff
@@ -93,17 +93,17 @@ class Logic {
...
     public:
-        bool has(const char* name) const { return defined_functions.has(name); }
+        bool has(const char* name) const { return defined_functions.find(name) != defined_functions.end(); }
 
         void insert(const char* name, TFun const & templ) {
             assert(not has(name));
             defined_functions_names.push();
-            defined_functions_names.last() = strdup(name);
-            defined_functions.insert(defined_functions_names.last(), templ);
+            defined_functions_names.last() = strdup(std::string(name).c_str());
```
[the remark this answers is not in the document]
aehyvari: Fixing to 

```defined_functions_names.last() = strdup(name);```

File: src/logics/Logic.h [hunk 20]
```diff
@@ -112,7 +112,9 @@ class Logic {
         }
 
         void getKeys(vec<const char*> & keys_out) {
-            defined_functions.getKeys(keys_out);
+            for (auto k : defined_functions_names) {
+                defined_functions_names.push(strdup(k));
```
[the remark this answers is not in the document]
aehyvari: Oops, right!

src/logics/SubstLoopBreaker.cc
```diff
@@ -236,45 +237,45 @@ Map<PTRef,PtAsgn,PTRefHash> SubstLoopBreaker::operator() (Map<PTRef,PtAsgn,PTRef
...
     return constructLooplessSubstitution(std::move(substs));
 }
 
-vec<SNRef> SubstLoopBreaker::breakLoops(const vec<vec<SNRef>>& loops) {
+vec<SNRef> SubstLoopBreaker::breakLoops(const std::vector<vec<SNRef>>& loops) {
     // Break the found loops.  Return the orphaned children as new start nodes
     vec<SNRef> orphans;
-    for (int i = 0; i < loops.size(); i++) {
-        int last_idx = loops[i].size()-1;
+    for (auto & loop : loops) {
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/smtsolvers/CoreSMTSolver.cc
aehyvari: Right, this is a bug as well. It should ignore the last element.

src/common/FastRational.cc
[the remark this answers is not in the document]
blishko: Yes

src/smtsolvers/CoreSMTSolver.cc
aehyvari: This comment seems to be a duplicate.

src/smtsolvers/CoreSMTSolver.h
aehyvari: Fixed.

src/smtsolvers/GhostSMTSolver.cc
aehyvari: Fixed.

src/smtsolvers/LAScore.h
aehyvari: Fixed to use opensmt::pair.

src/tsolvers/liasolver/Matrix.h
aehyvari: Fixed.

test/unit/test_Matrix.cpp
aehyvari: Fixed.

test/unit/test_Rationals.cpp
aehyvari: I added a unit test for FastRational::get_multiplicand, that's how it should have been in the first place I think.

blishko: Maybe also add a test that would exercise the use of unique in the implementation.
Maybe with values `1/3, 1/5, 2/5`?

blishko (commented):
I think this is good.
I would add one more test as suggested in the comment.
I will also run Travis in my clone on this, see if it finds any problems we missed.

test/unit/test_Rationals.cpp
aehyvari: I added another unit test here, as you suggested below.

blishko (approved):
Travis is green (not checking the last test): https://travis-ci.com/github/blishko/opensmt/builds/213392898

I think this is good to go!

PR #214, opened 2021-01-16
Add support for CircleCI

aehyvari:
Here's an idea how we could do CI in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 jobs:
-    build:
+    build-recent:
         docker:
           - image: cimg/base:stable
             auth:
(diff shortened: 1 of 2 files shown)


aehyvari:
It looks like we have to merge this pull request to master before I can enable the branch conditions for the jobs defined in circleci yaml.

blishko:
OK, let's merge it and see.

.circleci/config.yml
```diff
@@ -0,0 +1,118 @@
...
+            - run:
+                name: Clean build directory
+                command: |
+                    rm -rf build
+                    mkdir build
+            - run:
+                name: Install clang
+                command: |
+                    sudo apt-get install clang
+                    sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 60
```
blishko, line 45: This changes the C++ compiler to point to clang so that the next commands will compile with clang?

blishko (approved):

.circleci/config.yml
aehyvari: Correct.  debian uses this update-alternatives system to support several compilers.

PR #217, opened 2021-01-21
tiny cleanup in minisat Vec

blishko:
The copy assignment in `Vec` as currently written is incorrect. Fortunately, it is not used anywhere.
This PR removes it which prevents the copy assignment of `Vec` by the compiler.

If we ever decide to switch to the standard copy-constructor instead of using `copyTo`, I suggest to have it in the code, just commented out for now.


aehyvari (approved):
This looks good.

blishko:
Rebased.

PR #218, opened 2021-01-22
Fix manipulation with Pterms in Model

blishko:
This PR fixes the problem in Modell where a C++ reference to `Pterm` is kept and used in a loop where new terms can be created. This can potentially cause a segmentation fault if reallocation of the term table happens inside the loop.

### DIFF
--- a/src/models/Model.cc
+++ b/src/models/Model.cc
@@ -23,11 +23,11 @@ PTRef Model::evaluate(PTRef term) {
     }
     else {
         // complex term not seen before, compute and store the value
-        const Pterm & t = logic.getPterm(term);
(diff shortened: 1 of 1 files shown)


aehyvari (approved):
Right, ok, so we use recursion in evaluate.  Looks good.   I guess there's no risk of stack overflow here, but something to keep in mind.

blishko:
Well, there is some risk, potentially. I think it would be good to rewrite this to avoid the recursion.
This version was just the easiest and the most straightforward to implement.

aehyvari:
Let's keep it in mind then.

PR #220, opened 2021-01-28
Fixes the presentation of a single interpolant

aehyvari:
As pointed out in #213 the output of a single interpolant was not conforming to the standard.  This pull-request should fix the single-inteprolant printing as well.


blishko (approved):
LGTM!

Issue #146, opened 2020-07-22
Implement standard-compliant handling of inequality operators with more than 2 arguments


blishko:
Follow up on issue #137  and PR #138 
The proper handling of multiple arguments should be extended from equality also to inequality operators according to the SMT-LIB definitions of theories of Reals and Integers.

aehyvari:
This affects also xor.  In smtlib2 standard xor is marked as left-assoc, meaning that the term `(xor a b c)` should be interpreted as `(xor (xor a b) c)`.

blishko:
The same applies to `/` and `div`.

PR #221, opened 2021-01-29
Adding basic support for int mod and div operations

blishko:
I am opening this PR draft to discuss changes necessary to add support for `div` and `mod` in LIA logic.

So far, there is support for creating `div` and `mod` terms where the second argument is constant. 
A utility to rewrite formulas with these operations to equivalent formulas without them is still needed.


src/logics/LIALogic.h
```diff
@@ -32,6 +34,8 @@ class LIALogic: public LALogic
     static const char*  tk_int_plus;
     static const char*  tk_int_times;
     static const char*  tk_int_div;
+    static const char*  tk_int_mod;
+    static const char*  tk_int_abs;
```
blishko, line 38: I noticed that `Int` theory declares also `abs` function, so I put it here, but maybe I can remove it since it is not necessary for what I need right now.

blishko:
@aehyvari sorry for all the changes, but there was some inconsistencies in how `LALogic` delegates the queries to `LIALogic` and `LRALogic`, so I did some refactoring.

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+//
+
+#include "LARewriter.h"
+
+#include "LIALogic.h"
+#include "OsmtApiException.h"
+
+std::size_t DivModRewriter::counter = 0;
+const char * DivModRewriter::divPrefix = ".div";
+const char * DivModRewriter::modPrefix = ".mod";
```
blishko, line 12: I thought about if this should be in LIALogic or here in the rewriter. Now I am more inclined  to keep it here.
It feels closer to there it is actually used and I believe it should not cause any problems.

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+        toProcess.pop_back();
+    }
+    if (substitutions.empty()) {
+        assert(definitions.size() == 0);
+        return root;
+    }
+    auto it = substitutions.find(root);
+    assert(it != substitutions.end());
+    definitions.push(it->second);
+    return logic.mkAnd(definitions);
```
blishko, line 110: While writing this function, I realized it is very close to `Logic::simplifyTree`.
I think it would be good to unify the approach.
One idea would be to implement this as part of `LIA::visit`, but I have a different idea that I like better.
We would have a general rewriter (living outside of `Logic`) which would be rewriting the DAG based on registered plugins. The solver than can customize the rewriter for different theories, but it would also allow users to create their own rewriters.

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+        Pterm const & term = logic.getPterm(currentRef);
+        unsigned childrenCount = term.size();
+        if (currentEntry.nextChild < childrenCount) {
+            PTRef nextChild = term[currentEntry.nextChild];
+            ++currentEntry.nextChild;
+            if (done[Idx(logic.getPterm(nextChild).getId())] == 0) {
+                toProcess.push_back(DFSEntry(nextChild));
+            }
+            continue;
+        }
```
aehyvari, line 55: I've been writing the above code slightly different, see for instance [the corresponding loop in IteToSwitch.cc](https://github.com/usi-verification-and-security/opensmt/blob/e33b1470696e02e9360f4091ceadfdf8b709657d/src/itehandler/IteToSwitch.cc#L40).  Not sure which one is better, but I guess we should settle for one version.  I think grepping `unprocessed_children` would reveal most of these cases.

aehyvari: This is probably an argument for introducing the general visit pattern.  Or maybe that's yet another topic since we don't do the rewriting in all of the cases.

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+    std::vector<char> done;
+    done.resize(size, 0);
+    std::unordered_map<PTRef, PTRef, PTRefHash> substitutions;
+    std::vector<DFSEntry> toProcess;
+    toProcess.push_back(DFSEntry{root});
+    while (not toProcess.empty()) {
+        auto & currentEntry = toProcess.back();
+        PTRef currentRef = currentEntry.term;
+        Pterm const & term = logic.getPterm(currentRef);
+        unsigned childrenCount = term.size();
```
aehyvari, line 47: size_t would maybe work better below.

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+            if (done[Idx(logic.getPterm(nextChild).getId())] == 0) {
+                toProcess.push_back(DFSEntry(nextChild));
+            }
+            continue;
+        }
+        // If we are here, we have already processed all children
+        assert(done[Idx(term.getId())] == 0);
+        vec<PTRef> newArgs(childrenCount);
+        bool needsChange = false;
+        for (unsigned i = 0; i < childrenCount; ++i) {
```
aehyvari, line 60: `unsigned` => `auto` (at least if `childrenCount` is `size_t` this works)

src/simplifiers/LARewriter.h
```diff
@@ -0,0 +1,33 @@
...
+/// Class for removing div/mod expressions from a formula suitable for re-use
+class DivModRewriter {
+    static std::size_t counter;
+    static const char * divPrefix;
+    static const char * modPrefix;
+    struct DivModPair {
+        PTRef div;
+        PTRef mod;
+    };
+    std::unordered_map<std::pair<PTRef, PTRef>, DivModPair, PTRefPairHash> divModeCache;
```
aehyvari, line 27: Should this be called `divModCache` instead of `divModeCache`?

src/logics/LIALogic.h
```diff
@@ -47,59 +51,51 @@ class LIALogic: public LALogic
...
+    virtual const SymRef get_sym_Num_ZERO () const override { return sym_Int_ZERO; }
+    virtual const SymRef get_sym_Num_ONE () const override { return sym_Int_ONE; }
+    virtual const SymRef get_sym_Num_ITE () const override { return sym_Int_ITE; }
+    const SymRef get_sym_Int_MOD() const { return sym_Int_MOD; }
+    const SymRef get_sym_Int_ABS() const { return sym_Int_ABS; }
+    virtual const SRef get_sort_Num () const override { return sort_INTEGER; }
+
+    const PTRef mkIntMod(PTRef first, PTRef second);
+    const PTRef mkIntAbs(PTRef arg);
+    PTRef mkNumDiv(vec<PTRef> const & args) override;
```
aehyvari, line 98: Why this takes a `vec` and `mkIntMod` takes two `PTRef`s?

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+    std::string divName = divPrefix + num;
+    std::string modName = modPrefix + num;
+    return {logic.mkNumVar(divName.c_str()), logic.mkNumVar(modName.c_str())};
+}
+
+PTRef rewriteDivMod(LIALogic & logic, PTRef root) {
+    return DivModRewriter().rewriteDivMod(logic, root);
+}
+
+// TODO: figure out how to cache assigned variables to avoid duplication
```
aehyvari, line 25: Is this comment still valid?  It looks like there is a cache already in place, but maybe it refers to something else?

aehyvari (changes requested):
Looks excellent!  Is this code used somewhere outside of regression tests?

src/simplifiers/LARewriter.cc
```diff
@@ -0,0 +1,111 @@
...
+PTRef DivModRewriter::rewriteDivMod(LIALogic &logic, PTRef root) {
+    if (root == PTRef_Undef or not logic.hasSortBool(root)) {
+        throw OsmtApiException("Div/Mod rewriting should only be called on formulas, not terms!");
+    }
+    vec<PTRef> definitions;
+    struct DFSEntry {
+        DFSEntry(PTRef term) : term(term) {}
+        PTRef term;
+        unsigned int nextChild = 0;
```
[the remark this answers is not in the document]
blishko: I thought that range of unsigned int is sufficient and it requires less memory than `size_t`.

src/simplifiers/LARewriter.cc
blishko: I am thinking that we could have a const visitor that does not modify the terms, just collects information and than separately a rewriter that rewrites the terms.

blishko:
The code is not used anywhere outside the unit tests. I was thinking where to call it and came across the similarity with `Logic::visit`, and thought that I will try to unify them in a single mechanism.

Also, I think I have not yet added support in the parser for this.

src/simplifiers/LARewriter.h
blishko: Fixed.

src/logics/LIALogic.h
blishko: I think this is how it was before. When it was common for LIA and LRA.

Should we distinguish between the real operator `/` and the integer operator `div` at the level of `mk` methods?

src/simplifiers/LARewriter.cc
blishko: I think it is dealt with now.

blishko:
@aehyvari , I have rebased this on top of current master and rewritten it using the `Rewriter` template.
I am thinking of **not** using it in the preprocessing since it is not needed for the SMT-LIB benchmarks we actually support.
Rather, we can just keep it as a functionality available through API (that is enough for my use case).

I am also thinking of hiding `DivModConfig` inside `DivModRewriter`, but I can do it later, when addressing your anticipated comments.

src/rewriters/DivModRewriter.h
```diff
@@ -0,0 +1,111 @@
...
+    struct DivModPair {
+        PTRef div;
+        PTRef mod;
+    };
+
+    std::unordered_map<std::pair<PTRef, PTRef>, DivModPair, PTRefPairHash> divModCache;
+    vec<PTRef> definitions;
+
+    const char * divPrefix = ".div";
+    const char * modPrefix = ".mod";
```
aehyvari, line 31: Define the strings in LIATheory?

src/rewriters/DivModRewriter.h
```diff
@@ -0,0 +1,111 @@
...
+                auto upperBound = abs(divisorVal) - 1;
+                // dividend = divVar * divisor + modVar
+                // 0 <= modVar <= |dividend| - 1
+                definitions.push(logic.mkAnd(
+                        logic.mkEq(dividend, logic.mkNumPlus(logic.mkNumTimes(divisor, divVar), modVar)),
+                        logic.mkAnd(
+                                logic.mkNumLeq(logic.getTerm_NumZero(), modVar),
+                                logic.mkNumLeq(modVar, logic.mkConst(upperBound))
+                        )
+                ));
```
aehyvari, line 73: What if we move `definitions` to LIATheory and have this stuff in a later phase.  Also the cache could live there, this could help to handle incrementality.

blishko: I wanted to have independent rewriter that users (me) can use on their own.
I think we can achieve the same effect as what you suggest by having a single rewriter instance in the `LIATheory` and not recreate it for every call to the simplification.

src/rewriters/DivModRewriter.h
blishko: I think this would create a dependency of `DivModRewriter` on `LIATheory` that does not really make sense. I would keep the rewriter as much independent as possible.
If these were to move somewhere, I would move them to `LIALogic`, that dependency already exists and is necessary.

src/logics/LIALogic.cc
```diff
@@ -249,3 +204,67 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
...
+    if (sym == get_sym_Int_MOD())
+        return mkIntMod(terms);
+    if (sym == get_sym_Int_DIV())
+        return mkIntDiv(terms);
+    return LALogic::insertTerm(sym, terms);
+}
+
+PTRef LIALogic::mkIntDiv(vec<PTRef> const & args) {
+    if (args.size() != 2) { throw OsmtApiException("Incorrect number of arguments for DIV operator"); }
+    return _mkIntDiv(args);
```
aehyvari, line 218: Div is defined as `(div Int Int Int :left-assoc)`.  Maybe it wouldn't be too much work to split it here into binary calls?

src/logics/LIALogic.cc
```diff
@@ -249,3 +204,67 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
...
+    return _mkIntDiv({dividend, divisor});
+}
+
+PTRef LIALogic::_mkIntDiv(const vec<PTRef> & args) {
+    assert(args.size() == 2);
+    PTRef dividend = args[0];
+    PTRef divisor = args[1];
+    if (not isConstant(divisor)) {
+        throw LANonLinearException("Only division by constant is permitted in linear arithmetic!");
+    }
```
aehyvari, line 231: We should do the corresponding throws in `isNumZero(divisor)`

src/logics/LIALogic.cc
```diff
@@ -249,3 +204,67 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
...
+
+PTRef LIALogic::mkIntMod(PTRef dividend, PTRef divisor) {
+    return _mkIntMod({dividend, divisor});
+}
+
+PTRef LIALogic::_mkIntMod(vec<PTRef> const & args) {
+    assert(args.size() == 2);
+    PTRef dividend = args[0];
+    PTRef divisor = args[1];
+    if (not isNumConst(divisor)) { throw OsmtApiException("Divisor must be constant in linear logic"); }
```
aehyvari, line 257: Throw when `isNumZero(divisor)`

src/logics/LRALogic.cc
```diff
@@ -131,22 +133,32 @@ LRALogic::LRALogic() :
...
-const SymRef LRALogic::get_sym_Num_ONE () const {return sym_Real_ONE;}
-const SymRef LRALogic::get_sym_Num_ITE () const {return sym_Real_ITE;}
-const SRef LRALogic::get_sort_NUM () const {return sort_REAL;}
+PTRef LRALogic::mkRealDiv(const vec<PTRef>& args)
+{
+    SimplifyConstDiv simp(*this);
+    vec<PTRef> args_new;
+    SymRef s_new;
+    if (args.size() != 2) {
+        throw OsmtApiException("Division operation requries exactly 2 arguments");
```
aehyvari, line 142: We could fix this as well to comply to the standard (` (/ Real Real Real :left-assoc)`)

src/rewriters/DivModRewriter.h
aehyvari: Ok, maybe we could move them there.

src/rewriters/DivModRewriter.h
aehyvari: Ok, that sounds good.

test/unit/test_ArithmeticExpressions.cpp
```diff
@@ -46,7 +46,7 @@ TEST_F(ArithmeticExpressions_test, test_One_Arg_Multiplication)
     PTRef two = logic.mkConst("2");
     PTRef one = logic.mkConst("1");
     PTRef a = logic.mkNumVar("a");
-    PTRef div = logic.mkNumDiv(a, two);
+    PTRef div = logic.mkRealDiv(a, two);
```
aehyvari, line 49: We could maybe have unit tests for `LALogic::mkNumDiv` for both cases, where the instance is actually `LRALogic` and when it is `LIALogic`.

aehyvari (commented):
It looks like we're almost there, I just left a few comments!

src/logics/LALogic.h
```diff
@@ -49,32 +49,32 @@ class LALogic: public Logic
...
+    bool           isTheoryEquality(PTRef tr) const override { return isNumEq(tr); }
+    bool           isAtom(PTRef tr) const override { return isNumEq(tr) || isNumLt(tr) || isNumGt(tr) || isNumLeq(tr) || isNumGeq(tr) || Logic::isAtom(tr); }
+    bool           isUF(PTRef tr) const override { return isUF(term_store[tr].symb()); }
+    bool           isUF(SymRef sr) const override { return !sym_store[sr].isInterpreted(); }
     const char*    getDefaultValue(const PTRef tr) const override;
     PTRef          getDefaultValuePTRef(const SRef sref) const override;
 
 
     virtual const SymRef get_sym_Num_TIMES () const =0;
-    virtual const SymRef get_sym_Num_DIV () const =0;
```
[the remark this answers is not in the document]
blishko: Yes, I will bring this back.

src/logics/LALogic.h
```diff
@@ -86,36 +86,37 @@ class LALogic: public Logic
...
-    bool isNumPlus(SymRef sr) const;
-    virtual bool isNumPlus(PTRef tr) const;
-    bool isNumMinus(SymRef sr) const;
-    virtual bool isNumMinus(PTRef tr) const ;
-    bool isNumNeg(SymRef sr) const ;
-    virtual bool isNumNeg(PTRef tr) const ;
-    bool isNumTimes(SymRef sr) const;
-    virtual bool isNumTimes(PTRef tr) const ;
-    bool isNumDiv(SymRef sr) const;
-    virtual bool isNumDiv(PTRef tr) const;
```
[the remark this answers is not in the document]
blishko: These are still available (just below). They are no longer virtual, but rather rely on the virtual `get_sym_Num_XXX` methods.

src/logics/LALogic.h
```diff
@@ -150,7 +151,6 @@ class LALogic: public Logic
     virtual bool isLinearTerm(PTRef tr) const;
     virtual bool isLinearFactor(PTRef tr) const;
     virtual void splitTermToVarAndConst(const PTRef &term, PTRef &var, PTRef &fac) const;
-    virtual PTRef normalizeSum(PTRef sum);
```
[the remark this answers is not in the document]
blishko: Yes, `sumToNormalizedInequality` is doing that. Both get normalized to `0 <= z-x-y`. I'll check if we have tests for this.

src/logics/LRALogic.cc
blishko: I am not sure I understand what that means. If I write `x / 2 / 3`, it should be read as `(x/2)/3`, right? This means we can accept more than two arguments, but each except the first one must be constant, right? Can we do it in another PR?

src/rewriters/DivModRewriter.h
blishko: I am not sure anymore about moving them out of here.
Do we plan to use them anywhere else?
I now see it more like an internal thing of the rewriter.

src/logics/LALogic.h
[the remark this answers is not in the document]
blishko: Added a test for that for both LIA and LRA

src/logics/LIALogic.cc
blishko: Added.

aehyvari:
Would it be a problem if another rewriter used the same symbols
unintentionally?

src/rewriters/DivModRewriter.h
aehyvari: Ok, as discussed, let's go with the current solution.

src/logics/LALogic.h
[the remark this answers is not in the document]
aehyvari: This was resolved in gitter.

src/logics/LIALogic.cc
aehyvari: We'll do this later in a separate PR.

src/logics/LRALogic.cc
aehyvari: Ok, another PR.

src/logics/LIALogic.h
aehyvari: This was already resolved.

aehyvari (approved):

blishko:
I have made a note about support for `/` and `div` with multiple arguments here: https://github.com/usi-verification-and-security/opensmt/issues/146#issuecomment-811860257

Seems we have a couple of operators that should support it.

aehyvari (approved):

PR #223, opened 2021-02-02
Throwing non-linear exception on division by non-constant

blishko:
This PR introduces defensive behaviour in `LALogic::mkNumDiv` to not just rely on correct input, but throw exception if the arguments do not meet the requirements.


aehyvari (approved):
Looks good.  Maybe we should come up with a uniform way of phrasing error reporting, but we'll delegate this question to our future communications department :)

blishko:
> Looks good. Maybe we should come up with a uniform way of phrasing error reporting, but we'll delegate this question to our future communications department :)

Good idea! Let me know when there are guidelines I can follow and I will do that!

PR #224, opened 2021-02-03
Adding method for creating a unique variable

blishko:


src/logics/Logic.cc
```diff
@@ -956,6 +958,11 @@ PTRef Logic::mkVar(SRef s, const char* name) {
     return ptr;
 }
 
+PTRef Logic::mkUniqueVar(SRef s) {
+    std::string uniqueName = "@" + std::to_string(uniqueVarCount++);
```
aehyvari, line 962: Put the "@" in some const char near line 65 and use that instead of hardcoding it in the function.

aehyvari (approved):
Looks good!

src/logics/Logic.cc
blishko: Fixed.

blishko:
As discussed, the meaning is changing from a unique variable to a unique abstract value (for internal solver use).

aehyvari (approved):
I think this one's good to go.

PR #227, opened 2021-02-11
Fix Booleans inside UF

### LINKED ISSUE #226: Unsound result on QF_UF formula with booleans inside uninterpreted functions
On the following example:
```
(set-logic QF_UF)
(declare-fun uf6 (Bool) Bool)
(declare-fun x () Bool)
(declare-fun z () Bool)
(assert (uf6 x))
(assert (not (uf6 z)))
(assert (not (uf6 (not z))))
(check-sat)
```

OpenSMT answers SAT, but the formula is unsatisfiable. Z3 correctly reports UNSAT.

blishko:
Fixes #226


regression/QF_UF/issue226b.smt2
```diff
@@ -0,0 +1,11 @@
...
+(declare-fun uf6 (Bool) Bool)
+(declare-fun x () Bool)
+(declare-fun y () Bool)
+(declare-fun z () Bool)
+(assert (uf6 x))
+(assert (not (uf6 (or y z))))
+(assert (not (uf6 (or y (not z)))))
+(assert (not (uf6 (or (not y) z))))
+(assert (not (uf6 (or (not y) (not z)))))
+(check-sat)
```
aehyvari, line 11: This looks very familiar, we probably already have a regression test for it.  But in this case redundancy probably is not an issue.

blishko: May be, but we probably do not have a test exactly like this, because it turns out the solver was previously giving wrong answer also for this slightly more complicated  case.

Looking at our tests, I realised I could have add the `set-info` command to mark the expected answer.

aehyvari: Haha yes, looks like that, I missed the first line asserting `(uf6 x)`.  I actually didn't want to publish this comment yet, I thought I put it as a pending comment while writing the review.  But anyway, we should keep the regression test. I had just a feeling that I've written a pigeon hole problem inside an uninterpreted function to test these cases back then.

aehyvari (approved):

PR #228, opened 2021-02-11
Generic rewriter and visitor for Pterms

blishko:
This PR introduces a template for rewriting Pterms, as well as a template for visitor which collects information but does not modify the terms.

As an example, the rewriter is instantiated for substitution, as a replacement for `Logic::varsubstitute` and visitor is instantiated for collecting boolean terms that appear in UF terms.

The templates are inspired by Z3's implementation of the pattern:
https://github.com/Z3Prover/z3/blob/4c3c15c015a053300ca979b681793a9730c9f313/src/ast/rewriter/rewriter.h
https://github.com/Z3Prover/z3/blob/4c3c15c015a053300ca979b681793a9730c9f313/src/ast/rewriter/rewriter.txt

The templates are parametrized by a `RewriterConfig` class which (currently) is required to provided 2 methods:

- `previsit` called when a term is first seen. `Config` can return false if the children should not be visited.
- `rewrite` given an input term, should return the rewritten version of the term.

Similarly `VisitorConfig` class is required to provide `previsit` and `visit` methods, where `visit`
 is called after the children have already been visited.


src/rewriters/ArithmeticEqualityRewriter.h
```diff
@@ -0,0 +1,43 @@
...
+    PTRef rewrite(PTRef term) override {
+        if (logic.isNumEq(term)) {
+            Pterm const  & p = logic.getPterm(term);
+            PTRef a1 = p[0];
+            PTRef a2 = p[1];
+            vec<PTRef> args;
+            args.push(a1); args.push(a2);
+            PTRef i1 = logic.mkNumLeq(args);
+            PTRef i2 = logic.mkNumGeq(args);
+            logic.markSplitInequality(i1); // MB: this information should not be stored in Logic!
```
blishko, line 27: This information is only used in Lookahead solving. I think we should find a way to move it out of Logic.

blishko:
I have added `ArithmeticEqualityRewriter` for splitting equalities in linear arithmetic.
I think I am quite happy with it for now.

src/rewriters/ArithmeticEqualityRewriter.h
aehyvari: Ok, I'll commit a PR to fix this once we merge the current one.

aehyvari (approved):

PR #229, opened 2021-03-02
Do not copy termColors in iteration

aehyvari:
Apple clang version 12.0.0 complains that currently the iterator in question needs to copy a value.  I believe that's correct.  This PR fixes the iteration to use const & instead of copying.


blishko (approved):
I think this might have been intentional, since it is just a pair of PTRef and enum, but if the compiler complains, I am fine with changing it.

PR #230, opened 2021-03-02
Move parallel solving related splitting information from logic to theory

aehyvari:
As part of the process to clean Logic from extra things this PR suggests how to move the information that a linear inequality was derived by splitting an equality into inequalities from Logic to Theory.  This information is currently only used by the code that handles parallel solving, and even there only as an option.


src/rewriters/ArithmeticEqualityRewriter.h
```diff
@@ -8,10 +8,15 @@
 #include "Rewriter.h"
 
 class EqualityRewriterConfig : DefaultRewriterConfig {
+    std::unique_ptr<Map<PTRef,bool,PTRefHash>> notOkToPartition;
     LALogic & logic;
-
 public:
-    EqualityRewriterConfig(LALogic & logic): logic(logic) {}
+    EqualityRewriterConfig(LALogic & logic): logic(logic) { notOkToPartition = std::unique_ptr<Map<PTRef,bool,PTRefHash>>(new Map<PTRef,bool,PTRefHash>()); }
```
blishko, line 14:
```suggestion
    EqualityRewriterConfig(LALogic & logic): logic(logic), notOkToPartition(new Map<PTRef,bool,PTRefHash>()) {}
```
and put `notOkToPartition` after `logic` in the declaration of members.

src/rewriters/ArithmeticEqualityRewriter.h
```diff
@@ -8,10 +8,15 @@
...
+
+    std::unique_ptr<Map<PTRef,bool,PTRefHash>> getAndClearNotOkToPartition() {
+        auto tmp = std::unique_ptr<Map<PTRef,bool,PTRefHash>>(new Map<PTRef,bool,PTRefHash>());
+        std::swap(tmp, notOkToPartition); return tmp;
+    }
```
blishko, line 19: I am wondering if we need to clear it like that.
From the current use case, I would say no.
(I am just trying to avoid the allocation if it is not necessary.)

src/logics/Theory.h
```diff
@@ -173,14 +173,16 @@ class Theory
     virtual bool            simplify(const vec<PFRef>&, PartitionManager& pmanager, int) = 0; // Simplify a vector of PushFrames in an incrementality-aware manner
     SubstitutionResult      computeSubstitutions(PTRef fla);
     void                    printFramesAsQuery(const vec<PFRef> & frames, std::ostream & s) const;
-    virtual                ~Theory()                           {};
+    virtual bool            okToPartition(PTRef)    = 0;
```
blishko, line 176:
```suggestion
    virtual bool            okToPartition(PTRef)  { return true; }
```
How about having the default implementation? Then we do not need to add it to the `UFTheory` and `CUFTheory`.

blishko (commented):
Looks good, I would just do some small changes.

src/rewriters/ArithmeticEqualityRewriter.h
aehyvari: Ok.

src/rewriters/ArithmeticEqualityRewriter.h
aehyvari: I think it's safer this way.  The rewriter resets its state after this call to the initial state, and I think it's confusing otherwise (i.e., resetting to a state where `notOkToPartition` is `nullptr`).

src/logics/Theory.h
aehyvari: That's a good idea.

src/rewriters/ArithmeticEqualityRewriter.h
blishko: OK, I agree.

blishko (changes requested):

src/logics/LIATheory.cc
```diff
@@ -13,7 +13,9 @@ bool LIATheory::simplify(const vec<PFRef>& formulas, PartitionManager &pmanager,
         for (int i = 0; i < flas.size(); ++i) {
             PTRef & fla = flas[i];
             PTRef old = flas[i];
-            fla = ArithmeticEqualityRewriter(lialogic).rewrite(old);
+            auto arr = ArithmeticEqualityRewriter(lialogic);
+            fla = arr.rewrite(old);
+            notOkToPartition = arr.getAndClearNotOkToPartition();
```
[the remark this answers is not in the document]
aehyvari: I agree.

src/logics/UFLRATheory.cc
```diff
@@ -1,21 +1,29 @@
 #include "UFLRATheory.h"
+#include "TreeOps.h"
 #include "ArithmeticEqualityRewriter.h"
 
 bool UFLRATheory::simplify(const vec<PFRef>& formulas, PartitionManager &pmanager, int curr)
 {
     auto & currentFrame = pfstore[formulas[curr]];
     // Take care of UF simplifications as well
+    ArithmeticEqualityRewriter rewriter(lralogic);
```
blishko, line 9: This rewriter is not used, the branches still create their own rewriters. Also, `notOkToParition` is assigned only in the if branch, but not in the else branch.

blishko (commented):
I think it's good, just the `UFLRATheory` has still some problems.

src/logics/UFLRATheory.cc
aehyvari: Right, I fixed this and tried to make also the LATheory simplification a little more economic in lines of code.

blishko (approved):
I think it's good now!

PR #232, opened 2021-03-10
Osx build

aehyvari:
This PR adds a CircleCI job for building and testing osmt Release in MacOS Big Sur.  The workflow downloads homebrew, installs the dependency packages, and sets up the environment correctly, which is the presumed workflow of our users I guess.

Downloading homebrew takes almost 4 minutes which is a little annoying.  As a fix, I tried CircleCI's cache feature for storing the homebrew installation, but could not make it work.  I'm guessing that this is because unpacked cache of /usr/local is 1.3 GB and there's maybe some limit, but I'm not sure.  The error comes from `tar`.


blishko (approved):
Looks good!
We should try to make the cache working, it would decrease the build time, but also save some internet traffic, I guess.

aehyvari:
Maybe, but I'm not sure if it's doable.  Internet doesn't do it based on my googling.

Issue #233, opened 2021-03-14
pipe mode is broken


jwaldmann:
Pipe mode (reading from stdin) is useful (e.g., https://github.com/hgoes/smtlib2/tree/master/backends/pipe , I want it for https://gitlab.imn.htwk-leipzig.de/waldmann/pure-matchbox/-/blob/master/src/LP/SMT.hs )

current implementation seems broken, as I am getting all sorts of scanner/parser errors, e.g.,
```
echo "(set-logic QF_LRA)(declare-fun X () Real)(assert (< 1 X)) (check-sat) (get-value (X))" | ./build/opensmt -p
sat
At interactive input: syntax error, unexpected end of file, expecting ')'
(error "scanner")
```
I do have a fix, and I can submit a PR.

jwaldmann:
Declarations  of variables `par` and `i` must be hoisted outside the inner loop.

jwaldmann:
NB: the scanner tries to count the nesting level of s-expressions by counting characters `(` and `)`, this will break as soon as parens appear in comments, or in string literals.

aehyvari:
Thanks @jwaldmann , this looks very good!  I'll try to have a closer look by tonight.

aehyvari:
Hi @jwaldmann, apologies, it took a while.  Let us know if there are more issues!

PR #234, opened 2021-03-30
Fix pipe mode

aehyvari:
This PR fixes issue #233 by applying the patch from @jwaldman, cleaning the code a little, attempting to fix issues with parsing parentheses in comments, strings, and quoted symbols, and adding some regression tests.


aehyvari:
> I don't see any problem straightaway, so I am going to trust you on this one :)

Ok, if there's a serious problem I'm sure we'll notice :)

> Have you run the incremental SMT-LIB benchmarks in the pipe mode? It would be good to know that that is still working.

Good point.  I just ran the incremental regression tests in this mode and they all work, so I'll go ahead and assume that also the smt-comp incremental setup works...

PR #236, opened 2021-04-02
Avoid unnecessary allocations and deallocation of GNU MPs

blishko:
Playing with [heaptrack](https://github.com/KDE/heaptrack) I noticed that when we need to fall back to `mpq` representation of rationals, OpenSMT makes a huge amount of small allocations and deallocations related to handling this representation.

While trying to investigate that, I found a couple of places in the code where we make unnecessary copies of `FastRationals` or create unnecessary temporaries. This PR proposes some changes to avoid that.

### DIFF
--- a/src/common/FastRational.h
+++ b/src/common/FastRational.h
@@ -184,16 +184,21 @@ private:
             setMpqMemoryAllocated();
         }
     }
+
+    bool fitsWord() const
+    {
+        assert(not wordPartValid() and mpqPartValid()); // Do not call this method if word part is already valid
+        return mpz_fits_sint_p(mpq_numref(mpq)) and mpz_fits_uint_p(mpq_denref(mpq));
+    }
+
     //
     // Tries to convert the current rational
     // stored in mpq into word/uword
     //
     void try_fit_word()
     {
-        assert( mpqPartValid() );
-        assert(!wordPartValid()); // MB: It does not make sense to call this method if word is valid
-        if ( mpz_fits_sint_p(mpq_numref(mpq))
-             && mpz_fits_uint_p(mpq_denref(mpq))) {
+        assert(not wordPartValid() and mpqPartValid()); // Do not call this method if word part is already valid
+        if (fitsWord()) {
             num = mpz_get_si(mpq_numref(mpq));
             den = mpz_get_ui(mpq_denref(mpq));
             setWordPartValid();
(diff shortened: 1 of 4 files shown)


blishko:
[comparison.pdf](https://github.com/usi-verification-and-security/opensmt/files/6250575/comparison.pdf)
[comparison_onlyHeizmann.pdf](https://github.com/usi-verification-and-security/opensmt/files/6250577/comparison_onlyHeizmann.pdf)

Here are the results from the cluster for LRA. I think the improvement is mainly thanks to the changes in the [last commit](https://github.com/usi-verification-and-security/opensmt/pull/236/commits/d0c157caed1170de5ddb2581eefcea1bfce51d7b).
The overall improvement is only ~3%, but on the benchmarks from `2017-Heizmann-UltimateInvariantSynthesis`, the improvement is ~12%.

src/tsolvers/lasolver/Delta.h
```diff
@@ -99,7 +100,8 @@ class Delta {
     inline friend bool operator>=(const Real & c, const Delta & a);
 
     // Arithmetic overloadings
-    inline friend Delta operator+=(Delta & a, const Delta & b);
+    Delta& operator+=(const Delta & b);
+    Delta& operator+=(Delta && b);
 
     inline friend Delta operator-=(Delta & a, const Delta & b);
```
aehyvari, line 106: Should we do the above for `-=`, to turn it into a member operator and add the rvalue reference version?

src/common/FastRational.h
```diff
@@ -385,6 +387,10 @@ class FastRational
         }
     }
 
+    bool isOneHeuristic() const {
```
aehyvari, line 390: Maybe this could be just called `isOne()` ?

aehyvari (commented):
This looks good!  I left two small comments.

src/tsolvers/lasolver/Delta.h
blishko: Yes, that makes sense.

src/common/FastRational.h
blishko: I made this method because I needed a method that returns `true` if we can be sure it is `1`, but if it returns `false`, it does not necessarily means that it is not `1`.
To have a proper `isOne()` method, I would have to add also the check in case `wordPartValid()` returns `false`, similarly to `isZero()`.
On the other hand, I think we try to enforce the invariant that if the value fits into the wordPart that it always has `wordPartValid()`. I am just not sure if we do it in all places and that it is indeed an invariant in our current implementation.
So I went with this shortcut. But maybe the best solution would be just to go with proper `isOne` method. What do you think?

aehyvari: Ok, yes.  It should be the case that whenever this external `isOneHeuristic()` method is called, the invariant "if the number fits in word representation, then `wordPartValid()` returns `true`" holds.  Maybe we should add an assert and check if this breaks.  It'd be useful to know if it doesn't hold.

src/tsolvers/lasolver/Delta.h
blishko: Fixed. Although it seems operator -= is not used in our codebase.

aehyvari (approved):

PR #237, opened 2021-04-22
Small refactoring of Logic::isLit

blishko:
In `Logic` we have methods `isLit` and `isAtom` with custom implementation.
I believe `isLit` should just rely on `isAtom`.

### DIFF
--- a/src/cnfizers/Cnfizer.cc
+++ b/src/cnfizers/Cnfizer.cc
@@ -297,7 +297,7 @@ bool Cnfizer::checkCnf (PTRef formula)
 
 bool Cnfizer::checkConj (PTRef e)
 {
-    if (logic.isLit (e)) // A Boolean constant
+    if (isLiteral(e)) // A Boolean constant
         return true;
 
     Pterm &and_t = logic.getPterm (e);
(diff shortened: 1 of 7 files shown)


aehyvari (commented):
It's good.  Maybe we could move the declarations in the header next to each other. isLit is also not the best name...  Something like isLitOfAtom could be more descriptive.

blishko:
> It's good. Maybe we could move the declarations in the header next to each other. isLit is also not the best name... Something like isLitOfAtom could be more descriptive.

My definition of a literal is that it is an atom or a negation of an atom. Do you have a different definition?

aehyvari:
> My definition of a literal is that it is an atom or a negation of an atom. Do you have a different definition?

Maybe...  The difference being that a PTRef of a literal that corresponds to a Tseitin transformation's "definition literal" is a literal in the common language, but it's not a literal in how we've defined it now.  My suggestion there is to try to make the distinction also in the naming of the related functions in the PTRef level between the two.  But if you feel strongly otherwise I'm willing to be convinced.

blishko:
OK, I see your point. I was not thinking about Tseitin literals, I was working with the theoretical definition of literals in logical terminology.
We can try to agree on some appropriate naming.

But do we have/need such method that you propose? The `isLit` method, as originally was returning `false` for logical connectives, so it was matching my definition, not Tseitin literals.

aehyvari (approved):

PR #239, opened 2021-04-26
Unifying style, small refactorings

aehyvari:
Now that we're cleaning `Cnfizer` I also added some small things that maybe could be in line with the spirit of the original PR.
- use list comprehension for loops when possible
- make a pass on uniforming the indentation style in if - else -constructs and white spaces in function calls
- rename the current frame variable more appropriately
- Clean the constructor (and destructor) a little


blishko (approved):
LGTM!

PR #241, opened 2021-04-27
Refactoring of retrieving top-level formulae

blishko:
As a preparation for dealing with distinct in the pre-processing phase, this PR pulls the method to retrieve top-level formulae out of Cnfizer. This is needed to distinguish top-level distincts from other distincts in the formula.

This PR also fixes TermVisitor to prevent duplicate calls to `previsit` for the same term if it returned `false` the first time.

### DIFF
--- a/src/cnfizers/Cnfizer.cc
+++ b/src/cnfizers/Cnfizer.cc
@@ -26,6 +26,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 #include "Cnfizer.h"
 #include "SimpSMTSolver.h"
+#include "TreeOps.h"
 #include "OsmtInternalException.h"
 
 #include <queue>
@@ -137,9 +138,8 @@ lbool Cnfizer::cnfizeAndGiveToSolver(PTRef formula, FrameId frame_id)
         assert(pmanager.getPartitionIndex(formula) != -1);
         currentPartition = pmanager.getPartitionIndex(formula);
     }
-    vec<PTRef> top_level_formulae;
     // Retrieve top-level formulae - this is a list constructed from a conjunction
-    retrieveTopLevelFormulae (formula, top_level_formulae);
+    vec<PTRef> top_level_formulae = topLevelConjuncts(logic, formula);
     assert (top_level_formulae.size() != 0);
     TRACE("Top level formulae:")
     TRACE_FLA_VEC(top_level_formulae)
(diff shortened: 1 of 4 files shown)


src/common/TreeOps.h
```diff
@@ -57,6 +57,7 @@ class TermVisitor {
             PTRef currentRef = currentEntry.term;
             if (not cfg.previsit(currentRef)) {
                 toProcess.pop_back();
+                done[Idx(logic.getPterm(currentRef).getId())] = 1;
```
blishko, line 60: This is needed to prevent duplicate calls to `previsit` if it returned `false` the first time.
I think this makes sense regardless of the other changes.

blishko:
@aehyvari , I wanted to reuse the `TermVisitor` infrastructure. This may be using it somewhat unconventionally, but I think it makes sense. I needed to adjust it a bit, but I think that change makes sense regardless of whether or not we go with this implementation of `retrieveTopLevelConjuncts`.

Also, any suggestions for more tests?

src/cnfizers/Cnfizer.cc
```diff
@@ -427,33 +427,6 @@ bool Cnfizer::assertClause (PTRef f)
...
-
-    to_process.push (root);
-
-    while (to_process.size() != 0) {
-        PTRef f = to_process.last();
-        to_process.pop();
-        Pterm &cand_t = logic.getPterm (f);
-
-        if (logic.isAnd (f)) {
-            for (int i = cand_t.size() - 1; i >= 0; i--) {
```
blishko, line 447: For some reason this was iterating the conjuncts from the end. This is going to change now, so I should run the experiments to check the effect on run time.

aehyvari: Probably some old stuff from that I added to try to reproduce opensmt1 behaviour while hunting performance bugs long time ago.  It's ok to change it.

aehyvari (commented):
To me this looks like a very good idea for now.

blishko:
Results from LRA shows that this change should be fine.
[comparison.pdf](https://github.com/usi-verification-and-security/opensmt/files/6392684/comparison.pdf)

aehyvari:
> Results from LRA shows that this change should be fine.

Yes.  Are you planning of running others?  Maybe it's safer to check QF_LIA & QF_UF as well, since especially QF_UF might have very different structures.

blishko:
Ok, I'll check QF_UF. I would skip QF_LIA for now, there are too many benchmarks there that OpenSMT cannot solve yet, sol it takes quite a long time if most of nodes in the cluster are occupied. 
Unless you really insist.

blishko:
QF_UF seems also OK:
[comparison_uf.pdf](https://github.com/usi-verification-and-security/opensmt/files/6393671/comparison_uf.pdf)
The one benchmark that is the difference was solved in 569 seconds in master and not solved within 600 seconds in the new version. I checked on my local, in master solved in 459 seconds, in new version solved in 485 seconds. So seems slightly slower on this benchmark, but seems like a random change, I would not worry about it.

aehyvari (approved):
Then let's go with it!

PR #242, opened 2021-05-04
Fix handling of distincts that are not top-level

blishko:
The current implementation of `Egraph` solver does not handle correctly if `distinct` term (with more than two elements) is asserted negatively.
This PR adds a rewriting step in the preprocessing phase to fully expand all distincts that are not top-level conjunct in the queried formula. This ensures that no distinct will be asserted negatively in the `Egraph` solver.

Explicit check is also added in the `Egraph` to detect this situation and an exception is thrown if distinct would be asserted negatively at any point. This prevents the solver from giving an unsound result.


blishko:
@aehyvari , I want to add some unit tests and check what happens in LRA, but you can already start reviewing this code.

src/rewriters/DistinctRewriter.h
```diff
@@ -0,0 +1,87 @@
...
+    bool previsit(PTRef term) override {
+        return logic.hasSortBool(term);
+    }
+
+    PTRef rewrite(PTRef ptr) override {
+        if (logic.isDisequality(ptr) and doRewriteDistinct(ptr)) {
+            Pterm const & term = logic.getPterm(ptr);
+            vec<PTRef> args;
+            args.capacity(term.nargs());
+            for(PTRef arg : term) {
```
aehyvari, line 27: space

aehyvari (commented):
So in the end the suggestion is to not do this from Cnfizer but from Theory.  I think that's a clean design.

blishko:
> So in the end the suggestion is to not do this from Cnfizer but from Theory. I think that's a clean design.

Yeah, that's what we agreed on in our last discussion and now I also think it's cleaner.

src/rewriters/DistinctRewriter.h
blishko: fixed

aehyvari (approved):

PR #243, opened 2021-05-05
Fixed handling of distincts in LRA and LIA

blishko:
Similar to #242, which fixed handling of `distinct` in UF, this PR does the same for LRA and LIA, with the difference that _all_ distincts are rewritten (including the top-level ones).


aehyvari (commented):
This looks good.  Should we have some interpolation regression tests?

aehyvari (approved):

PR #244, opened 2021-05-11
Models

aehyvari:
This PR suggests a partial redesign of the model system motivated by the need to produce models for EUF.

 - The substitution - rewriting part of the code has some changes resulting in more substitutions being tracked.
 - The substitution handling in model production is delegated to theory solvers
 - The Logic::TFun class is refactored into two standalone classes for template functions and template function signatures
 - The model printing complies to the smtlib2 standard
 - The UF models for functions are stored in a tree structure that avoids (hopefully) efficiently insertion of duplicate models

For now the models are tested for QF_LIA, QF_LRA, and QF_UF, but only on a relatively short, 10 s timeout running on unoptimised debug code so I can't rule out surprises.  Before merging it would make sense to do a bigger test in cluster.

There seems to be in general an overhead because of the substitution-rewriting changes.  However, in all three tested logics the number of instances solved with the new substitution is bigger than with the old substitution.

### QF_UF
[This plot](https://user-images.githubusercontent.com/8374962/117820523-ea2c9580-b26a-11eb-9687-e8435d156a26.png) shows the overhead in pre-processing only code (checking disabled).  There is a clear overhead.

It looks like the change has a positive effect though on the overall solving, although this is debatable.  [See here](https://user-images.githubusercontent.com/8374962/117820768-24963280-b26b-11eb-825d-5fa804812299.png) for the comparison, and in particular note the long-running SAT instances, the speedup in a few UNSAT instances.  The timeouts in master solved in uf-models might be a random effect, a result of non-equal timeouts, or a real speed-up with the new simplifications.

### QF_LIA
[Comparison in QF_LIA](https://user-images.githubusercontent.com/8374962/117994084-8f656d80-b340-11eb-9ec9-89c626d871e8.png) shows that the substitutions make a big difference.  Overall result is positive in terms of number of solved instance.

### QF_LRA
For linear real arithmetics [the results](https://user-images.githubusercontent.com/8374962/117995200-6b565c00-b341-11eb-803f-89fe0167130e.png) suggest a %-overhead.  I can't really explain this, since the pre-processing is just an initial pass and therefore should be a constant overhead.  I think I'm over-interpreting the plot.


src/logics/Theory.cc
```diff
@@ -52,26 +52,27 @@ Theory::SubstitutionResult Theory::computeSubstitutions(const PTRef fla)
...
-        // remember the substitutions
-        vec<Map<PTRef,PtAsgn,PTRefHash>::Pair> newsubsts_vec;
-        newsubsts.getKeysAndVals(newsubsts_vec);
-        for (int i = 0; i < newsubsts_vec.size(); ++i) {
-            PTRef key = newsubsts_vec[i].key;
-            if (!allsubsts.has(key) && newsubsts_vec[i].data.sgn == l_True) {
-                allsubsts.insert(key, newsubsts_vec[i].data);
+        auto newRootAndSubsts(Substitutor<PtAsgn>(getLogic(), newsubsts).rewrite(root));
+        PTRef new_root = newRootAndSubsts.first;
+        auto usedSubstitutions(std::move(newRootAndSubsts.second));
```
aehyvari, line 65: Here's essentially why I felt that I need to change the interface of the Substitutor.  The substitutions that are computed are used by the models (and indeed also by this very substitution loop).  It might be good to write some templates to avoid having to deal every time with the returned substitution map in places where it's not needed.

src/logics/Logic.h
```diff
@@ -93,37 +51,33 @@ class Logic {
     Map<SymRef,bool,SymRefHash,Equal<SymRef> >      ites;
     Map<SRef,SymRef,SRefHash>                       sortToIte;
     Map<SRef,bool,SRefHash,Equal<SRef> >            ufsorts;
+    Map<SRef,PTRef,SRefHash,Equal<SRef>>            defaultValueForSort;
 
+    bool isKnownToUser(SymRef sr) const { return getSymName(sr)[0] != s_abstract_value_prefix[0]; }
```
blishko, line 56: I am wondering if `s_abstract_value_prefix` should not be just a single `char` instead of `const char*`.

src/logics/Logic.h
```diff
@@ -93,37 +51,33 @@ class Logic {
...
+        bool has(const std::string& name) const { return defined_functions.find(name) != defined_functions.end(); }
 
-        void insert(const char* name, TFun const & templ) {
+        void insert(const std::string& name, TemplateFunction && templ) {
             assert(not has(name));
-            defined_functions_names.push();
-            defined_functions_names.last() = strdup(name);
-            defined_functions[defined_functions_names.last()] = templ;
+            defined_functions_names.emplace_back(name);
+            defined_functions[defined_functions_names.back()] = std::move(templ);
```
blishko, line 70:
```suggestion
            defined_functions[name] = std::move(templ);
```

src/minisat/mtl/Map.h
```diff
@@ -136,6 +136,12 @@ class Map {
 
     // PRECONDITION: the key must *NOT* exist in the map.
     void insert (const K& k, const D& d) { if (checkCap(size+1)) rehash(); _insert(k, d); size++; }
+    /**
+     * Check if k exists in the map.  If it exists, return h[k] i d.  If it does not, d remains unchanged
```
blishko, line 140: What does `return h[k] i d` mean?

src/models/Model.cc
```diff
@@ -3,6 +3,30 @@
...
+       {
+           SymRef sr = entry.first;
+           const TemplateFunction & templFun = entry.second;
+           if (not logic.isUF(sr)) { std::cout << "Not UF: " << logic.getSymName(sr) << std::endl; return false; }
+           const Symbol & s = logic.getSym(sr);
+           if (templFun.getName() != logic.getSymName(sr)) { std::cout << "Name doesn't match" << std::endl; return false; }
+           if (s.nargs() != templFun.getArgs().size_()) { std::cout << "Argument number doesn't match" << std::endl; return false; }
+           if (logic.getSortRef(sr) != templFun.getRetSort()) { std::cout << "Return sorts don't match" << std::endl; return false; }
+           for (auto i = 0; i < (int)s.nargs(); i++)
+               if (s[i] != logic.getSortRef(templFun.getArgs()[i])) { std::cout << "argument sort doesn't match" << std::endl; return false; }
```
blishko, line 26: Should these be `std::cerr` instead of `std::cout`?

blishko (commented):
I made a first pass with some small comments.
I have to look at the changes for the substitutions in more detail.
I am not very keen on having the templated version for both `PTRef and PTAsgn` and I don't like the templated `DefaultRewriteConfig` or that the `rewrite` method would return a pair. I hope there is another way to achieve what is needed.

src/bin/Interpret.cc
```diff
@@ -709,18 +709,64 @@ void Interpret::getModel() {
...
+ *   )
+ * )
+ */
+std::string Interpret::printDefinitionSmtlib(PTRef tr, PTRef val) {
+    std::stringstream ss;
+    const char *s = logic->getSymName(tr);
+    SRef sortRef = logic->getSym(tr).rsort();
+    ss << "  (define-fun " << s << " () " << logic->getSortName(sortRef) << '\n';
+    char* val_string = logic->pp(val);
+    std::string val_s(val_string);
```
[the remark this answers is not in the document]
aehyvari: Looks like so.

src/logics/Logic.cc
```diff
@@ -501,6 +515,9 @@ bool Logic::declare_sort_hook(SRef sr) {
     ites.insert(tr, true);
     sortToIte.insert(sr, tr);
 
+    std::stringstream ss;
+    ss << Logic::s_abstract_value_prefix << "d" << sort_store.numSorts();
```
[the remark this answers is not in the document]
aehyvari: Ok.

src/logics/Logic.cc
```diff
@@ -283,7 +283,17 @@ Logic::protectName(const char* name) const
...
     char *name_escaped;
     int printed_chars = hasQuotableChars(name) ? asprintf(&name_escaped, "|%s|", name)
-            : asprintf(&name_escaped, "%s", name);
+                                               : asprintf(&name_escaped, "%s", name);
+    (void)printed_chars;
+    return name_escaped;
+}
+
+char*
+Logic::protectName(const std::string& name) const
```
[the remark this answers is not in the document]
aehyvari: Yes.

src/logics/Logic.h
aehyvari: Yes.

src/minisat/mtl/Map.h
aehyvari: *in

src/models/Model.cc
aehyvari: I'll just remove the prints.

src/logics/Logic.h
aehyvari: I'm not sure.  In the end it probably matters little.

regression/model/moreNonUFsIntroducedForEgraph.smt2
```diff
@@ -0,0 +1,1001 @@
...
+(assert (= (= x453 x578) x85))
+(assert (= x286 (and x134 x448)))
+(assert (= x238 (not x206)))
+(assert (= (and x1 x511) x512))
+(assert (= (x123 x408) x341))
+(assert (= x609 (ite x186 x410 x578)))
+(assert (= x574 (= x611 x550)))
+(check-sat)
+(get-model)
+(exit)
```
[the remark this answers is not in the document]
aehyvari: It could actually be ok to shorten this one.  There's another regression test now that's been more useful at catching problems.

aehyvari:
I rebased this on current master, rebased the couple of commits from @blishko's models branch on top of this and addressed some of the comments.  I'm running some bigger scale tests to see if things are starting to work...

blishko (commented):
Small batch of comments, I still have not checked how the `Egraph` computes the model and how this information gets to `Model`.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -213,23 +215,131 @@ void Egraph::getConflict( bool deduction, vec<PtAsgn>& cnfl )
...
+            seen_substs.insert(tr, true);
+            enode_store.peekERef(target_tr, target_er);
+            if (target_er == ERef_Undef) {
+                // Substitution removed target_tr from input instance, it was never introduced to the SAT solver, and
+                // therefore also theory solver is not aware of it.
+                // The original term is guaranteed to be in the solver and is authoritative in this case.
+                target_er = er;
+            }
+        } else {
+            assert(([&](PTRef tr) { ERef tmp; enode_store.peekERef(tr, tmp); return tmp == er; })(tr));
```
blishko, line 281: This is a nice trick :)
But I am not sure what it says. Should `tr` be in a the store or not? I think this can be expressed more clearly.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -213,23 +215,131 @@ void Egraph::getConflict( bool deduction, vec<PtAsgn>& cnfl )
...
+            assert(enode_store.peekERef(tr, target_er));
+            assert(target_er == er);
```
blishko, line 283: Actually, don't these two asserts say the same thing as the one with lambda?

src/logics/Logic.cc
```diff
@@ -1255,28 +1264,31 @@ lbool Logic::retrieveSubstitutions(const vec<PtAsgn>& facts, MapWithKeys<PTRef,P
...
     std::vector<char> notChangedElems(substs.getSize(), 0); // True if not changed in last iteration, initially False
     while (changed) {
         changed = false;
         for (int i = 0; i < keys.size(); ++i) {
-            auto & val = substs[keys[i]];
-            if (val.sgn != l_True || notChangedElems[i]) { continue; }
-            PTRef oldVal = val.tr;
-            PTRef newVal = Substitutor(*this, substs).rewrite(oldVal);
+            PTRef & val = getPTRefFromAsgn<TAsgn>(substs[keys[i]]);
+            if (notChangedElems[i]) { continue; }
```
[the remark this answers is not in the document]
aehyvari: Isn't this a bug?  I'd say that it's not correct to rewrite the disabled substitutions.

src/logics/Logic.cc
```diff
@@ -1285,6 +1297,9 @@ void Logic::substitutionsTransitiveClosure(MapWithKeys<PTRef, PtAsgn, PTRefHash>
     }
 }
 
+template void Logic::substitutionsTransitiveClosure<PtAsgn>(MapWithKeys<PTRef,PtAsgn,PTRefHash> &);
+template void Logic::substitutionsTransitiveClosure<PTRef>(MapWithKeys<PTRef,PTRef,PTRefHash> &);
+
```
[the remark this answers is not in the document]
aehyvari: The explicit template instantiations are needed here at least by clang on my mac.  I think it's got something to do with the way templates are created on one go, but I'm not sure if there's some differences how compilers do it.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: I think it's just a consistency check on `enode_store`, that it's map from `PTRef` -> `ERef` is consistent with the other direction, `ERef -> PTRef`.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: Yes, I agree.  The lambda looks a little nicer, at least it's everything in one place...

src/logics/Theory.cc
```diff
@@ -57,19 +56,25 @@ Theory::SubstitutionResult Theory::computeSubstitutions(const PTRef fla)
...
+
+        // remember the substitutions for models
+        for (PTRef key : newsubsts.getKeys()) {
             const auto target = newsubsts[key];
             if (!allsubsts.has(key) && target.sgn == l_True) {
-                allsubsts.insert(key, target);
+                allsubsts.insert(key, target.tr);
             }
         }
+        getLogic().substitutionsTransitiveClosure(allsubsts);
```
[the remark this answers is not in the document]
aehyvari: It sounds to me that once outside the loop should be enough.

PR #245, opened 2021-05-11
Enabling the option to simplify interpolants also in SMT-LIB scripts

blishko:
When OpenSMT is used as a library, the project has access to enable simplification of the interpolants immediately after they are computed through the field `SMTConfig::simplify_interpolant`.
However, this option cannot be set from SMT-LIB script using the standard `(set-option)` command.

This PR removes the field and instead uses the standard way to set and retrieve the value of this option.


blishko:
Not sure if this is the best time to introduce changes in `SMTConfig` but the ability to set this option can be useful.
At some point we should do some cleanup of `SMTConfig`.

aehyvari (approved):
This makes sense.  And also the cleaning at a future point.

PR #248, opened 2021-05-21
Remove getKeysAndVals and co from Map and implement MapWithKeys

aehyvari:
This PR removes the methods `Map::getKeysAndVals` and similar, which traverse the whole hash table, constructs a vector of (key,value)-pairs, and returns the vector.  The methods were never supposed to be used in non-debug code, and are now removed.  The replacement is a new class added to our minisat template library called `MapWithKeys`, which is a simple wrapper around `Map` but stores the keys in a vector.  For now I didn't implement removal of keys/values, since it's not needed anywhere in opensmt. 

The PR introduces some changes in the search order since at least simplification code traverses the values of the maps in different order now.  The good news would be that after this change the order is a little more deterministic since rehashing doesn't affect the order.

In general the effect seems to be positive (see the attached plots).  Surprisingly there's little difference in QF_UF.  Both QF_LRA and QF_LIA vary more, and especially QF_LIA results show that we have some work to do in making the search algorithm more robust.

In general it makes sense that the average speedup (sp in plots) is good and the total speedup (sp tot in plots) changes less, since this fix should mostly affect the preprocessing which is more significant in easy instances.

#### QF_UF
non-incremental-QF_UF-master-2021-05-18_vs_no-getkeysandvalues-2021-05-19

#### QF_LRA
non-incremental-QF_LRA-master-2021-05-19_vs_no-getkeysandvalues-2021-05-19

#### QF_LIA
non-incremental-QF_LIA-master-2021-05-20_vs_no-getkeysandvalues-2021-05-19


src/logics/SubstLoopBreaker.cc
```diff
@@ -210,29 +205,29 @@ std::vector<vec<SNRef>> SubstLoopBreaker::findLoops(vec<SNRef>& startNodes) {
...
 
-Map<PTRef,PtAsgn,PTRefHash> SubstLoopBreaker::constructLooplessSubstitution(Map<PTRef,PtAsgn,PTRefHash>&& substs)
+MapWithKeys<PTRef,PtAsgn,PTRefHash> SubstLoopBreaker::constructLooplessSubstitution(MapWithKeys<PTRef,PtAsgn,PTRefHash>&& substs)
 {
-    auto keysAndValsPtrs = substs.getKeysAndValsPtrs();
-    for (int i = 0; i < keysAndValsPtrs.size(); i++) {
-        auto pair = keysAndValsPtrs[i];
-        if (pair->data.sgn != l_True)
+    for (PTRef key : substs.getKeys()) {
+        PtAsgn & data(substs[key]);
```
blishko, line 211:
```suggestion
        PtAsgn & data = substs[key];
```

src/logics/Theory.cc
```diff
@@ -88,7 +86,7 @@ Theory::SubstitutionResult Theory::computeSubstitutions(const PTRef fla)
     }
 #endif
     result.result = root;
-    allsubsts.moveTo(result.usedSubstitution);
+    std::swap(allsubsts, result.usedSubstitution);
```
blishko, line 89:
```suggestion
    result.usedSubstitutions = std::move(allsubsts);
```

src/logics/CUFLogic.h
```diff
@@ -29,15 +29,15 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+    MapWithKeys<PTRef,bool,PTRefHash> comm_eqs;         // a+b <-> b+a
+    MapWithKeys<PTRef,bool,PTRefHash> diseq_eqs;        // a>b -> a != b
+    MapWithKeys<PTRef,bool,PTRefHash> diseq_split;      // a != b -> (a>b) || (a<b)
+    MapWithKeys<PTRef,bool,PTRefHash> mod_ineqs;        // b > 0 -> 0 <= a % b < b, b < 0 -> b < a % b <= 0
+    MapWithKeys<PTRef,bool,PTRefHash> inc_diseqs;       // a++ != a (< is not safe for overflows for some compiler semantics)
+    MapWithKeys<PTRef,bool,PTRefHash> compl_diseqs;     // ~a != a
 
   public:
-    void getCommEqs(vec<PTRef>& out) const { comm_eqs.getKeys(out); }
+    const vec<PTRef> & getCommEqs() const { return comm_eqs.getKeys(); }
```
blishko, line 40: I would expect that this method is not needed anymore and I would remove it...

src/logics/Logic.cc
```diff
@@ -1263,14 +1263,14 @@ lbool Logic::retrieveSubstitutions(const vec<PtAsgn>& facts, Map<PTRef,PtAsgn,PT
     return l_Undef;
 }
 
-void Logic::substitutionsTransitiveClosure(Map<PTRef, PtAsgn, PTRefHash> & substs) {
+void Logic::substitutionsTransitiveClosure(MapWithKeys<PTRef, PtAsgn, PTRefHash> & substs) {
     bool changed = true;
-    auto keyValPairs = substs.getKeysAndValsPtrs(); // We can use direct pointers, since no elements are inserted or deleted in the loop
-    std::vector<char> notChangedElems(keyValPairs.size(), 0); // True if not changed in last iteration, initially False
+    auto & keys(substs.getKeys()); // We can use direct pointers, since no elements are inserted or deleted in the loop
```
blishko, line 1268:
```suggestion
    auto & keys = substs.getKeys(); // We can use direct pointers, since no elements are inserted or deleted in the loop
```

src/logics/Logic.cc
```diff
@@ -1263,14 +1263,14 @@ lbool Logic::retrieveSubstitutions(const vec<PtAsgn>& facts, Map<PTRef,PtAsgn,PT
...
+    std::vector<char> notChangedElems(substs.getSize(), 0); // True if not changed in last iteration, initially False
     while (changed) {
         changed = false;
-        for (int i = 0; i < keyValPairs.size(); ++i) {
-            auto & val = keyValPairs[i]->data;
+        for (int i = 0; i < keys.size(); ++i) {
+            auto & val = substs[keys[i]];
             if (val.sgn != l_True || notChangedElems[i]) { continue; }
             PTRef oldVal = val.tr;
             PTRef newVal = Substitutor(*this, substs).rewrite(oldVal);
```
blishko, line 1276: A side note for future: since `Substitutor` depends on Logic, I don't think it should be used inside Logic. I think this whole method `substitutionsTransitiveClosure` should be moved outside the `Logic` class and put somewhere among helper methods.

src/logics/Theory.cc
```diff
@@ -39,37 +39,35 @@ Theory::SubstitutionResult Theory::computeSubstitutions(const PTRef fla)
...
         PTRef new_root = Substitutor(getLogic(), newsubsts).rewrite(root);
         bool cont = new_root != root;
         // remember the substitutions
-        vec<Map<PTRef,PtAsgn,PTRefHash>::Pair> newsubsts_vec;
-        newsubsts.getKeysAndVals(newsubsts_vec);
-        for (int i = 0; i < newsubsts_vec.size(); ++i) {
-            PTRef key = newsubsts_vec[i].key;
-            if (!allsubsts.has(key) && newsubsts_vec[i].data.sgn == l_True) {
-                allsubsts.insert(key, newsubsts_vec[i].data);
+        auto & newsubsts_vec(newsubsts.getKeys());
```
blishko, line 66:
```suggestion
        const auto & newsubsts_vec = newsubsts.getKeys();
```

src/logics/Theory.cc
```diff
@@ -39,37 +39,35 @@ Theory::SubstitutionResult Theory::computeSubstitutions(const PTRef fla)
...
+        for (PTRef key : newsubsts_vec) {
+            const auto & el(newsubsts[key]);
```
blishko, line 68:
```suggestion
            const auto & el = newsubsts[key];
```
Since it is `PtAsgn`, which is a small object I would maybe even prefer not use a reference here.

src/logics/Theory.cc
```diff
@@ -104,17 +102,15 @@ Theory::printFramesAsQuery(const vec<PFRef> & frames, std::ostream & s) const
...
-
 PTRef Theory::flaFromSubstitutionResult(const Theory::SubstitutionResult & sr) {
     Logic & logic = getLogic();
     vec<PTRef> args;
-    auto entries = sr.usedSubstitution.getKeysAndValsPtrs();
-    for (auto * entry : entries) {
-        assert(entry->data.sgn == l_True);
-        if (entry->data.sgn == l_True) {
-            args.push(logic.mkEq(entry->key, entry->data.tr));
+    auto & entries = sr.usedSubstitution.getKeys();
```
blishko, line 108:
```suggestion
    auto const & entries = sr.usedSubstitution.getKeys();
```

src/minisat/mtl/Map.h
```diff
@@ -155,65 +155,6 @@ class Map {
...
-        if (size == 0) return {};
-        vec<const Pair*> out;
-        for (int i = 0; i < cap; i++) {
-            if (table[i] == NULL) continue;
-            for (int j = 0; j < table[i].size(); j++)
-                out.push(&table[i][j]);
-        }
-        return out;
-    }
-
```
blishko, line 216: As discussed, I would suggest to keep these methods in the original Map.

src/models/ModelBuilder.cc
```diff
@@ -4,24 +4,38 @@
...
+            // We store only values of reals variables
+            continue;
+        }
+        assert(logic.isVar(key));
+        PtAsgn target = copy[key];
+        if (target.sgn == l_True) {
+            PTRef mappedTerm = target.tr;
+            PTRef val = model->evaluate(mappedTerm);
+            assert(logic.isConstant(val));
+            assignCopy.insert(std::make_pair(key, val));
```
blishko, line 38: Looks like the same code twice. Copy-paste error?

src/rewriters/Substitutor.h
```diff
@@ -43,7 +43,7 @@ class Substitutor : public Rewriter<SubstitutionConfig> {
     SubstitutionConfig config;
 
 public:
-    Substitutor(Logic &logic, Map<PTRef, PtAsgn, PTRefHash> const &substs) :
+    Substitutor(Logic &logic, MapWithKeys<PTRef, PtAsgn, PTRefHash> const &substs) :
```
blishko, line 46:
```suggestion
    Substitutor(Logic &logic, SubstitutionConfig::SubMap const &substs) :
```

src/tsolvers/TSolverHandler.h
```diff
@@ -68,8 +68,9 @@ class TSolverHandler
     virtual const Logic& getLogic() const = 0;
     virtual PTRef getInterpolant(const ipartitions_t& mask, map<PTRef, icolor_t>*, PartitionManager& pmanager) = 0;
 
-    void    setSubstitutions(Map<PTRef,PtAsgn,PTRefHash>& substs_) { substs_.moveTo(substs); }
-    Map<PTRef,PtAsgn,PTRefHash> const & getSubstitutions() const { return substs; }
+    void    setSubstitutions(MapWithKeys<PTRef,PtAsgn,PTRefHash>&& substs_) { std::swap(substs_, substs); }
```
blishko, line 71:
```suggestion
    void    setSubstitutions(MapWithKeys<PTRef,PtAsgn,PTRefHash>&& substs_) { substs = std::move(substs_); }
```

src/minisat/mtl/MapWithKeys.h
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // PRECONDITION: the key must already exist in the map.
+    D& operator [] (const K& k)
+    {
+        return map.operator[](k);
+    }
+
+    // PRECONDITION: the key must *NOT* exist in the map.
+    void insert (const K& k, const D& d) { map.insert(k, d); keys.push(k); }
```
blishko, line 34: We could add `assert(not has(k))` here to enforce the precondition in Debug.

blishko (changes requested):
Looks good! There are a few small things to polish and one weird code duplication.

src/logics/CUFLogic.h
aehyvari: Makes sense.

src/logics/Logic.cc
aehyvari: I agree.  We should find it a better home in the future.

src/logics/Theory.cc
aehyvari: We can do that.

src/models/ModelBuilder.cc
aehyvari: Indeed, strange!

src/rewriters/Substitutor.h
aehyvari: Ok.

src/tsolvers/TSolverHandler.h
aehyvari: OK.

src/minisat/mtl/MapWithKeys.h
aehyvari: Ok. I'm going to put it in mtl::Map.

src/minisat/mtl/Map.h
aehyvari: These are brought back.

aehyvari: Reintroduced.

blishko (approved):
I think it's all good now!

PR #249, opened 2021-05-22
Fixed incorrect answer in UF

blishko:
When boolean term is an argument of UF, we have to not only to merge that term with True, but also its negation with False.
This was not happening if such term was itself an equality of uninterpreted sort.


aehyvari (approved):
This works, let's merge it!

PR #250, opened 2021-05-23
Rewrite ITEs with auxiliary variables

### LINKED ISSUE #247: Invalid model in the present of ITEs
On the following example OpenSMT currently returns invalid model:
```
(set-logic QF_LRA)
(declare-fun v1 () Real)
(declare-fun v2 () Real)
(declare-fun v3 () Real)
(declare-fun v4 () Real)

(declare-fun b1 () Bool)
(declare-fun b2 () Bool)
(declare-fun b3 () Bool)
(declare-fun b4 () Bool)

(assert (= (+ v2 (ite b1 v2 v3)) v1))
(assert (= b2 (<= v4 v1)))
(assert (= b3 (not b2)))
(assert (= b4 (and b3 (not b1))))
(assert b4)
(assert (distinct v2 v3))
(check-sat)
(get-model)
```

blishko:
This PR proposes to rewrite ITE terms with auxiliary variables  when a formula is inserted to the solver (after ITE optimization).
The benefit is that after the ITE optimization step, definitions of ITE values are introduced and ITE terms should be treated as variables from this point on. This complicates the rest of the code where this needs to be considered all the time.
This complicates the rest of the code and introduces possibility of subtle bugs (as witnessed by #247).
If we eliminate ITE terms at the beginning, we do not need to introduce special handling for ITEs anywhere else.

This PR only introduces the elimination step, cleanup of the rest of the code base (to remove special handling of ITEs) will follow.

Fixes #247.


blishko:
Hi @aehyvari , I have updated this PR after our discussion, but not entirely as we discussed.
Let's see if I can explain my reasoning.

First, I kept the substitution map (from ITEs to auxiliary variables) as a member of `IteHandler`. The main reason is that I see the objective of this class to rewrite *multiple formulas* and since we need a substitution map when rewriting them, I belive it is better to reuse a single map for all formulas rather than building a fresh map for each formula to be rewritten.
For the same purpose I kept `IteHandler` as a member of `MainSolver`.

The second thing is that I hid the second mode of operation (for interpolation) entirely insider the `IteHandler` class. Now the object of this class is configured at the construction to either guarantee that it will use unique auxiliary variables on each call to `rewrite`, or it will re-use the auxiliary variables for the same terms across rewriting of different formula.
In this way I can clearly describe two ways in which this class can operate and the usage is the same, the only difference is at the construction of the object.
To me the operation of the class is easier to describe in this way and the desired behaviour can be easily tested.

Are these arguments convincing for you or would you prefer the solution we discussed today?

aehyvari (approved):

blishko:
I have added a regression test where previously invalid model was computed.

To sum up: the agreed implementation is not the one described above, rather now we do not use any map to remember auxiliary variables for ITE terms. Instead we rely on Logic's mechanism to return the same PTRef when always asked for a variable with same name (and sort). When different auxiliary variables are needed for the same ITE (as in partitioned formula for interpolation) the `IteHandler` can be handled a unique number at the construction which it uses as a suffix for the variables name.

aehyvari (approved):

PR #251, opened 2021-05-27
Fix crash in UF interpolator

### LINKED ISSUE #231: Bug in the EUF interpolation
When running UpProver in EUF encoding for the attached example, an exception is thrown in `getSingleInterpolant(interpolants, mask)`. 

Up to OpenSMT version 555c592f the verification was successful, but it seems after d1592f50 onwards verification of the example crashes.

- Example: see [here](https://gist.github.com/sepideha/a803d4497288b54cfd538cfa2ce6f83a)

- Exact command line resulting in the issue: run in UpProver (4cf51de8 or any older version): `./upprover --logic qfuf --bootstrapping ex7-v5.c`

- What behaviour did you expect: UNSAT in EUF with generated interpolants

- What happened instead: exception thrown in `getSingleInterpolant(interpolants, mask)` with message `libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found`

P.S. I also tried OpenSMT ec7cf31c but the following exception is thrown in the same method `libc++abi.dylib: terminating with uncaught exception of type OsmtInternalException: No color detected for term`

blishko:
Because we re-use one theory solver for all theory conflicts, it may happen that the explanation of a conflict depends on a term not present in that particular conflict (but is present in the input formula).
This breaks the invariant assumed before, i.e., that local coloring information is enough to color the `CGraph` (congruence graph representing the explanation).
But even in global coloring information, a term (like `f(x)`) can originally be assigned only color `A` (when present only in partition `A`) while in the `CGraph` we need to color it as `AB` because both `f` and `x` have color `AB`.
This required a small change in the code which was not prepared for such situation.

Added are a regression and a unit test that show this problem, although they are sensitive to the order in which the terms are introduced.

Fixes #231.

### DIFF
--- a/src/tsolvers/egraph/UFInterpolator.cc
+++ b/src/tsolvers/egraph/UFInterpolator.cc
@@ -1227,7 +1227,7 @@ void UFInterpolator::splitEdge(CEdge * edge, PTRef intermediateTerm) {
     PTRef intermediate_next_reason = PTRef_Undef;
     if (cgraph.hasNode(intermediateTerm)) {
         intermediate = cgraph.getNode(intermediateTerm);
-        if (intermediate->next != nullptr) {
+        if (intermediate->next) {
             intermediate_next = intermediate->next->target;
             intermediate_next_reason = intermediate->next->reason;
             cgraph.removeCEdge(intermediate->next);
--- a/src/tsolvers/egraph/UFInterpolator.cc
+++ b/src/tsolvers/egraph/UFInterpolator.cc
@@ -521,7 +521,7 @@ UFInterpolator::J(const path_t & p, vector<path_t> & b_paths) {
 
     vec<PTRef> conj;
 
-    for (auto path : b_paths) {
+    for (const auto & path : b_paths) {
         conj.push(logic.mkEq(path.first->e, path.second->e));
     }
 
@@ -555,13 +555,13 @@ UFInterpolator::Iprime(const path_t & pi) {
         conj.push(I(theta));
     }
 
-    for (auto path : b_paths)
+    for (const auto & path : b_paths)
         conj.push(I(path));
 
     // Finally compute implication
     vec<PTRef> conj_impl;
 
-    for (auto path : b_paths)
+    for (const auto & path : b_paths)
         conj_impl.push(logic.mkEq(path.first->e, path.second->e));
 
     PTRef implicant = logic.mkAnd(conj_impl);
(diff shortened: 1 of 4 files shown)


src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -342,16 +295,32 @@ bool UFInterpolator::colorEdgesFrom(CNode * x) {
...
 
+void UFInterpolator::colorCongruenceEdge(CEdge * edge) {
+    assert(edge);
+    CNode * from = edge->source;
+    CNode * to = edge->target;
+    if (from->color == to->color) {
+        assert (from->color);
+        edge->color = from->color == I_AB ? resolveABColor() : from->color;
+    } else { // Different colors: choose intersection
+        // It is not possible that are incompatible
```
aehyvari, line 306: the *colors* are incompatible?

blishko: Updated.

src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -1264,3 +1233,50 @@ bool UFInterpolator::checkColors() {
...
+    cgraph.addCEdge(from, intermediate, PTRef_Undef);
+    assert(from->next != nullptr); // the added edge is from->next
+    assert(from->next->target == intermediate);
+    // Choose a color; we know that intermediate is AB, so we color edge based on from
+    assert (from->color == I_A || from->color == I_B || from->color == I_AB);
+    from->next->color = from->color == I_AB ? resolveABColor() : from->color;
+
+    cgraph.addCEdge(intermediate, to, PTRef_Undef);
+//    intermediate->next->color = to->color;
+    if (intermediate_next != nullptr) {
```
aehyvari, line 1272:
```suggestion
    if (intermediate_next) {
```

src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -1264,3 +1233,50 @@ bool UFInterpolator::checkColors() {
...
+    CNode * from = edge->source;
+    CNode * to = edge->target;
+    assert (from->next->target == to);
+    // Adds corresponding node
+    CNode * intermediate = nullptr;
+    CNode * intermediate_next = nullptr;
+    PTRef intermediate_next_reason = PTRef_Undef;
+    if (cgraph.hasNode(intermediateTerm)) {
+        intermediate = cgraph.getNode(intermediateTerm);
+        if (intermediate->next != nullptr) {
```
aehyvari, line 1248:
```suggestion
        if (intermediate->next) {
```

aehyvari (approved):
Looks good to me.  I left two small suggestions.  While going through the PR I was cleaning some code you didn't touch.  Should I commit them to this PR or make a separate one?  They're minor but in my opinion improve readability.

src/tsolvers/egraph/UFInterpolator.cc
blishko: Looks good!

blishko:
Let me rebase this, then you can commit your changes and then we can merge.

aehyvari:
OK I committed my cleanup.

PR #253, opened 2021-06-01
Function templates

aehyvari:


src/models/FunctionTools.h
```diff
@@ -0,0 +1,59 @@
...
+        signature.ret_sort = ret_sort;
+    }
+    TemplateFunction(FunctionSignature&& signature, PTRef body) : signature(std::move(signature)), tr_body(body) {}
+    TemplateFunction() : tr_body(PTRef_Undef) {}
+    TemplateFunction(const TemplateFunction& other) : tr_body(other.tr_body) {
+        signature.ret_sort = other.getRetSort();
+        signature.name = other.getName();
+        other.getArgs().copyTo(signature.args);
+    }
+    TemplateFunction& operator= (TemplateFunction&& other) noexcept {
```
blishko, line 45: If it needs move assignment operator, should we add also move constructor?
Wouldn't actually the default move-assignment operator work here? It would require the move-assignment on FunctionSignature but that could also be default, no?

blishko (commented):
Looks good, I have just a few small comments.

src/logics/Logic.cc
```diff
@@ -1634,27 +1633,26 @@ Logic::dumpFunction(ostream& dump_out, const TFun& tpl_fun)
...
-            return PTRef_Undef;
-        PtAsgn subst_target = {subst_target_tr, l_True};
-        substs_asgn.insert(args[i], subst_target);
+        if (!subst.has(args[i])) {
+            std::string argName = pp(args[i]);
+            throw OsmtApiException(
+                    "No value provided for function " + tpl_fun.getName() + " argument " + argName);
+        }
+        if (getSortRef(subst[args[i]]) != getSortRef(args[i])) {
+            throw OsmtApiException("Substitution source and target and sort mismatch" );
```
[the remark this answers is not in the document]
aehyvari: Removed second and.

src/logics/Logic.h
```diff
@@ -99,33 +55,27 @@ class Logic {
...
 
     class DefinedFunctions {
-        std::map<std::string,TFun> defined_functions;
-        vec<char*> defined_functions_names;
+        std::unordered_map<std::string,TemplateFunction> defined_functions;
+        std::vector<std::string> defined_functions_names;
 
     public:
         bool has(const char* name) const { return defined_functions.find(name) != defined_functions.end(); }
+        bool has(const std::string& name) const { return has(name.c_str()); }
```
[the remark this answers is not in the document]
aehyvari: I'll remove `Logic::DefinedFunctioons::has(const char*) const`.

src/models/FunctionTools.h
aehyvari: I think that the default doesn't work for FunctionSignature because it has `mtl::vec` as a field.

The move assignment (or move constructor) is used from `Logic::defineFun`.  I'll add the mode constructor.

blishko: But `mtl::vec` has both move constructor and move-assignment operator, so I still think it should work, no?

PR #255, opened 2021-06-02
Reducing memory allocations - part 1

blishko:
[Heaptrack](https://github.com/KDE/heaptrack) identified some parts of OpenSMT making a lot of memory allocations. This PR is the first one in a series attempting to reduce the number of memory allocations.
`Polynomial::merge` is the key method that keeps the Tableau of LASolver in a consistent state and is called quite a number of times.
It uses additional space (storage) to properly merge two polynomials into a new one. However, this additional storage was not re-used as much as it could.
This PR proposes to save this temporary storage in the Tableau, so it can be does not have to be allocated and destroyed repeatedly.
Moreover, the method `Tableau::normalizeRow` was not re-using this storage previously at all, even though  it could.
Both `normalizeRow` and `pivot` now reuse the storage associated with the `Tableau`.

In addition to re-using the vector itself, we also try to re-use the `FastRationals` stored in the vector, so we don't clear and re-fill the storage every time.

### DIFF
--- a/src/tsolvers/lasolver/Polynomial.h
+++ b/src/tsolvers/lasolver/Polynomial.h
@@ -81,7 +81,10 @@ public:
 template<typename ADD, typename REM>
 void Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, ADD informAdded,
                   REM informRemoved, std::vector<Term>& storage) {
-    storage.reserve(this->poly.size() + other.poly.size());
+    if (storage.size() < this->poly.size() + other.poly.size()) {
+        storage.resize(this->poly.size() + other.poly.size(), Term(LVRef_Undef, 0));
+    }
+    std::size_t storageIndex = 0;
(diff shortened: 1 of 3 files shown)


blishko:
[Here](https://github.com/usi-verification-and-security/opensmt/files/6587278/comparison_lra_polyreuse.pdf) is the effect on LRA benchmarks. Overall, this shows ~4% speedup.

aehyvari (approved):
This looks good!  My only question, not even related to this PR, is whether we should remove `Polynomial::merge` (on line 38 of Polynomial.h) so that nobody uses it by mistake.  I think that it's only used in regression tests (but I didn't check for uses in assert).

blishko:
That's a good question, I see points for both keeping and removing.
It is true that we should re-use storage if possible, especially in performance critical code.
On the other hand, it might be a handy shortcut in non-critical code.

Now it seems to be used only in tests. So I think we could remove it if you believe that's better.

PR #256, opened 2021-06-08
Reduction of memory allocations - part2

blishko:
This PR represents part 2 in reducing memory allocations, it builds on top of #255.
It itself has two parts.
One is just avoiding unnecessary copies by using `const &` in a few places where this was missing.
The other one is more important and it tries to avoid allocating a temporary vector of `Term`s in `Polynomial::merge`.
An important note here is that we need to be careful not to end up with lots of polynomials in the `Tableau` with small size but huge capacity. This would blow up in terms of memory usage.
So sometimes we need to shrink them to release unused capacity. The goal of this change is to do that only when necessary.
The experiments show the current version seems to be working well.

[comparison_lra_mem1_mem2.pdf](https://github.com/usi-verification-and-security/opensmt/files/6614559/comparison_lra_mem1_mem2.pdf)

### DIFF
--- a/src/tsolvers/lasolver/Polynomial.h
+++ b/src/tsolvers/lasolver/Polynomial.h
@@ -133,15 +133,29 @@ void Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, ADD 
             ++otherIt;
         }
     }
+    // At this point the right elements are in `storage`, from beginning to the `storageIndex`.
+    // We need to get these elements to `this->poly`. Note that we never change the `storage` container, we just move out
+    // the appropriate elements.
+    // However, we observed that we need to shrink `poly` if its size is much smaller than the capacity.
+    // The reason is that keeping large free capacity around for many rows blows up the memory
+    // (worse case quadratic in the size of the tableau).
     auto polySize = poly.size();
     if (storageIndex > polySize) {
+        // In this case we have more elements to move than the current size of the `poly` container.
+        // We move the elements that fit the current `poly` size and then we insert the rest of the elements.
         std::move(storage.begin(), storage.begin() + polySize, poly.begin());
         poly.insert(poly.end(), std::make_move_iterator(storage.begin() + polySize), std::make_move_iterator(storage.begin() + storageIndex));
     }
     else if (/*storageIndex <= poly.size() and */ polySize <= 2 * storageIndex) {
+        // In this case we have less elements that need to move than what we currently already have in `poly`, but not too litle.
+        // We just move the appropriate elements and destroy the excess elements of `poly`.
+        // Since we are removing too many elements, we avoid shrinking which would require re-allocation.
         std::move(storage.begin(), storage.begin() + storageIndex, poly.begin());
         poly.erase(poly.begin() + storageIndex, poly.end());
-    } else {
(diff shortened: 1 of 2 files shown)


src/tsolvers/lasolver/Polynomial.h
```diff
@@ -133,7 +133,17 @@ void Polynomial::merge(const Polynomial &other, const opensmt::Real &coeff, ADD
...
+    if (storageIndex > polySize) {
+        std::move(storage.begin(), storage.begin() + polySize, poly.begin());
+        poly.insert(poly.end(), std::make_move_iterator(storage.begin() + polySize), std::make_move_iterator(storage.begin() + storageIndex));
+    }
+    else if (/*storageIndex <= poly.size() and */ polySize <= 2 * storageIndex) {
+        std::move(storage.begin(), storage.begin() + storageIndex, poly.begin());
+        poly.erase(poly.begin() + storageIndex, poly.end());
+    } else {
+        std::vector<Term>(std::make_move_iterator(storage.begin()), std::make_move_iterator(storage.begin() + storageIndex)).swap(poly);
+    }
```
aehyvari, line 146: Similarly, on line 82, should we also allocate a bigger chunk, say `2*(this->poly.size() + other.poly.size())`?  This stuff is a little mind-twisting...

aehyvari (commented):
[Edited after discussions]
Code looks good!  I left a few comments, but many of them are not in the area this PR changes so they have weird references to line numbers...

The more interesting part of course is the storage size management and how it transfers to the polynomial sizes.

The memory management for polynomials works as follows:
 1. `storageIndex <= poly.size() <= 2*storageIndex`: keep the current capacity for the polynomial vector
 2. `poly.size() > 2* storageIndex`: reallocate a smaller polynomial vector
 3. `storageIndex > poly.size()`: Resize the polynomial vector, possibly reallocating, to fit the new polynomial
 
Could you document some intuition for these rules?

src/tsolvers/lasolver/Polynomial.h
aehyvari: To close this, the storage size should simply grow to at most twice the number of columns and then stay there.  It doesn't sound it's worth the effort to do anything explicitly for this and therefore the implementation in the PR is ok.

blishko:
I have added comments documenting the last phase of `Polynomial::merge`. Hopefully they explain the reasoning for the different cases.

src/tsolvers/lasolver/Simplex.cc
```diff
@@ -206,7 +206,7 @@ LVRef Simplex::findNonBasicForPivotByBland(LVRef basicVar) {
     else if (isModelOutOfUpperBound(basicVar)) {
         auto curr_var_id_y = max_var_id;
         // look for nonbasic terms to fix the unbounding
-        for (auto term : tableau.getRowPoly(basicVar)) {
+        for (auto const & term : tableau.getRowPoly(basicVar)) {
             auto y = term.var;
             assert(basicVar != y);
             assert(tableau.isNonBasic(y));
```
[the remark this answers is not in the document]
blishko: Fixed.

aehyvari (approved):
This looks good now.

PR #258, opened 2021-06-28
Seed global random generator with seed from SMTConfig to ensure reproducible runs.

blishko:
A number of methods in proof reduction uses  C's rand() method to randomly decide a choice when all heuristics fail.
This PR seeds the random number generator at the beginning of the reduction algorithm.
It's a workaround that could be improved later (for example by fixing the default choice and removing all the calls to rand()).


aehyvari (approved):
I think that this is a reasonable fix for now.  I'll create an issue so we remember to either remove or replace with MiniSat's rng the system calls.

PR #261, opened 2021-07-07
Proof transformations cleanup - part 1 of N

blishko:
This PR is first in the series of cleaning up proof transformation code.
The goal is to pinpoint and fix an error that I am seeing in CHC benchmarks, where after a reduction iteration a node ends up with both antecedents being the same node. This is something that should not be happening.

This first part contains style changes in two methods: `ProofGraph::mergeClauses` and `ProofGraph::recycleUnits`.

### DIFF
--- a/src/proof/PGHelp.cc
+++ b/src/proof/PGHelp.cc
@@ -198,70 +198,88 @@ void ProofGraph::topolSortingBotUp(vector<clauseid_t>& DFS)
 }
 
 // Linear merge for resolution
-bool ProofGraph::mergeClauses(vector<Lit>& A, vector<Lit>& B, vector<Lit>& resolv, Var pivot)
+bool ProofGraph::mergeClauses(std::vector<Lit> const & A, std::vector<Lit> const & B, std::vector<Lit>& resolv, Var pivot)
 {
-    size_t i, j;
-    i = 0;
-    j = 0;
-    const size_t Asize= A.size();
-    const size_t Bsize= B.size();
-    size_t ressize=0;
-    if(resolv.size() < Asize + Bsize - 2)
-    {
-        resolv.reserve(Asize + Bsize - 2);
+    const std::size_t Asize = A.size();
+    const std::size_t Bsize = B.size();
+    std::size_t ressize = 0;
+    if (resolv.size() < Asize + Bsize - 2) {
         resolv.resize(Asize + Bsize - 2);
     }
     assert(resolv.size() >= Asize + Bsize - 2);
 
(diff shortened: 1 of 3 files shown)


blishko:
I recommend to use the option to hide whitespace changes in the diff-view on GitHub, otherwise it's impossible to properly see the changes.

src/proof/PGHelp.cc
```diff
@@ -198,70 +198,86 @@ void ProofGraph::topolSortingBotUp(vector<clauseid_t>& DFS)
...
+                ressize++;
+            }
             j++;
         }
     }
-    if (i < Asize)
-    {
+    if (i < Asize) {
         for (size_t p = i; p < Asize; p++)
-            if(var(A[p])!=pivot && A[p]!=resolv[ressize-1]){ resolv[ressize]=A[p]; ressize++; }
```
aehyvari, line 251: Brackets here?

src/proof/PGHelp.cc
```diff
@@ -198,70 +198,86 @@ void ProofGraph::topolSortingBotUp(vector<clauseid_t>& DFS)
...
+                resolv[ressize] = B[j];
+                ressize++;
+            }
             j++;
         }
     }
-    if (i < Asize)
-    {
+    if (i < Asize) {
         for (size_t p = i; p < Asize; p++)
```
aehyvari, line 261: Brackets here?

src/proof/PGHelp.cc
```diff
@@ -198,70 +198,86 @@ void ProofGraph::topolSortingBotUp(vector<clauseid_t>& DFS)
...
-            if(var(A[p])!=pivot && A[p]!=resolv[ressize-1]){ resolv[ressize]=A[p]; ressize++; }
-    }
-    else if(j < Bsize)
-    {
+            if (var(A[p]) != pivot && A[p] != resolv[ressize - 1]) {
+                resolv[ressize] = A[p];
+                ressize++;
+            }
+    } else if (j < Bsize) {
         for (size_t p = j; p < Bsize; p++)
```
aehyvari, line 267: Brackets here?

aehyvari (approved):
Apart from some assertions this is just about reformatting the source code to conform to our style.

src/proof/PGHelp.cc
blishko: Fixed.

src/proof/PGHelp.cc
blishko: Fixed.

PR #262, opened 2021-07-08
Proof transformations cleanup - part 2 of N: Refactor ProofGraph::mergeClauses

blishko:
The implementation of the method has been simplified and assertions for preconditions and postconditions have been added.
Several unit tests have been added to check correctness of the implementation.

### DIFF
--- a/src/proof/PGHelp.cc
+++ b/src/proof/PGHelp.cc
@@ -197,12 +197,20 @@ void ProofGraph::topolSortingBotUp(vector<clauseid_t>& DFS)
     delete visited_count_;
 }
 
-// Linear merge for resolution
+/*
+ * Given two clauses A and B, and a pivot variable, computes the resolvent clause after resolution.
+ *
+ * PRECONDITION: Literals in the input clauses must be sorted and the clauses must contain the pivot variable.
+ * POSTCONDITION: Literals in the resolvent clause are sorted and the clause does not contain the pivot.
+ */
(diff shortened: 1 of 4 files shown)


aehyvari:
I added a suggestion that would avoid a bit the duplication of the code.  Do you think it'd be better this way or do you prefer the old version?

src/proof/PGHelp.cc
```diff
@@ -220,37 +220,35 @@ bool ProofGraph::mergeClauses(std::vector<Lit> const & A, std::vector<Lit> const
...
-            if (var(A[i]) != pivot) {
-                assert(res == 0 or resolv[res - 1] != A[i]); // no duplicates
-                resolv[res++] = A[i];
-            }
+        if (A[i] == B[j]) {
+            addIfNotPivot(A[i]);
+            ++i;
+            ++j;
+        } else if (A[i] < B[j]) {
+            addIfNotPivot(A[i]);
```
blishko, line 236: I like the lambda, but here I prefer my version of the `if` structure, so we do not need to call the lambda in both branches.
```
if (A[i] <= B[j] {
	if (A[i] == B[j] { ++j };
	addIfNotPivot(A[i]);
	++i;
}
```

aehyvari: Ok, let's go with that then if you prefer!

aehyvari (approved):

blishko:
Looks like I need to rebase. I am curious what will happen to the other PRs. In theory, after merging this one, the next one should be automatically redirected to merge to master instead of this one. Let's see if it works that way.

PR #263, opened 2021-07-08
Proof transformations cleanup - part 3 of N: ProofGraph::recyclePivotsIter

blishko:
This PR contains style fixes for method `ProofGraph::recyclePivotsIter`.
In addition, two separate phases of this method has been identified:
In the first phase, the proof is traversed from bottom to top and redundant nodes are identified. The redundant edge is removed.
In the second phase, the proof is traversed from top to bottom and redundant nodes identified in the previous phase are replaced with their (now single) parent. This can expose more redundant nodes as now some resolutions might not be necessary anymore.

The first phase has been extracted to standalone method `ProofGraph::recyclePivotsIter_RecyclePhase`. The second phase is still kept in the original method unchanged (except the style fixes).

Depends on #262.

### DIFF
--- a/src/proof/PGTransformationAlgorithms.cc
+++ b/src/proof/PGTransformationAlgorithms.cc
@@ -160,7 +160,7 @@ double ProofGraph::recyclePivotsIter() {
             mem_used == 0 ? 0 : mem_used / 1048576.0);
     }
 
-    auto hasBeenReplaced = [](ProofNode * node) { return node->getAnt1() != nullptr and node->getAnt2() == nullptr; };
+    auto hasBeenReplaced = [](ProofNode * node) { return node->getAnt1() and not node->getAnt2(); };
     // Restructuring, from leaves to root
     assert(isResetVisited2());
     std::deque<clauseid_t> q;
@@ -170,14 +170,14 @@ double ProofGraph::recyclePivotsIter() {
         assert(id < getGraphSize());
         ProofNode * n = getNode(id);
         q.pop_front();
-        if (n == nullptr or isSetVisited2(id)) { continue; }
+        if (not n or isSetVisited2(id)) { continue; }
         if (n->getAnt1() and not isSetVisited2(n->getAnt1()->getId())) { continue; }
         if (n->getAnt2() and not isSetVisited2(n->getAnt2()->getId())) { continue; }
         // Mark node as visited if both antecedents visited
         setVisited2(id);
         //Non leaf node
         if (n->getAnt1() or n->getAnt2()) {
-            assert(not(n->getAnt1() == nullptr and n->getAnt2() != nullptr));
+            assert(n->getAnt1() or not n->getAnt2());
             // If replaced assign children to replacing node and remove
             if (hasBeenReplaced(n)) {
                 ProofNode * replacing = n->getAnt1();
@@ -192,7 +192,7 @@ double ProofGraph::recyclePivotsIter() {
                     else opensmt_error_();
                     assert(res->getAnt1() != res->getAnt2());
                     replacing->addRes(resId);
-                    assert(!isSetVisited2(resId));
+                    assert(not isSetVisited2(resId));
                     q.push_back(resId);
                 }
                 replacing->remRes(id);
@@ -204,9 +204,9 @@ double ProofGraph::recyclePivotsIter() {
             } else { // node stays
                 assert(n->getAnt1() and n->getAnt2());
                 assert((n->getAnt1()->getAnt1() and n->getAnt1()->getAnt2())
-                       or (n->getAnt1()->getAnt1() == nullptr and n->getAnt1()->getAnt2() == nullptr));
+                       or (not n->getAnt1()->getAnt1() and not n->getAnt1()->getAnt2()));
                 assert((n->getAnt2()->getAnt1() and n->getAnt2()->getAnt2())
-                       or (n->getAnt2()->getAnt1() == nullptr and n->getAnt2()->getAnt2() == nullptr));
+                       or (not n->getAnt2()->getAnt1() and not n->getAnt2()->getAnt2()));
 
                 // NOTE not clear how this might happen
                 if (n->getAnt1() == n->getAnt2()) {
@@ -279,7 +279,7 @@ double ProofGraph::recyclePivotsIter() {
                             else if (res->getAnt2() == n) { res->setAnt2(replacing); }
                             else opensmt_error_();
                             replacing->addRes(clauseid);
-                            assert(!isSetVisited2(clauseid));
+                            assert(not isSetVisited2(clauseid));
                             // Enqueue resolvent
                             q.push_back(clauseid);
                         }
--- a/src/proof/PGTransformationAlgorithms.cc
+++ b/src/proof/PGTransformationAlgorithms.cc
@@ -40,7 +40,7 @@ void ProofGraph::recyclePivotsIter_RecyclePhase() {
     mpz_init(incr_safe_lit_set_2);
 
     const size_t size = getGraphSize();
-    mpz_t * safe_lit_set = new mpz_t[size];
+    auto * safe_lit_set = new mpz_t[size];
 
     // Allocate root bitset
     mpz_init(safe_lit_set[getRoot()->getId()]);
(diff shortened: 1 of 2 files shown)


blishko:
Base has been automatically changed to master, which is good, but I still need to rebase this.

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,325 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+    // present in all paths in its subgraph down to the root
+
+    // Increased sets for taking into account pivots
+    mpz_t incr_safe_lit_set_1;
+    mpz_init(incr_safe_lit_set_1);
+    mpz_t incr_safe_lit_set_2;
+    mpz_init(incr_safe_lit_set_2);
+
+    const size_t size = getGraphSize();
+    mpz_t * safe_lit_set = new mpz_t[size];
```
aehyvari, line 43: Use `auto`?

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,325 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+    // Allocate root bitset
+    mpz_init(safe_lit_set[getRoot()->getId()]);
+
+    //DFS vector
+    vector<clauseid_t> DFSvec;
+    topolSortingBotUp(DFSvec);
+
+    assert(isResetVisited1());
+    // To initialize pivots set to the set of the first resolvent
+    for (std::size_t i = 0; i < DFSvec.size(); ++i) {
```
aehyvari, line 54: We could use a range-based iterator.

aehyvari (commented):
I'm still working on this one, but I noticed that while the code you covered conforms to our style, the rest of it doesn't.  Would it make sense to make a pass on the rest of it as well?  Or is this in some later PR?

blishko:
I just covered the methods I have been examining to find the cause of my problem. I agree that it would be good to make a pass on the rest of the proof-related code. But it's not a small task :) I only covered one more method after this one, it's in part 5.
Do you want to do the big pass or are you suggesting I should do it? :)

aehyvari:
> Do you want to do the big pass or are you suggesting I should do it? :)

I was thinking of taking the task, without knowing how big it actually is :)

blishko:
> I was thinking of taking the task, without knowing how big it actually is :)

Have a look at the code and let me know how you decide :)
I would suggest to do it afterwards, as part 6 of the cleanup.

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
-				{
-					mpz_setbit( incr_safe_lit_set_1, pos_piv);
-					mpz_setbit( incr_safe_lit_set_2, neg_piv);
-				}
-				else
-				{
-					mpz_setbit( incr_safe_lit_set_2, pos_piv);
-					mpz_setbit( incr_safe_lit_set_1, neg_piv);
-				}
+    auto hasBeenReplaced = [](ProofNode * node) { return node->getAnt1() != nullptr and node->getAnt2() == nullptr; };
```
aehyvari, line 162: Not sure if this would be clearer...
```suggestion
    auto hasBeenReplaced = [](ProofNode * node) { return node->getAnt1() and not node->getAnt2(); };
```

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+        ProofNode * n = getNode(id);
+        q.pop_front();
+        if (n == nullptr or isSetVisited2(id)) { continue; }
+        if (n->getAnt1() and not isSetVisited2(n->getAnt1()->getId())) { continue; }
+        if (n->getAnt2() and not isSetVisited2(n->getAnt2()->getId())) { continue; }
+        // Mark node as visited if both antecedents visited
+        setVisited2(id);
+        //Non leaf node
+        if (n->getAnt1() or n->getAnt2()) {
+            assert(not(n->getAnt1() == nullptr and n->getAnt2() != nullptr));
```
aehyvari, line 179:
```suggestion
            assert(n->getAnt1() or not n->getAnt2());
```

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+                replacing->remRes(id);
+                assert(n == getNode(id));
+                removeNode(id);
+                assert(replacing->getNumResolvents() > 0);
+                // NOTE extra check
+                if (proofCheck() > 1) { checkClause(replacing->getId()); }
+            } else { // node stays
+                assert(n->getAnt1() and n->getAnt2());
+                assert((n->getAnt1()->getAnt1() and n->getAnt1()->getAnt2())
+                       or (n->getAnt1()->getAnt1() == nullptr and n->getAnt1()->getAnt2() == nullptr));
```
aehyvari, line 206:
```suggestion
                       or (not n->getAnt1()->getAnt1() and not n->getAnt1()->getAnt2());
```

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+                assert((n->getAnt2()->getAnt1() and n->getAnt2()->getAnt2())
+                       or (n->getAnt2()->getAnt1() == nullptr and n->getAnt2()->getAnt2() == nullptr));
```
aehyvari, line 208:
```suggestion
                       or (not n->getAnt2()->getAnt1() and not n->getAnt2()->getAnt2()));
```

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
-		reportf( "; Memory used after recycling, before restructuring: %.3f MB\n",  mem_used == 0 ? 0 : mem_used / 1048576.0 );
-	}
+                        auto const & resolvents = n->getResolvents();
+                        for (clauseid_t clauseid : resolvents) {
+                            assert(clauseid < getGraphSize());
+                            ProofNode * res = getNode(clauseid);
+                            assert(res);
+                            if (res->getAnt1() == n) { res->setAnt1(replacing); }
+                            else if (res->getAnt2() == n) { res->setAnt2(replacing); }
+                            else opensmt_error_();
```
aehyvari, line 279: or `OsmtInternalException`?

aehyvari (commented):
I left some comments that maybe could be written here, nothing critical.

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -25,379 +25,324 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
...
+            int neg_piv = toInt(mkLit(piv, true));
+
+            short out = n->getAnt1()->hasOccurrenceBin(piv);
+            assert(out != -1);
+            // TODO guarantee positive occurrence pivot in ant1
+            bool ant1_has_pos_occ_piv = (out == 0);
+
+            assert(n->getAnt1());
+            assert(n->getAnt2());
+            // 0 if no replacement, 1 if replacement by ant1, 2 if by ant2
```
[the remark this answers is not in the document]
blishko: Good idea! I used enum class.

src/proof/PGTransformationAlgorithms.cc
blishko: I was not sure myself, but let's go with your suggestion.

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc
blishko: There is some shuffling  in the next PR that touches this code, so I think it's better to do this in the later PR.

aehyvari (approved):
I think that this is good to go.

PR #264, opened 2021-07-08
Proof transformations cleanup - part 4 of N

blishko:
This PR adds two unit tests for testing `ProofGraph::recyclePivotsIter`.
The second one shows that it is possible that during this algorithm a situation might occur where a node ends up with antecedents being one and the same node. This follows from the properties of the algorithm and the algorithm can handle it simply by replacing such node with its (single) parent.
At the end of the algorithm, the proof is always in a valid state and there will not be any node with identical parents.

Depends on #263.

### DIFF
--- a/src/proof/PGTransformationAlgorithms.cc
+++ b/src/proof/PGTransformationAlgorithms.cc
@@ -236,7 +236,7 @@ double ProofGraph::recyclePivotsIter() {
                         // Choose one of the two antecedents heuristically
                         choose_ant1 = chooseReplacingAntecedent(n);
                     } else {
-                        choose_ant1 = not piv_in_ant1 and piv_in_ant2;
+                        choose_ant1 = not piv_in_ant1;
                     }
 
                     ProofNode * replacing = choose_ant1 ? n->getAnt1() : n->getAnt2();
--- a/src/proof/PGTransformationAlgorithms.cc
+++ b/src/proof/PGTransformationAlgorithms.cc
@@ -20,6 +20,7 @@ along with Periplo. If not, see <http://www.gnu.org/licenses/>.
 
 #include "PG.h"
 #include "CoreSMTSolver.h" // TODO: MB: deal with reportf and remove this include
+#include "OsmtInternalException.h"
 
 #include <unordered_set>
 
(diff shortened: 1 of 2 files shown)


src/proof/PGTransformationAlgorithms.cc
```diff
@@ -189,7 +190,7 @@ double ProofGraph::recyclePivotsIter() {
                     assert(n == res->getAnt1() or n == res->getAnt2());
                     if (res->getAnt1() == n) { res->setAnt1(replacing); }
                     else if (res->getAnt2() == n) { res->setAnt2(replacing); }
-                    else opensmt_error_();
+                    else { throw OsmtInternalException("Invalid proof structure " + std::string(__FILE__) + ", " + std::to_string(__LINE__)); }
```
aehyvari, line 193: This is protected from happening by the assert in line 190.  Should we maybe remove the assert then?

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -208,96 +209,74 @@ double ProofGraph::recyclePivotsIter() {
...
-                        //Remove resolution step, remove n, ant without pivots gains n resolvents
-                        bool choose_ant1 = false;
-                        //Pivot not in ant1 and not in ant2
-                        if (not piv_in_ant1 and not piv_in_ant2) {
-                            // Choose one of the two antecedents heuristically
-                            choose_ant1 = chooseReplacingAntecedent(n);
-                        } else {
-                            choose_ant1 = not piv_in_ant1 and piv_in_ant2;
-                        }
+                        choose_ant1 = not piv_in_ant1 and piv_in_ant2;
```
aehyvari, line 239: Here we know from the assert in line 230 and the condition in line 235 that either `pid_in_ant1 and not piv_in_ant2`, or `not piv_in_ant1 and piv_in_ant2`.  Maybe it would be more natural to say simply
```
choose_ant1 = not piv_in_ant1;
```
since the latter is implied.  Meaning that we `choose_ant1` if and only if the pivot is not there.

src/proof/PGTransformationAlgorithms.cc
blishko: I would keep the `assert`, I like asserts, asserts are expressive :)
For me the question is if we should throw the exception, so any problem is caught also in Release and not only in Debug.
Since the proof related code is quite complex, I think it makes sense to keep it as well.

src/proof/PGTransformationAlgorithms.cc
blishko: Yeah I was considering it, then decided to not change the code, but I agree with you. I'll update it now.

src/proof/PGTransformationAlgorithms.cc
aehyvari: Ok!

aehyvari (approved):

PR #266, opened 2021-07-09
Proof transformations cleanup - part 5 of N: ProofGraph::proofTransformAndRestructure

blishko:
This PR deals with cleaning up the method `ProofGraph::proofTransformAndRestructure`.

It adds two unit tests for checking basic functionality of the method, the second one exposing a path in the code that previously triggered error. It is a situation similar to the one in #264, i.e., while restructuring and reducing the proof, the algorithm might run into a situation where a node has identical antecedents. Previously this was reported as an error and the program terminated.
However, as the unit test shows, such a situation can occur for a valid proof as input. The same solution as in #264 can be applied, the algorithm only needs to replace the node with its (single) antecedent and continue as normal.

Apart from this, small style fixes has been applied.
Depends on #264.

### DIFF
--- a/src/proof/PGTransformationAlgorithms.cc
+++ b/src/proof/PGTransformationAlgorithms.cc
@@ -555,68 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int 
     }
 
     double init_time = cpuTime();
-    assert(!(max_num_loops > 0 && left_time > 0));
+    assert(not (max_num_loops > 0 and left_time > 0));
     //Flag to check if in a loop at least a transformation has been applied
     bool some_transf_done;
     long curr_num_loops = 0;
-    RuleContext ra1, ra2;
     std::deque<clauseid_t> q;
-    clauseid_t id;
     // Main external loop
     do {
+        assert(isResetVisited1() and isResetVisited2());
         // Enqueue leaves first
         q.assign(leaves_ids.begin(), leaves_ids.end());
         some_transf_done = false;
         do {
-            id = q.front();
+            clauseid_t id = q.front();
             assert(id < getGraphSize());
             q.pop_front();
             ProofNode * n = getNode(id);
             //Node not visited yet
-            if (n != nullptr && !isSetVisited1(id)) {
+            if (n and not isSetVisited1(id)) {
                 // Wait if one antecedent is not visited yet;
                 // the node will be enqueued anyway by that antecedent
-                if ((n->getAnt1() != nullptr && !isSetVisited1(n->getAnt1()->getId()))
-                    || (n->getAnt2() != nullptr && !isSetVisited1(n->getAnt2()->getId()))) {
+                if ((n->getAnt1() and not isSetVisited1(n->getAnt1()->getId()))
+                    or (n->getAnt2() and not isSetVisited1(n->getAnt2()->getId()))) {
                     continue;
                 }
                 // ELSE, both antecedents ready, process the node
                 // Mark node as visited
                 setVisited1(id);
                 // Non leaf node
-                if ((n->getAnt1() != nullptr || n->getAnt2() != nullptr)) {
+                if ((n->getAnt1() or n->getAnt2())) {
                     assert(n->getAnt1());
-                    assert (n->getAnt2());
-                    assert((n->getAnt1()->getAnt1() != nullptr && n->getAnt1()->getAnt2() != nullptr)
-                           || (n->getAnt1()->getAnt1() == nullptr && n->getAnt1()->getAnt2() == nullptr));
-                    assert((n->getAnt2()->getAnt1() != nullptr && n->getAnt2()->getAnt2() != nullptr)
-                           || (n->getAnt2()->getAnt1() == nullptr && n->getAnt2()->getAnt2() == nullptr));
+                    assert(n->getAnt2());
+                    assert((n->getAnt1()->getAnt1() and n->getAnt1()->getAnt2())
+                           or (n->getAnt1()->getAnt1() == nullptr and n->getAnt1()->getAnt2() == nullptr));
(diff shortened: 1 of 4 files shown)


src/proof/PGTransformationAlgorithms.cc
```diff
@@ -555,73 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
-                           || (n->getAnt2()->getAnt1() == nullptr && n->getAnt2()->getAnt2() == nullptr));
-
-                    // NOTE not clear how this might happen
-                    assert(n->getAnt1() != n->getAnt2());
-                    if (n->getAnt1() == n->getAnt2()) {
-                        opensmt_error("Bad proof structure!");
-                    }
+                    assert(n->getAnt2());
+                    assert((n->getAnt1()->getAnt1() and n->getAnt1()->getAnt2())
+                           or (n->getAnt1()->getAnt1() == nullptr and n->getAnt1()->getAnt2() == nullptr));
```
aehyvari, line 590: `or (not n->getAnt1()->getAnt1() and not n->getAnt1()->getAnt2()));`?

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -555,73 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
+                    assert((n->getAnt2()->getAnt1() and n->getAnt2()->getAnt2())
+                           or (n->getAnt2()->getAnt1() == nullptr && n->getAnt2()->getAnt2() == nullptr));
```
aehyvari, line 592: `or (not n->getAnt2()->getAnt1() and not n->getAnt2()->getAnt2()));`?

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -665,48 +656,51 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
 
                         for (clauseid_t resolvent : resolvents) {
                             assert(resolvent < getGraphSize());
                             ProofNode * res = getNode(resolvent);
                             assert(res);
-                            assert(res->getAnt1() == n || res->getAnt2() == n);
+                            assert(res->getAnt1() == n or res->getAnt2() == n);
                             if (res->getAnt1() == n) { res->setAnt1(replacing); }
                             else if (res->getAnt2() == n) { res->setAnt2(replacing); }
                             else {opensmt_error("Error in the structure of the proof"); }
```
aehyvari, line 687: Raise an exception instead.

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -555,73 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
-    RuleContext ra1, ra2;
     std::deque<clauseid_t> q;
-    clauseid_t id;
     // Main external loop
     do {
+        assert(isResetVisited1() and isResetVisited2());
         // Enqueue leaves first
         q.assign(leaves_ids.begin(), leaves_ids.end());
         some_transf_done = false;
         do {
```
aehyvari, line 569: Can we change this to a `while (not q.empty())`?

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -555,73 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
-    assert(!(max_num_loops > 0 && left_time > 0));
+    assert(not (max_num_loops > 0 and left_time > 0));
     //Flag to check if in a loop at least a transformation has been applied
     bool some_transf_done;
     long curr_num_loops = 0;
-    RuleContext ra1, ra2;
     std::deque<clauseid_t> q;
-    clauseid_t id;
     // Main external loop
     do {
```
aehyvari, line 564: I think we can change this to `while (...)`

src/proof/PGTransformationAlgorithms.cc
```diff
@@ -555,73 +555,64 @@ void ProofGraph::proofTransformAndRestructure(const double left_time, const int
...
+                            mergeClauses(n->getAnt1()->getClause(), n->getAnt2()->getClause(), n->getClause(), n->getPivot());
                             setVisited2(id);
                         }
                         // NOTE extra check
                         if (proofCheck() > 1) { checkClause(n->getId()); }
 
                         if (do_transf) {
+                            RuleContext ra1, ra2;
                             //Look for rules applicability
                             getRuleContext(n->getAnt1()->getId(), n->getId(), ra1);
```
aehyvari, line 617: The return value of these are not checked.  It looks like changing the signature to `RuleContext ProofGraph::getRuleContext(clauseid_t idv3, clauseid_t idv)`.  Do you think that'd be better?

src/proof/PGTransformationAlgorithms.cc
blishko: Yes, I think we can, but we have to be careful about the loop iteration counter, since we increment it at the end of the loop.

src/proof/PGTransformationAlgorithms.cc
blishko: Yes! Changed.

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc

src/proof/PGTransformationAlgorithms.cc
blishko: Signature changed as you proposed.

src/proof/PGTransformationAlgorithms.cc
blishko: Changed. Nothing needs to change about the loop iteration counter, I was confused.

aehyvari (approved):

PR #269, opened 2021-07-15
Prevent re-introduction of duplicate literals into clauses

blishko:
Currently if a clause to be added to the SAT solver has duplicate literal which is know to be false, and we are in the interpolating mode, all occurences of such literal would be added back to the clause for proof logging purpose.

Since the literals are sorted, we can prevent this by checking if we have already encountered such literal before.


blishko:
I will try to come up with a test case where the previous situation would actually cause an error in the proof handling code, but I might not be able to find a small one.

I was seeing the problems in the CHC benchmarks, but I am not able to create tests cases from those runs, yet.

aehyvari (approved):
Ok.  This doesn't look like a bug that would be easily reintroduced anyway.

blishko:
I managed to find a relatively simple test in the end.

PR #270, opened 2021-07-20
Models

aehyvari:


src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+        const ValuationNode * valuationNode = symbolSigVal.second.second;
+        SRef sr = logic.getSortRef(symbolSigVal.first);
+        PTRef body = valuationTreeToFunctionBody(valuationNode, sr);
+        TemplateFunction templateFun(std::move(symbolSigVal.second.first), body);
+        builtDefinitions[symbolSigVal.first] = std::move(templateFun);
+    }
+    return std::unique_ptr<Model>(new Model(logic, assignment, builtDefinitions));
+}
+
+void ModelBuilder::addToTheoryFunction(SymRef sr, vec<PTRef> vals, PTRef val)
```
blishko, line 20:
```suggestion
void ModelBuilder::addToTheoryFunction(SymRef sr, vec<PTRef> const & vals, PTRef val)
```

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+{
+    if (logic.getSortRef(sr) != logic.getSortRef(val)) {
+        throw OsmtApiException("Incompatible sort for symbol valuation");
+    }
+    Symbol & sym = logic.getSym(sr);
```
blishko, line 25:
```suggestion
    Symbol sym = logic.getSym(sr);
```

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+    }
+    for (int i = 0; i < vals.size(); i++) {
+        if (sym[i] != logic.getSortRef(vals[i])) {
+            throw OsmtApiException("Incorrect valuation for symbol: sort mismatch");
+        }
+    }
+
+    if (not hasTheoryFunction(sr)) {
+        vec<PTRef> formalArgs; formalArgs.capacity(vals.size());
+        std::string symName(logic.getSymName(sr));
```
blishko, line 37: This seams to be an unused variable? This should be caught by the compiler, we should check the CI.

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+        for (PTRef v : vals) {
+            std::stringstream ss;
+            ss << formalArgPrefix << uniqueNum++;
+            formalArgs.push(logic.mkVar(logic.getSortRef(v), ss.str().c_str()));
+        }
+        FunctionSignature templateSig(logic.protectName(logic.getSymName(sr)), std::move(formalArgs), logic.getSortRef(sr));
+        definitions.insert({sr,opensmt::pair<FunctionSignature,ValuationNode*>{std::move(templateSig), nullptr}});
+    }
+    auto & signatureAndValuation = definitions.at(sr);
+    FunctionSignature & templateSignature = signatureAndValuation.first;
```
blishko, line 50: `templateSignature` can be inlined, it has only one usage.

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+    assert(valuation.size() >= 1);
+
+    if (root == nullptr) {
+        root = valuationNodeFactory.getValuationNode(valuation[0].first, PTRef_Undef);
+    }
+
+    ValuationNode * current = root;
+
+    for (int i = 0; i < valuation.size(); i++) {
+        auto & varVal = valuation[i];
```
blishko, line 70: `varVal` can be inlined

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
+
+        bool foundChild = false;
+        for (auto child: *current) {
+            if (child->getValue() == val) {
+                current = child;
+                foundChild = true;
+                break;
+            }
+        }
+        if (foundChild) {
```
blishko, line 83: This control-flow can be simplified, no?

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
-        PTRef target = copy[key];
-        PTRef val = model->evaluate(target);
-        assert(logic.isConstant(val));
-        assignCopy.insert(std::make_pair(key, val));
     }
-    this->assignment = std::move(assignCopy);
+    return root;
 }
+
+// Copied from TermVisitor.  Maybe it'd be possible to template TermVisitor to handle also ValuationNodes
```
blishko, line 95: Maybe we should try to build some kind of generic DAG visitor...

src/rewriters/Rewriter.h
```diff
@@ -79,6 +79,12 @@ class DefaultRewriterConfig {
 public:
     virtual bool previsit(PTRef term) { return true; } // should continue visiting
     virtual PTRef rewrite(PTRef term) { return term; } // don't do anything
+    virtual bool treatIteAsVar() const { return true; }
+};
+
+class IteRewriterConfig : public DefaultRewriterConfig {
```
blishko, line 85: I was hoping we do not need this specialization for ITEs (and remove `treatIteAsVar` from `Config`).
Can we do that or is this necessary?

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -2251,9 +2252,42 @@ void CoreSMTSolver::fillBooleanVars(ModelBuilder &modelBuilder) {
     Logic& logic = theory_handler.getLogic();
     for (Var v = 0; v < model.size(); ++v) {
         PTRef atom = theory_handler.varToTerm(v);
-        if (logic.isBoolAtom(atom) && model[v] != l_Undef) {
-            PTRef val = model[v] == l_True ? logic.getTerm_true() : logic.getTerm_false();
+        PTRef target_tr = theory_handler.getSubstitution(atom);
```
blishko, line 2255: Here I was also hoping we can get rid of the substitutions when building the model...

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -213,23 +215,129 @@ void Egraph::getConflict( bool deduction, vec<PtAsgn>& cnfl )
...
+ * @param orig_tr the original pterm reference (with arity > 0)
+ * @param target_er the target enode reference
+ * @param substs the substitutions (used for obtaining values for the arguments of orig_tr
+ */
+void Egraph::addTheoryFunctionEvaluation(ModelBuilder & modelBuilder, PTRef orig_tr, ERef target_er, const MapWithKeys<PTRef,PTRef,PTRefHash>& substs) const {
+    ERef orig_er = enode_store.getERef(orig_tr);
+    vec<ERef> args = enode_store.getArgTermsAsVector(orig_er);
+    assert(args.size() > 0);
+    vec<PTRef> vals; vals.capacity(args.size());
+    for (ERef child_er: args) {
```
blishko, line 325:
```suggestion
    for (ERef child_er : args) {
```

blishko (commented):
Looks good! Just left a few comments.

I would just really like if we would not need  the specialized IteRewriterConfig and if we would not need the substitutions for UF models.

blishko:
We should also add a test that bit us in SMT-COMP, the `define-fun` with `ite` inside. Or a test like that already in here?

blishko:
I tried ignoring the substitutions in `Egraph::fillTheoryFunctions` and the model regression tests seem to went through.
If I have not missed something, it seems worthwhile to try to remove the substitutions from model building.

regression_models/instances/different_default_sorts.smt2
```diff
@@ -0,0 +1,11 @@
...
+(set-logic QF_UF)
+(declare-sort x4 0)
+(declare-sort x1 0)
+(declare-fun x2 () x1)
+(declare-fun x6 (x4) x1)
+(declare-fun x3 (x1) x4)
+(declare-fun x5 () x4)
+(assert (= (x3 x2) x5))
+(assert (= (x6 x5) x2))
+(check-sat)
```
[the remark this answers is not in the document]
aehyvari: Yes, (get-model) is inserted by the smt-comp scrambler.  The system is quite complicated... Let's see if it turns out to be sustainable.

src/models/Model.cc
```diff
@@ -3,6 +3,30 @@
...
+           const TemplateFunction & templFun = entry.second;
+           if (not logic.isUF(sr)) { return false; }
+           const Symbol & s = logic.getSym(sr);
+           if (templFun.getName() != logic.getSymName(sr)) { return false; }
+           if (s.nargs() != templFun.getArgs().size_()) { return false; }
+           if (logic.getSortRef(sr) != templFun.getRetSort()) { return false; }
+           for (auto i = 0; i < (int)s.nargs(); i++)
+               if (s[i] != logic.getSortRef(templFun.getArgs()[i])) { return false; }
+           return true;
+       }));
```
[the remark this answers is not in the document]
aehyvari: `SymbolDefinition` is just an alias for `unordered_map`, but I'll make it into a separate function `Model::isCorrect`.

src/models/ModelBuilder.cc
```diff
@@ -3,23 +3,136 @@
...
-    subst.copyTo(copy);
-    logic.substitutionsTransitiveClosure(copy);
-    auto assignCopy = this->assignment;
-    auto model = this->build();
-    for (auto key : copy.getKeys()) {
-        if (logic.isIte(key)) {
-            // We store only values of reals variables
+std::unique_ptr<Model> ModelBuilder::build() {
+    std::unordered_map<SymRef,TemplateFunction,SymRefHash> builtDefinitions;
+    for (auto & symbolSigVal : definitions) {
```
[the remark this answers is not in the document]
aehyvari: Apparently it can't because we move the signature a few lines below for the constructor of `TemplateFunction`.  We could copy it instead of course, but I'm not sure if that's the right solution.  This also means that `ModelBuilder::build()` should be called only once since `definitions` will have undefined values after the first call.

src/models/ModelBuilder.cc
aehyvari: Ok.

src/models/ModelBuilder.cc
aehyvari: The copy doesn't seem necessary.  Maybe you meant `const Symbol & sym = logic.getSym(sr)` ?

src/models/ModelBuilder.cc
aehyvari: Yes, it's not caught for some reason.  Could it be because it uses the `std::string`'s `char*` constructor and could therefore have a side effect?

src/models/ModelBuilder.cc
aehyvari: Ok.

src/models/ModelBuilder.cc
aehyvari: Ok.

src/models/ModelBuilder.cc
aehyvari: OK.

src/models/ModelBuilder.cc
aehyvari: Indeed, something like that.

src/models/ModelBuilder.cc
blishko: Ah, sorry, I got `Symbol` confused with `SymRef`. You are right.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: Ok.

aehyvari:
> We should also add a test that bit us in SMT-COMP, the `define-fun` with `ite` inside. Or a test like that already in here?

I thought we had it but couldn't find it.  Anyway now I added a regression test.

src/models/ModelBuilder.cc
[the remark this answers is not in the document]
blishko: OK, I missed the `move`. It's fine this way.

src/models/Model.cc
```diff
@@ -23,16 +47,74 @@ PTRef Model::evaluate(PTRef term) {
...
+        PTRef val;
+        if (symDef.find(symbol) != symDef.end()) {
+            const TemplateFunction & tfun = symDef.at(symbol);
+            const vec<PTRef> & tfunArgs = tfun.getArgs();
+            MapWithKeys<PTRef,PTRef,PTRefHash> substMap;
+            for (int i = 0; i < nargs.size(); i++) {
+                substMap.insert(tfunArgs[i], nargs[i]);
+            }
+            PTRef root = tfun.getBody();
+            val = IteSubstitutor(logic, substMap).rewrite(root);
```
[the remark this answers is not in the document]
aehyvari: I think we don't, I'm going to remove it.

src/rewriters/Rewriter.h
aehyvari: I'll remove it.

src/smtsolvers/CoreSMTSolver.cc
aehyvari: I have also some experimental evidence now that this code is not necessary.

aehyvari:
> I tried ignoring the substitutions in `Egraph::fillTheoryFunctions` and the model regression tests seem to went through.
> If I have not missed something, it seems worthwhile to try to remove the substitutions from model building.

I ran model validation on all QF_UF benchmarks with a 10-second timeout.  I'm quite optimistic now that we can remove this code.

blishko (approved):
The recent changes looks great!
I've only noticed one leftover unused statement, but otherwise I believe we can merge this.

src/models/ModelBuilder.cc
blishko: That's possible, I don't know.

src/tsolvers/bvsolver/BitBlaster.cc
```diff
@@ -2116,39 +2117,32 @@ PTRef BitBlaster::simplify( PTRef formula )
...
-        for (int j = 0; j < bs[blast].size(); j++)
-        {
+        for (int j = 0; j < bs[blast].size(); j++) {
             PTRef b = bs[blast][j];
-            auto valPair = mainSolver.getValue(b);
-            char* val = valPair.val;
-            assert(strcmp(val,Logic::tk_true) == 0 || strcmp(val, Logic::tk_false) == 0);
-            Real bit = strcmp(val,Logic::tk_true) == 0 ? 1 : 0;
+            auto val = localModel->evaluate(b);
+            logic.getTerm_true();
```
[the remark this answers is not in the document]
aehyvari: OK.

PR #273, opened 2021-07-24
Interpolating mode: Ensure SAT solver is aware of empty frames

### LINKED ISSUE #272: Incorrect answer in incremental solving with interpolation enabled
For the following script OpenSMT produces incorrect answer:
```
(set-option :produce-interpolants true)
(set-logic QF_LRA)
(push 1)
(push 1)
(assert false)
(check-sat)
(pop 1)
(check-sat)
```

The expected answer is
```
unsat
sat
```

The obtained answer is
```
unsat
unsat
```

The behaviour is correct when interpolation is disabled (first line in the script is removed).

blishko:
Given how we detect which frame is the last unsatisfiable frame in the SAT solver, it must be aware of all the frames, otherwise the numbers do not match and incorrect information is propagated about which frame is the last unsatisfiable one.

Previously in interpolating mode, empty frames or frames containing only `true` were not passed to the SAT solver.

Fixes #272.


blishko:
This fix is not ideal, but it seems to be the quickest way for now. Both the partitioning code and code related to incremental solving could use some refactoring/rethinking...

aehyvari (approved):
It does look funny, but since this is an urgent fix, let's have it done.

Issue #271, opened 2021-07-21
Behaviour change in Substitutor class


blishko:
Either there was a bug in `Substitutor` before or and the change in #260 fixed it, or it was correct before and not ti is buggy, I am not sure which one is true.
In any case, it's behaviour has changed for certain special cases and now it can break the invariant of `Map` that the same key should not be inserted twice.

Given a formula `and a b` and a substitution map `a -> true`, `b -> a` it now behaves differently than before.

aehyvari:
the `mtl::Map` implementation does not work as expected in case the same key is inserted twice (with different data).  This causes in some cases problems.  If a substitution is recorded first in https://github.com/usi-verification-and-security/opensmt/blob/c0e3bd029fc61b61659cec7b75fd306faa03eff5/src/rewriters/Rewriter.h#L61 and then in https://github.com/usi-verification-and-security/opensmt/blob/c0e3bd029fc61b61659cec7b75fd306faa03eff5/src/rewriters/Rewriter.h#L66 only the first substitution will be visible (at least in some cases?).

aehyvari:
I don't want to change the behaviour of mtl::Map.  We could change the Rewriter code instead?  A simple fix, by adding
```C++
substitutions.remove(currentRef);
```
above https://github.com/usi-verification-and-security/opensmt/blob/c0e3bd029fc61b61659cec7b75fd306faa03eff5/src/rewriters/Rewriter.h#L66
fixes the problem and goes through in our ci.  There could be a more elegant fix, but I'm not sure about that.

blishko:
Yes, the second insertion is lost, I believe. Or rather, it will be inserted into the map, but lookup will always find the first inserted value.

I don't think adding `remove` is the proper solution, because it has the precondition that the key is already in the map, but that might not be the case when line 66 is reached.
I would suggest to test if the key is present and if yes then overwrite it, otherwise insert it.

My question for the discussion would be this: Should the rewriter try to rewrite a term that has been changed because its children have been rewritten? I am not sure if we require this behaviour somewhere or not, but it made me think that this does not behave as a simultaneous substitution, which is desired in some cases.
For example, the behaviour in the example I gave at the beginning.
Given a substitution `a -> true`, `b->a` and a formula `and a b` there are two possible behaviours.
Applying simultaneous substitution yield `a`, while applying the substitutions gradually (on the parent after children have been rewritten) yields `true`.
[...]

PR #275, opened 2021-08-05
LIA preprocessing: Rewrite div/mod constraints in the input formula

blishko:
This PR adds step for rewriting any div/mod constraints in the input formula to the theory preprocessing.
This step is added for `LIALogic` while for `LRALogic` it is basically a no-op.
After this change, OpenSMT will be able to decide satisfiability of formulas with div/mod constraints (if it can decide it for the rewritten pure LIA formula).

The rewriter is already present in the code (added in #221). The rewriting works as follows:
1. For any `div/mod` constraint `t div N` (or `t mod N`) we introduce two variables `d` and `m` representing values of the terms `t div N` and `t mod N`, respectively. 
2. Then a formula `F` is rewritten to `F'` such that `F' = F[t div N -> d; t mod N -> m] and t = d * N + m and 0 <= m <= |N| - 1`.
3. `F'`is then equisatisfiable with `F`. Any model of `F` can be extended to the model of `F'` by setting the value of `d` to the value of `t div N` (and the same for `m`). In the other direction, any model of `F'` is a model if `F` (ignoring the fresh variables `d` and `m`).


aehyvari (commented):
This looks quite reasonable.  Should we include some regression-style unit tests?

blishko:
Yeah, we probably should. I will add some basic ones. If you have any suggestions, let me know.

blishko:
I had to update two assertions regarding arithmetic equalities elimination to use `isNumVarLike` instead of `isNumVarOrIte` to handle also `div/mod` terms.

I also added some very basic regression tests.

blishko:
Given that we now have `RDLogic` and `IDLogic` that also use the `LATheory` template, I set the default for `rewriteDivMod` to do nothing and only implemented the specialization for `LIALogic`.
Before the template was undefined and specializations were defined separately for `LIALogic` and `LRALogic`, but this meant adding the "do nothing" definition for each new Logic separately. Having the default seems like a better option.

aehyvari:
As discussed here's my suggestion on dealing with the div/mod question so that it won't bite us back with combinations of linear integer arithmetic and uninterpreted functions, where it would be possible for the user to define functions called `div` and `mod`.   There's also a fix for the nested `div`s and `mod`s.

aehyvari:
So basically the idea is that while working over the API the functions mkIntDiv and mkIntMod are immediately available.  The feature needs to be specifically enabled only when the features are accessed through the automatic symbol resolution, such as when called from Interpret.  This is then implemented on `Interpret`: it checks the status from `SMTConfig` every time it sees an `(assert ...)` statement in the AST.  Do you think this is practical enough?

src/rewriters/Rewriter.h
```diff
@@ -57,12 +58,10 @@ class Rewriter {
...
             }
             PTRef newTerm = needsChange ? logic.insertTerm(term.symb(), newArgs) : currentRef;
-            if (needsChange) {
-                substitutions.insert(currentRef, newTerm);
-            }
             // The reference "term" has now been possibly invalidated! Do not access it anymore!
+
             PTRef rewritten = cfg.rewrite(newTerm);
-            if (rewritten != newTerm) {
+            if (rewritten != newTerm or needsChange) {
```
blishko, line 64: Nice! This is simpler and more elegant that the solutions we discussed before.

blishko (commented):
This looks good. I'm fine with the option for allowing `div/mod` from the interpreter and the fix of the `Rewriter` is very nice.

blishko:
@aehyvari, from my perspective, your changes are great. 
I cannot approve PR that I have started, so you will have to do that, if you are also satisfied with this.

blishko:
I believe your changes in the `Substitutor` also resolve the main problem I raised in #271.

aehyvari (approved):

PR #276, opened 2021-08-12
LASolver: Replace LABoundList in bounds data structure

blishko:
This prevents a memory leak in current implementation of the addition of new bounds after the solver has been initialized.

### DIFF
--- a/src/tsolvers/lasolver/LABounds.cc
+++ b/src/tsolvers/lasolver/LABounds.cc
@@ -1,19 +1,19 @@
 #include "LABounds.h"
 #include "LRALogic.h"
 
-LABound::LABound(BoundT type, LVRef var, const Delta& delta, int id)
+LABound::LABound(BoundT type, LVRef var, Delta && delta, int id)
     : type(type.t)
     , bidx(UINT32_MAX)
     , id(id)
     , var(var)
-    , delta(delta)
+    , delta(std::move(delta))
 {}
 
-LABoundRef LABoundAllocator::alloc(BoundT type, LVRef var, const Delta& delta)
+LABoundRef LABoundAllocator::alloc(BoundT type, LVRef var, Delta && delta)
 {
     uint32_t v = RegionAllocator<uint32_t>::alloc(laboundWord32Size());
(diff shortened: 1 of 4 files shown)


blishko:
I ran SMT-LIB benchmarks a few weeks ago.
This change does not have any visible effect on LRA benchmarks, but it has positive effect on LIA benchmarks. There we had a lot of memouts due to the memory leak when new bounds were added as part of the LIA splits.
With this change, there are no more memouts. Some were solved now, but many turned into timeouts, so now the runtime for the whole benchmark set has increased.

blishko:
LRA: 
[comparison_lra_master_vs_del-bound-list.pdf](https://github.com/usi-verification-and-security/opensmt/files/6974233/comparison_lra_master_vs_del-bound-list.pdf)

blishko:
LIA:
[comparison_lia_master_vs_del-bound-list.pdf](https://github.com/usi-verification-and-security/opensmt/files/6974234/comparison_lia_master_vs_del-bound-list.pdf)

aehyvari:
I added a commit that tries to avoid an unnecessary copy of opensmt::Real in bounds construction.  The change is a little unrelated to this, but could fit nicely in the PR since it touches nearby code.  Let me know if you think it's ok.

src/tsolvers/lasolver/LABounds.cc
```diff
@@ -1,19 +1,19 @@
 #include "LABounds.h"
 #include "LRALogic.h"
 
-LABound::LABound(BoundT type, LVRef var, const Delta& delta, int id)
+LABound::LABound(BoundT type, LVRef var, Delta && delta, int id)
     : type(type.t)
     , bidx(UINT32_MAX)
     , id(id)
     , var(var)
     , delta(delta)
```
blishko, line 9: To really avoid the copy, we need to change also this to `delta(std::move(delta))`, I believe.

aehyvari: ok.

aehyvari (approved):

PR #281, opened 2021-08-17
Fix: Implemented the standard-conforming version of `declare-const`

aehyvari:
Previously we allowed the use of `declare-const` for declaring constants in the sense that
```
(set-logic QF_UF)
(declare-sort U 0)
(declare-const () a U)
(declare-const () b U)
(assert (= a b))
(check-sat)
```
would return `unsat`.  This does not conform to the smtlib2 standard for two reasons:
 1.`declare-const` does not take the empty parentheses as arguments
 2. `(declare-const a U)` is an abbreviation for `(declare-fun () a U)`

This fix removes the special meaning of `declare-const` and imposes the standard syntax.  The special meaning is still available through the API, by calling `Logic::mkConst`.


test/unit/test_Model.cc
```diff
@@ -156,6 +156,41 @@ TEST_F(UFModelBuilderTest, test_functionModel) {
...
 
+class UFConstModelTest : public ::testing::Test {
+protected:
+    UFConstModelTest() : logic(), ms(logic, c, "uf-solver") {
+        char * msg;
+        SRef U = logic.declareSort("U", &msg);
+        SymRef fs = logic.declareFun("f", U, {U}, &msg, false);
+        x = logic.mkVar(U, "x");
+        a = logic.mkConst(U, "a");
+        f_a = logic.mkUninterpFun(fs, {x});
```
blishko, line 167:
```suggestion
        f_a = logic.mkUninterpFun(fs, {a});
```

With this change, the unit test shows the problem I mentioned.

aehyvari: Fixed in the below commit, thanks!

blishko (approved):
Looks good! Can you add also a changelog entry for the change in `declare-const` command of SMTLIB2?

PR #284, opened 2021-08-20
Switch to C++17

blishko:
This PR switches C++ version to C++17 and show examples how post-C++11 features can be used in our code base.

These are (showed here):
 * `make_unique`
 * structured bindings
 * initializer inside `if` statements


blishko:
I am wondering, if we do switch to C++17, then the next release should also probably be a new minor version, not just a patch.

aehyvari (approved):
c++20 support doesn't seem to be there yet, so we decided to stick to c++17.

PR #286, opened 2021-09-03
Introduce and use Span data structure

blishko:
I would like to propose our version of `Span` object, which is basically a (non-owning) view over a contiguous range of  elements in memory (same as C++20's `std::span`).
The proposed version contains only the bare minimum we require and corresponds to `ConstSpan`, where the elements in memory cannot be modified.

In my opinion this can save us a bunch of temporary `vec`s that we now create when creating terms.


blishko:
In the end, we figured out that the Span is not the best way to pass the arguments for creating new terms and we went with #292.

PR #290, opened 2021-09-08
Unify arithmetic solvers - part 1

blishko:
This PR represents the first step towards unifying `LRASolver` and `LIASolver` under one and only `LASolver`.
All the functionality from `LRASolver` is moved to `LASolver`, with `LIASolver` still customizing certain points.

To bring `LIASolver`'s functionality under `LASolver`, we probably need first a unified `LALogic` that is able to tell if a given term has sort `Real` or `Int`.


aehyvari (commented):
That seems to have been quite pain-free.  Can we move the contents of `src/tsolvers/LRASolver/` to `src/tsolvers/LASolver/` and delete the former?

aehyvari (approved):

PR #293, opened 2021-09-14
Normalize arithmetic equalities.

### LINKED ISSUE #267: Consider normalizing equalities in Linear Arithmetic
The inequalities in Linear Arithmetic are normalized on creation in LALogic.
We should consider normalizing also LA equalities in a similar manner.
Currently, it is possible to create a term `(= (+ x y) (+ x 1) ` which could be immediately simplified to `(= y 1)` but it is currently not.

blishko:
Depends on #292.
Resolves #267.

Currently, arithmetic equalities are not normalized, they are created as given, in the same way as equalities for uninterpreted sorts.
This PR adds normalization process to arithmetic equalities on creation, which works almost the same as the normalization for arithmetic inequalities.

### DIFF
--- a/src/logics/LIALogic.cc
+++ b/src/logics/LIALogic.cc
@@ -120,7 +120,17 @@ LIALogic::LIALogic() :
     sym_Int_EQ = term_store.lookupSymbol(tk_equals, {term_Int_ZERO, term_Int_ZERO});
 }
 
+/**
+ * Normalizes a sum term a1x1 + a2xn + ... + anxn + c such that the coefficients of non-constant terms are coprime integers
+ * Additionally, the normalized term is separated to constant and non-constant part, and the constant is modified as if
+ * it was placed on the other side of an equality.
+ * Note that the constant part can be a non-integer number after normalization.
+ *
+ * @param sum
+ * @return Constant part of the normalized sum as LHS and non-constant part of the normalized sum as RHS
+ */
 opensmt::pair<FastRational, PTRef> LIALogic::sumToNormalizedPair(PTRef sum) {
+    assert(isNumPlus(sum));
     vec<PTRef> varFactors;
     PTRef constant = PTRef_Undef;
     Pterm const & s = getPterm(sum);
--- a/src/logics/LIALogic.cc
+++ b/src/logics/LIALogic.cc
@@ -120,8 +120,7 @@ LIALogic::LIALogic() :
     sym_Int_EQ = term_store.lookupSymbol(tk_equals, {term_Int_ZERO, term_Int_ZERO});
 }
 
-
-PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
+opensmt::pair<FastRational, PTRef> LIALogic::sumToNormalizedPair(PTRef sum) {
     vec<PTRef> varFactors;
     PTRef constant = PTRef_Undef;
     Pterm const & s = getPterm(sum);
@@ -150,7 +149,7 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
     bool changed = false; // Keep track if any change to varFactors occurs
 
     bool allIntegers = std::all_of(coeffs.begin(), coeffs.end(),
-                                   [](opensmt::Number const & coeff) { return coeff.isInteger(); });
+        [](opensmt::Number const & coeff) { return coeff.isInteger(); });
     if (not allIntegers) {
         // first ensure that all coeffs are integers
         using Integer = FastRational; // TODO: change when we have FastInteger
@@ -203,8 +202,19 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
     // in LIA we can strengthen the inequality to
     // ceiling(-constantValue) <= normalizedSum
     constantValue.negate();
-    constantValue = constantValue.ceil();
-    return mkFun(get_sym_Num_LEQ(), {mkConst(constantValue), normalizedSum});
+    return {std::move(constantValue), normalizedSum};
+}
+
+
+PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
+    auto[lhsVal, rhs] = sumToNormalizedPair(sum);
+    return mkFun(get_sym_Num_LEQ(), {mkConst(lhsVal.ceil()), rhs});
+}
+
+PTRef LIALogic::sumToNormalizedEquality(PTRef sum) {
+    auto[lhsVal, rhs] = sumToNormalizedPair(sum);
+    if (not lhsVal.isInteger()) { return getTerm_false(); }
+    return mkFun(get_sym_Num_EQ(), {mkConst(lhsVal), rhs});
 }
 
 PTRef LIALogic::insertTerm(SymRef sym, vec<PTRef> && terms) {
--- a/src/logics/LIALogic.cc
+++ b/src/logics/LIALogic.cc
@@ -116,6 +116,8 @@ LIALogic::LIALogic() :
     //sym_store[sym_Real_ITE].setLeftAssoc();
     sym_store[sym_Int_ITE].setNoScoping();
     sym_store.setInterpreted(sym_Int_ITE);
+
+    sym_Int_EQ = term_store.lookupSymbol(tk_equals, {term_Int_ZERO, term_Int_ZERO});
 }
 
 
--- a/src/logics/LIALogic.cc
+++ b/src/logics/LIALogic.cc
@@ -134,13 +134,13 @@ opensmt::pair<FastRational, PTRef> LIALogic::sumToNormalizedPair(PTRef sum) {
     vec<PTRef> varFactors;
     PTRef constant = PTRef_Undef;
     Pterm const & s = getPterm(sum);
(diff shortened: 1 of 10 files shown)


src/logics/LIALogic.cc
```diff
@@ -201,8 +212,19 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
...
     // ceiling(-constantValue) <= normalizedSum
     constantValue.negate();
-    constantValue = constantValue.ceil();
-    return mkFun(get_sym_Num_LEQ(), {mkConst(constantValue), normalizedSum});
+    return {std::move(constantValue), normalizedSum};
+}
+
+
+PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
+    auto[lhsVal, rhs] = sumToNormalizedPair(sum);
```
aehyvari, line 220: missing a space

src/logics/LIALogic.cc
```diff
@@ -201,8 +212,19 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
...
+    return mkFun(get_sym_Num_LEQ(), {mkConst(lhsVal.ceil()), rhs});
+}
+
+PTRef LIALogic::sumToNormalizedEquality(PTRef sum) {
+    auto[lhsVal, rhs] = sumToNormalizedPair(sum);
```
aehyvari, line 225: Missing a space

src/logics/LIALogic.cc
```diff
@@ -201,8 +212,19 @@ PTRef LIALogic::sumToNormalizedInequality(PTRef sum) {
     // in LIA we can strengthen the inequality to
     // ceiling(-constantValue) <= normalizedSum
```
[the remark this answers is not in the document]
aehyvari: On line 2010 above we assume that the sum is normalised, which should indeed be the case with the current code.  I wonder would it make sense to write an assert for that?

aehyvari (commented):
This looks great!

I tried out a refactoring where I put to a function `std::pair<FastRational, vec<PTRef>>LALogic::getConstantAndFactors(PTRef)` the common part of `LIALogic::sumToNormalizedPair` and `LRALogic::sumToNormalizedPair`.  If you want I can push it here.

src/logics/LRALogic.cc
```diff
@@ -161,4 +163,58 @@ PTRef LRALogic::insertTerm(SymRef sym, vec<PTRef> &&terms) {
...
+    opensmt::Number constantVal = getNumConst(constant);
+    assert(varFactors.size() > 0);
+    termSort(varFactors);
+    PTRef leadingFactor = varFactors[0];
+    // normalize the sum according to the leading factor
+    PTRef var, coeff;
+    splitTermToVarAndConst(leadingFactor, var, coeff);
+    opensmt::Number normalizationCoeff = abs(getNumConst(coeff));
+    // varFactors come from a normalized sum, no need to call normalization code again
+    PTRef normalizedSum = varFactors.size() == 1 ? varFactors[0] : mkFun(get_sym_Num_PLUS(), std::move(varFactors));
```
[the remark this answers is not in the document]
aehyvari: (see the related comment above for `LIALogic`)

src/logics/LIALogic.cc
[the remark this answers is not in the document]
blishko: Yes, the comment is obsolete, I'll remove it.

> On line 210 above we assume that the sum is normalised, which should indeed be the case with the current code. I wonder would it make sense to write an assert for that?

Yes, but how should it look like? Just `assert(normalizedSum == mkNumPlus(varFactors)`? Or do you have something else in mind?

src/logics/LRALogic.cc
```diff
@@ -161,4 +163,58 @@ PTRef LRALogic::insertTerm(SymRef sym, vec<PTRef> &&terms) {
...
     return LALogic::insertTerm(sym, std::move(terms));
+}
+
+/**
+ * Normalizes a sum term a1x1 + a2xn + ... + anxn + c such that the leading coefficient is either 1 or -1.
+ * Additionally, the normalized term is separated to constant and non-constant part, and the constant is modified as if
+ * it was placed on the other side of an equality.
+ *
+ * @param sum
+ * @return Constant part of the normalized sum as LHS and non-constant part of the normalized sum as RHS
```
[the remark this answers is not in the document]
blishko: You are right. I think I originally wanted to have a pair of PTRefs and then realized than in LIA I need the Rational value rather than PTRef. It seems to be fine to unify it in the way you suggest.

src/logics/LIALogic.cc
```diff
@@ -116,10 +116,21 @@ LIALogic::LIALogic() :
...
+ * Note that the constant part can be a non-integer number after normalization.
+ *
+ * @param sum
+ * @return Constant part of the normalized sum as LHS and non-constant part of the normalized sum as RHS
+ */
+opensmt::pair<FastRational, PTRef> LIALogic::sumToNormalizedPair(PTRef sum) {
+    assert(isNumPlus(sum));
     vec<PTRef> varFactors;
     PTRef constant = PTRef_Undef;
     Pterm const & s = getPterm(sum);
```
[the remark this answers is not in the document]
blishko: Fixed.

src/logics/LRALogic.cc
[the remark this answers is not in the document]
blishko: Yeah. the idea is to avoid normalization process when we know it's not needed.

src/logics/LRALogic.cc
```diff
@@ -161,4 +163,58 @@ PTRef LRALogic::insertTerm(SymRef sym, vec<PTRef> &&terms) {
...
+ */
+opensmt::pair<PTRef, PTRef> LRALogic::sumToNormalizedPair(PTRef sum) {
+    assert(isNumPlus(sum));
+    vec<PTRef> varFactors;
+    PTRef constant = PTRef_Undef;
+    Pterm const & s = getPterm(sum);
+    for (int i = 0; i < s.size(); i++) {
```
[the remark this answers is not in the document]
blishko: Fixed.

src/logics/LIALogic.cc
blishko: Fixed.

src/logics/LIALogic.cc
blishko: Fixed.

blishko:
I think I addressed all the comments. There is the question of adding an assertion for normalized sum.

> I tried out a refactoring where I put to a function std::pair<FastRational, vec<PTRef>>LALogic::getConstantAndFactors(PTRef) the common part of LIALogic::sumToNormalizedPair and LRALogic::sumToNormalizedPair. If you want I can push it here.

Yes, there seems to be some common code. You can add it here. Or you can create a separate branch for that and create a PR to my PR :)

src/logics/LIALogic.cc
```diff
@@ -188,9 +175,10 @@ opensmt::pair<FastRational, PTRef> LIALogic::sumToNormalizedPair(PTRef sum) {
     assert(std::all_of(coeffs.begin(), coeffs.end(), [](opensmt::Number const & coeff) { return coeff.isInteger(); }));
     // Now make sure all coeffs are coprime
     auto coeffs_gcd = abs(coeffs[0]);
-    for (std::size_t i = 1; i < coeffs.size() && coeffs_gcd != 1; ++i) {
-        coeffs_gcd = gcd(coeffs_gcd, abs(coeffs[i]));
+    for (auto const & coeff : coeffs) {
+        coeffs_gcd = gcd(coeffs_gcd, abs(coeff));
```
blishko, line 179: This now adds one unnecessary `gcd` operation of the first coefficient with itself.
I think I would rather stick to the original version.

aehyvari: Right, it slipped past me, I didn't want to commit it.  I'll take it off.

blishko: OK. Otherwise the refactoring is nice!

src/logics/LRALogic.cc
```diff
@@ -173,25 +173,11 @@ PTRef LRALogic::insertTerm(SymRef sym, vec<PTRef> &&terms) {
...
-            assert(isLinearFactor(arg));
-            varFactors.push(arg);
-        }
-    }
 
-    if (constant == PTRef_Undef) { constant = getTerm_NumZero(); }
-    opensmt::Number constantVal = getNumConst(constant);
-    assert(varFactors.size() > 0);
-    termSort(varFactors);
+opensmt::pair<FastRational, PTRef> LRALogic::sumToNormalizedPair(PTRef sum) {
```
aehyvari, line 177: I'm going to change this into 

`opensmt::pair<opensmt::Number, PTRef> LRALogic::sumToNormalizedPair(PTRef sum) {`

blishko: OK!

blishko:
Do you want to add assertions to the normalized sums? Or can we merge the current state?

aehyvari (approved):

aehyvari:
> Do you want to add assertions to the normalized sums? Or can we merge the current state?

I think we can merge the current state.

Issue #296, opened 2021-09-24
get-model uses real (rational) literal syntax "1/2" instead of "(/ 1 2)"


jwaldmann:
This is a recent change, and it is confusing my parser.

```
$ echo "(set-logic QF_LRA)(declare-fun X () Real)(assert (= X (/ 1 2))) (check-sat) (get-model)" | ./build/opensmt -p
sat
(
  (define-fun X () Real
    1/2)
)
```
I was expecting what some previous version printed 
```
$ echo "(set-logic QF_LRA)(declare-fun X () Real)(assert (= X (/ 1 2))) (check-sat) (get-model)" | opensmt -p
sat
(model
(define-fun X () Real (/ 1 2))
)
```

jwaldmann:
Possibly related: with `get-value`, the output format also changed, but in the other direction:

current: 
```
$ echo "(set-logic QF_LRA)(declare-fun X () Real)(assert (= X (/ 1 2))) (check-sat) (get-value (X))" | ./build/opensmt -p
sat
((X (/ 1 2)))
```
earlier:
```
$ echo "(set-logic QF_LRA)(declare-fun X () Real)(assert (= X (/ 1 2))) (check-sat) (get-value (X))" | opensmt -p
sat
((X 1/2))
```

jwaldmann:
I am trying to trace/debug this. Running the examples with `gdb` I found 
* `get-model` calls `Logic:pp`, produces `1/2`,
* `get-value` calls `logic.printTerm`, producing `(/ 1 2)`.

blishko:
Hi @jwaldmann , thanks for noticing this. It seems the model validator is fine with this, but I agree that the SMT-LIB syntax is more appropriate here.

blishko:
Resolved by #297

PR #297, opened 2021-09-24
for #296

jwaldmann:
as explained in the issue.


blishko:
@aehyvari , what do you think about this? Was there any specific reason to use `Logic::pp` there before? The SMT-LIB format of `Logic::printTerm` seems more appropriate here.

aehyvari:
To me this looks like the correct fix!

blishko (approved):

PR #298, opened 2021-09-28
Model fix

aehyvari:
A recent fix on model printing broke UF models so that sort information was no longer printed for models.  This PR enables back their printing.

While preparing the fix I noticed a regression test where (as far as I understand) the sort was incorrectly omitted for UF models when queried with `get-value`.  This is now fixed.

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -363,9 +363,8 @@ Logic::printTerm_(PTRef tr, bool ext, bool safe) const
         bool first = true;
         for (auto arg : t) {
             char *current_term_p = printTerm_(arg, ext, safe);
-            std::string current_term = current_term_p;
+            ss << (first ? "" : " ") << current_term_p;
             free(current_term_p);
-            ss << (first ? "" : " ") << current_term;
             first = false;
         }
         ss << ")" << (ext ? " <" + std::to_string(tr.x) + ">" : "");
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -340,46 +340,37 @@ Logic::pp(PTRef tr) const
 char*
 Logic::printTerm_(PTRef tr, bool ext, bool safe) const
 {
-    char* out;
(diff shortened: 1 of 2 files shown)


src/logics/Logic.cc
```diff
@@ -340,46 +340,37 @@ Logic::pp(PTRef tr) const
...
+        return strdup(ss.str().data());
+    } else {
+        // Here we know that t.size() > 0
+        ss << "(" << name_escaped;
+        bool first = true;
+        for (auto arg : t) {
+            char *current_term_p = printTerm_(arg, ext, safe);
+            std::string current_term = current_term_p;
+            free(current_term_p);
+            ss << (first ? "" : " ") << current_term;
```
blishko, line 368: I think you don't need the temporary string here.
```suggestion
            ss << (first ? "" : " ") << current_term_p;
            free(current_term_p);
```

blishko (commented):
I think it looks good.

Two comments:
1. I would like to fix the model validation script to exit with non-zero status when it finds an invalid model. This problem should have been caught by our CI, but currently the script seems to return with exit code 0 even if it detects invalid model.
2. This is maybe just a note for the future, but seeing how the strings get copied here, I was thinking that this whole recursive creation of the string representation could be done using one stream which would be passed around. I think it would be more efficient.

aehyvari:
> I would like to fix the model validation script to exit with non-zero status when it finds an invalid model. This problem should have been caught by our CI, but currently the script seems to return with exit code 0 even if it detects invalid model.

Yes, I have it in an upcoming PR.

> This is maybe just a note for the future, but seeing how the strings get copied here, I was thinking that this whole recursive creation of the string representation could be done using one stream which would be passed around. I think it would be more efficient.

Probably that's true.  I seem to remember that there used to be some problems with passing `std::stringstream` around on some older compilers, but didn't try it here.  I think we can do that cleanup later.

blishko (approved):

PR #299, opened 2021-09-28
CI: fix model validation to actually report errors

aehyvari:
This PR fixes model validator CI scripts so that they return with non-zero value if model is not validated.  It also fixes a bug that resulted in temporary files being deleted even when a flag was given that was supposed to prevent the deletion for easier debugging.

### DIFF
--- a/regression_models/bin/run-and-validate.sh
+++ b/regression_models/bin/run-and-validate.sh
@@ -107,7 +107,7 @@ if [[ $(grep '^sat' ${tmpout}) ]]; then
         exit 1
     fi
 else
-    echo "Not satsfiable: $1"
+    echo "Not satisfiable: $1"
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


regression_models/bin/run-and-validate.sh
```diff
@@ -98,17 +102,17 @@ if [[ $(grep '^sat' ${tmpout}) ]]; then
...
             > ${output}
-    grep "model_validator_status=VALID" ${output} >/dev/null || \
-        (echo "Invalid model for $1"; exit 1)
+    if (! $(grep "model_validator_status=VALID" ${output} >/dev/null)); then
+        echo "Invalid model: $1";
+        exit 1
+    fi
 else
-    echo "not sat"
+    echo "Not satsfiable: $1"
```
blishko, line 110:
```suggestion
    echo "Not satisfiable: $1"
```

blishko (commented):
I think it is good.
But why it was not working before? The `exit 1` was there, no? Was it in some subshell and it did not propagate to the top-level?

aehyvari:
> But why it was not working before? The `exit 1` was there, no? Was it in some subshell and it did not propagate to the top-level?

Yes, it was just exiting from the sub shell opened by the `()`.  With the fix, it exists with non-zero value from the actual shell.  Apparently I should have written regression tests for our regression tests.  And then regression tests for those...

blishko (approved):

PR #300, opened 2021-09-29
CI: regression test for correct printing of rational numbers in models

aehyvari:
A syntax test for rational representation.  The model validator accepts a more relaxed syntax but for easier parsing we need to conform to the official standard.  Hopefully prevents #296 from showing up again.


blishko (approved):
Should we add a test also for the minus operator? So that we produce `(- 1)` and not `-1`?

aehyvari:
> Should we add a test also for the minus operator? So that we produce `(- 1)` and not `-1`?

That's a good idea.  I added the tests.

blishko (approved):

PR #301, opened 2021-09-29
Fix bound store not being cleared up properly

blishko:
Previously, `LABoundStore` was not clearing one of its data structure.
This could result in garbage being remembered between consecutive `check-sat` calls.


aehyvari (approved):
The bug was introduced relatively recently.  The version of opensmt used in smt-comp for instance doesn't seem to be affected.

PR #302, opened 2021-09-29
LA: Cleanup of a helper class LAExpression

blishko:
Includes formatting changes, some code modernisation and removal of unused methods.

### DIFF
--- a/src/simplifiers/LA.cc
+++ b/src/simplifiers/LA.cc
@@ -30,15 +30,10 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 void LAExpression::initialize(PTRef e, bool do_canonize) {
     assert(logic.isNumEq(e) || logic.isNumLeq(e));
 
-    vector<PTRef> curr_term;
-    vector<opensmt::Real> curr_const;
-
     PTRef lhs = logic.getPterm(e)[0];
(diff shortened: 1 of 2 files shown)


blishko:
I recommend to hide whitespace differences when looking at the diff.

src/simplifiers/LA.cc
```diff
@@ -151,43 +143,40 @@ void LAExpression::print(ostream & os) const {
...
             else
-                os << " (* " << term.second << " " << logic.printTerm(term.first) << ")";
+                os << " (* " << factor << " " << logic.printTerm(var) << ")";
         }
         os << ")";
     }
     if (r == OP::EQ || r == OP::LEQ)
         os << " " << constant << ")";
-    os << '\n';
+    os << std::endl;
```
blishko, line 156: This is the only thing I would probably not do.
If you are passed a stream, it does not feel like the responsibility of this function to flush the stream.
Maybe the caller does not want that. I would keep `\n` here.

blishko (commented):
Nice additional cleanup!

src/simplifiers/LA.cc
aehyvari: Ok, I'll take it off!

aehyvari (approved):

aehyvari:
> Nice additional cleanup!

Thanks!  Approved.

PR #304, opened 2021-09-30
Unit test for correct printing of terms

aehyvari:
This PR provides a unit test that catches the problem fixed in PR #303.


blishko (approved):
LGTM.

I am not sure why it is named "SplitTest", though. It is testing term printing, so I would expect something related to that.

aehyvari:
> I am not sure why it is named "SplitTest", though. It is testing term printing, so I would expect something related to that.

You're right of course, but I was thinking of adding more splitting-related tests here in the future.  Somehow the unit test is more related to the eventual PR on parallel solving, but since the bug was fixed in master it makes sense to have the unit test in master as well.

PR #308, opened 2021-10-05
Fix internal error on top-level distinct in incremental mode

### LINKED ISSUE #307: Internal error: Negated distinct asserted to Egraph
Exception is thrown on the following example:
```
(set-logic QF_UF)
(declare-sort U 0)
(declare-fun x0 () U)
(declare-fun x1 () U)
(declare-fun x2 () U)

(push 1)
(assert (distinct x0 x1 x2))
(check-sat)
(pop 1)
(push 1)
(assert (= x0 x1))
(check-sat)
(exit)
```

blishko:
Having `distinct` literal from popped frames in the solver when new information comes in that implies the negation of the distinct could cause a problem in Egraph as it cannot deal with distinct literals asserted negatively.

The proposed fix keeps the top-level distincts only in the bottom frame. This prevents the problem described above.

Fixes #307.


aehyvari (approved):
Good!

PR #314, opened 2021-10-28
Simplify Egraph data structures

blishko:
The aim of this PR is to simplify data structures used in Egraph and document the invariants they maintain.

The representation of `Enode` changes. Instead of using three different types of `Enodes` (Symbol, List, Term) and using the `car/cdr` representation, the new version creates only `Enodes` for terms.
The internal representation closely mirrors our `Pterm` representation.
The children are stored at the end of `Enode` object, as now Enode can have different number of children.
Moreover, `Enode` also remembers indices in its children's use-vectors. These are stored right after the children.
The `EnodeAllocator` is responsible for allocating enough space and nothing else can create new `Enodes`.

The main data-structures, signature table and use-vectors, are adjusted to deal with the new representation.
This is reflected in the algorithm in the steps just before and after `merge`, in order to maintain the invariants of the signature table and use-vectors.


src/tsolvers/egraph/Enode.h
```diff
@@ -48,8 +47,32 @@ struct ERef {
...
+    static UseVectorIndex DuplicateIndex;
+};
+
+
+
+static_assert(sizeof(ERef) == sizeof(UseVectorIndex));
+
 inline void swap(ERef & y, ERef & z) { ERef tmp = y; y = z; z = tmp; }
 
+class ERefSpan {
```
aehyvari, line 64: Let's have the generic span, i.e., `template<typename T> Span {...}` with `T const * begin() const` and `T const * end() const`?

src/tsolvers/egraph/Enode.h
```diff
@@ -59,154 +82,86 @@ struct ELRef {
...
+    /*
+     * Pointer to the end of object, where ENodeAllocator allocated enough space for Enode's children and
+     * their use-vector indices.
+     */
+    ERef args[0];
 
     friend class EnodeAllocator;
-    friend class EnodeStore;
+    Enode(SymRef symbol, ERefSpan children, ERef myRef, PTRef ptr);
+    Enode(Enode const&) = delete;
```
aehyvari, line 122: Apparently [deleted members should be public](https://stackoverflow.com/questions/18931133/must-a-deleted-constructor-be-private)

src/tsolvers/egraph/Enode.h
```diff
@@ -59,154 +82,86 @@ struct ELRef {
...
+
+    // Term representation
+    SymRef symb;
+    uint32_t argSize;
+
+    /*
+     * Pointer to the end of object, where ENodeAllocator allocated enough space for Enode's children and
+     * their use-vector indices.
+     */
+    ERef args[0];
```
aehyvari, line 118: I wonder if it would make sense to interleave these?  It feels like some of the update algorithms could be written simpler that way, but I didn't try myself.  What do you think?

src/tsolvers/egraph/Enode.h
```diff
@@ -59,154 +82,86 @@ struct ELRef {
...
+    void  setForbid     (ELRef r)       { forbid = r; }
+    void  setDistClasses( const dist_t& d) { dist_classes = d; }
+    dist_t getDistClasses() const { return dist_classes; }
+
+    uint32_t getSize() const { return argSize; }
+    SymRef getSymbol() const { return symb; }
+    ERef operator[](std::size_t i) const { return *(args + i); }
+    ERef const * begin() const { return args; }
+    ERef const * end() const { return args + argSize; }
+    UseVectorIndex getIndex(uint32_t i) { return UseVectorIndex{static_cast<int32_t>((args + argSize + i)->x)}; }
```
aehyvari, line 156: If we make UseVectorIndex::x an `uint32_t` there's no need for the `static_cast`

src/tsolvers/egraph/Enode.h
```diff
@@ -59,154 +82,86 @@ struct ELRef {
...
+    void setIndex(uint32_t i, UseVectorIndex index) { (args + argSize + i)->x = static_cast<uint32_t>(index.x); }
```
aehyvari, line 157: Same here: If we make UseVectorIndex::x an uint32_t there's no need for the static_cast

src/tsolvers/egraph/Enode.h
```diff
@@ -215,42 +170,28 @@ class EnodeAllocator : public RegionAllocator<uint32_t>
...
         // MB: The data structures used in the satisfiability checking algorithm (UseVector) requires at the moment that
         // the values of ERef cannot exceed 2^30. The benchmarks that we are dealing with at the moment are far below this limit,
         // but here is a dynamic check just in case.
         if (v >= (static_cast<uint32_t>(-1) >> 2)) { throw OutOfMemoryException(); }
-        ERef eid;
-        eid.x = v;
-        assert(t != Enode::et_list || ptr == PTRef_Undef);
-        new (lea(eid)) Enode(car, cdr, *this, eid, n_enodes++, ptr);
-        return eid;
+        ERef eref;
```
aehyvari, line 190: Would `ERef eref{v};` be more readable?

src/tsolvers/egraph/Enode.h
```diff
@@ -59,154 +82,86 @@ struct ELRef {
...
-
-struct ERef_vecEq {
-    bool operator () (const vec<ERef>& s1, const vec<ERef>& s2) const {
-        if (s1.size() != s2.size()) return false;
-        for (int i = 0; i < s1.size(); i++)
-            if (s1[i] != s2[i]) return false;
-        return true; }
-};
-
 class EnodeAllocator : public RegionAllocator<uint32_t>
```
aehyvari, line 165: There's so much code duplication with these allocators.  At some point somebody should try to write them with a template.

src/tsolvers/egraph/Enode.h
```diff
@@ -266,13 +207,7 @@ class EnodeAllocator : public RegionAllocator<uint32_t>
     void free(ERef eid)
```
aehyvari, line 207: Above this, is the `const Enode & dbg` used and if it's not should we remove it?

aehyvari (commented):
So far looks great!  It'd be good to update the comment on line 70 - 72 of `Egraph.h`

I'm going to continue the pass once I have more time, now I need to go...

src/tsolvers/egraph/EnodeStore.h
```diff
@@ -35,31 +35,54 @@ class Logic;
...
+    struct SignatureHash {
+        EnodeAllocator const & ea;
+
+        SignatureHash(EnodeAllocator const & ea) : ea{ea} {}
+
+        std::size_t operator()(ERef ref) const {
+            Enode const & node = ea[ref];
+            std::size_t seed = node.getSymbol().x;
+            for(uint32_t i = 0; i < node.getSize(); ++i) {
+                seed ^= ea[node[i]].getRoot().x + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
aehyvari, line 48: How did you come up with this hash function?  Is it a concern that (I assume) most of the time the most significant bits of `x` will be zero?  I don't really have an alternate suggestion at the moment, but just asking out of curiosity and maybe documentation purposes.

src/tsolvers/egraph/EnodeStore.h
```diff
@@ -35,31 +35,54 @@ class Logic;
...
+    struct SignatureEqual {
+        EnodeAllocator const & ea;
+
+        SignatureEqual(EnodeAllocator const & ea) : ea{ea} {}
+
+        bool operator()(ERef a, ERef b) const {
+            Enode const & anode = ea[a];
+            Enode const & bnode = ea[b];
+            if ((anode.getSize() != bnode.getSize()) or anode.getSymbol() != bnode.getSymbol()) { return false; }
+            for(uint32_t i = 0; i < anode.getSize(); ++i) {
```
aehyvari, line 63: Space after `for`

src/tsolvers/egraph/EnodeStore.h
```diff
@@ -113,59 +135,29 @@ class EnodeStore {
...
     }
 
-    ERef addAnonSymb(PTRef tr);
-
-// DEBUG
-#ifdef PEDANTIC_DEBUG
-    bool checkInvariants();
-#endif
-
     char* printEnode(ERef);
```
aehyvari, line 158: This is no more.

aehyvari (commented):
Here's my last batch of comments!

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1829,122 +1601,89 @@ uint32_t UseVector::getFreeSlotIndex() {
...
+ * @param parent parent enode
+ * @param childIndex index in the list of children of child under consideration
+ */
+bool Egraph::childDuplicatesClass(ERef parent, uint32_t childIndex) {
+    assert(childIndex < getEnode(parent).getSize());
+    if (childIndex == 0) { return false; }
+    Enode const & parentNode = getEnode(parent);
+    ERef child = parentNode[childIndex];
+    auto childClass = getEnode(getEnode(child).getRoot()).getCid();
+    for (uint32_t i = 0; i < childIndex; ++i) {
```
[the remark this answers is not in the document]
blishko: Maybe yes, but I don't really miss it here.

src/tsolvers/egraph/EnodeStore.h
blishko: Removed

src/tsolvers/egraph/EnodeStore.h
blishko: Fixed

src/tsolvers/egraph/EnodeStore.h
```diff
@@ -35,31 +35,54 @@ class Logic;
...
+
+    struct SignatureHash {
+        EnodeAllocator const & ea;
+
+        SignatureHash(EnodeAllocator const & ea) : ea{ea} {}
+
+        std::size_t operator()(ERef ref) const {
+            Enode const & node = ea[ref];
+            std::size_t seed = node.getSymbol().x;
+            for(uint32_t i = 0; i < node.getSize(); ++i) {
```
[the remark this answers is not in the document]
blishko: Fixed.

src/tsolvers/egraph/Enode.h
blishko: Fixed.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1829,122 +1601,89 @@ uint32_t UseVector::getFreeSlotIndex() {
...
-    assert(cdrParentIndex >= 0);
-    assert(UseVector::entryToERef(parents[getEnode(cdr.getRoot()).getCid()][cdrParentIndex]) == parent);
-    parents[getEnode(cdr.getRoot()).getCid()].clearEntryAt(cdrParentIndex);
+/**
+ * Removes the parent enode from the use-vectors of its children except the one with the given congruence id.
+ *
+ * Same as removeFromUseVectors except we keep the parent entry in the use-vector with the given congrunce id.
+ */
+void Egraph::removeFromUseVectorsExcept(ERef parent, CgId cgid) {
+    Enode & parentNode = getEnode(parent);
```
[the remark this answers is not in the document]
blishko: Fixed.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1829,122 +1601,89 @@ uint32_t UseVector::getFreeSlotIndex() {
...
-    assert(UseVector::entryToERef(parents[getEnode(cdr.getRoot()).getCid()][cdrParentIndex]) == parent);
-    parents[getEnode(cdr.getRoot()).getCid()].clearEntryAt(cdrParentIndex);
+/**
+ * Removes the parent enode from the use-vectors of its children.
+ *
+ * Note that the indices in the parent are not reset, they remain invalid.
+ *
+ */
+void Egraph::removeFromUseVectors(ERef parent) {
+    Enode & parentNode = getEnode(parent);
```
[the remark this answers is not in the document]
blishko: Fixed.

src/tsolvers/egraph/Enode.h
blishko: Fixed.

src/tsolvers/egraph/Enode.h
```diff
@@ -48,8 +47,32 @@ struct ERef {
 };
 static struct ERef ERef_Undef = {INT32_MAX};
 
+struct UseVectorIndex {
+    int32_t x;
```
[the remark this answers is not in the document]
blishko: Changed.

src/tsolvers/egraph/Enode.cc
```diff
@@ -27,80 +27,26 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
-Enode::Enode(ERef car_,
-             ERef cdr_,
-             EnodeAllocator& ea,
-             ERef er_,
-             enodeid_t id_,
-             PTRef ptr)
-     : header({et_list, id_}) // Fix the type to term later if necessary
-     , er(er_)
-     , cid(cgid_ctr++)
+UseVectorIndex UseVectorIndex::NoIndex = {-1};
```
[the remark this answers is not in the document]
blishko: Changed.
Since we do not really need to distinguish between `NoIndex` and `DuplicateIndex` I just combined them to one `NotValidIndex`.

src/tsolvers/egraph/Enode.h
blishko: The `static_cast`s have been removed.

src/tsolvers/egraph/Enode.h
blishko: See above.

src/tsolvers/egraph/EnodeStore.cc
```diff
@@ -169,12 +115,10 @@ bool EnodeStore::needsEnode(PTRef tr) const {
  * @return the vector of terms as erefs
  */
 vec<ERef> EnodeStore::getArgTermsAsVector(ERef er) const {
```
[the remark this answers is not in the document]
blishko: Removed.

src/tsolvers/egraph/Enode.h
```diff
@@ -48,8 +47,32 @@ struct ERef {
 };
 static struct ERef ERef_Undef = {INT32_MAX};
```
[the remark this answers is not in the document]
blishko: Yeah, it's `ERef_Undef` and also `ELRef_Undef`.
I think it should be safe to change them to `UINT32_MAX`. I can do it here.

src/tsolvers/egraph/Enode.h
blishko: I wanted to say that I would like to keep them separate in order to have an ability to iterate over the children. Then I realized that the only place I iterate over the children is now instead of the removed `EnodeStore::getArgTermsAsVector` method.
So maybe it could be done, I was considering something like `<ERef,UseVectorIndex>` pairs.
But I don't know, it does not look very nice. 
Where do you think the update algorithms could be improved?

I was also thinking about this *zero-lenght* array pattern.
As far as I understood, this is GNU extension, which also happens to be supported by clang. But we would probably have problems if we ever want to support Windows platform as well.
I believe the compilers even warn about this with `-Wpedantic`.
But I have not found a good alternative yet.

src/tsolvers/egraph/Enode.h
blishko: It does not seem to be. I am removing it.

src/tsolvers/egraph/Enode.h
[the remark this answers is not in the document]
blishko: Changed.

src/tsolvers/egraph/Enode.h
blishko: Removed.

src/tsolvers/egraph/Enode.cc
```diff
@@ -27,80 +27,26 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+UseVectorIndex UseVectorIndex::DuplicateIndex = {-2};
 
+Enode::Enode(SymRef symbol, ERefSpan children, ERef myRef, PTRef term)
+     : cid(cgid_ctr++), pterm(term), symb(symbol), argSize(children.getSize())
```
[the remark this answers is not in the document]
blishko: Done.

src/tsolvers/egraph/Enode.h
blishko: I have generalized this span.

src/tsolvers/egraph/EgraphSolver.cc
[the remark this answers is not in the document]
aehyvari: Like this:
```
    opensmt::span<ERef const> sp(parentNode.begin(), childIndex);
    return std::any_of(std::begin(sp), std::end(sp),
                [&](ERef c) {
        return childClass == getEnode(getEnode(c).getRoot()).getCid();
    });
```
or if that's too adventurous, then like this:
```
    for (auto precChild : ERefSpan(parentNode.begin(), childIndex)) {
        auto otherClass = getEnode(getEnode(precChild).getRoot()).getCid();
        if (otherClass == childClass) { return true; }
    }
    return false;
```

src/tsolvers/egraph/Enode.h
aehyvari: On the zero-length array situation I wouldn't know...  I don't see the world going towards running smt solvers in windows right now, but obviously a more standard solution would be great.  Let's keep it in mind?

On the interleaving, I'm still not sure (= haven't tried) but it feels like `void Egraph::addToUseVectors(ERef parentRef)` could be implemented nicely using this.

Even if we had to go through all children in an interleaved representation, this could be done somewhat elegantly with the `for (auto [eref, _] : enode)` construct.

aehyvari: Anyway we might leave this for another PR if you want and think it's a good idea.

src/tsolvers/egraph/EnodeStore.h
aehyvari: Ok thanks!

aehyvari (commented):
Here's my last comments.  I think it looks pretty good, but maybe have a look if there's something useful.

src/tsolvers/egraph/Enode.h
aehyvari: Yes, I like it 🥇

src/tsolvers/egraph/EgraphSolver.cc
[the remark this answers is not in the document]
blishko: I like the second version and after some consideration I started liking even the first version so I am going with that.

src/tsolvers/egraph/Enode.cc
```diff
@@ -27,80 +27,26 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+
+Enode::Enode(SymRef symbol, opensmt::span<ERef> children, ERef myRef, PTRef term) :
+    root(myRef),
+    cid(cgid_ctr++),
+    eq_next(myRef),
+    eq_size(1),
+    pterm(term),
+    forbid(ELRef_Undef),
+    dist_classes(0),
+    exp_reason(PtAsgn(PTRef_Undef, l_Undef)),
```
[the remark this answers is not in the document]
blishko: Fixed.

src/tsolvers/egraph/EnodeStore.h
blishko: Added comment in the code.

src/tsolvers/egraph/Enode.h
blishko: Yeah, maybe it is possible.
But let's try it in another PR.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -336,16 +323,12 @@ void Egraph::declareTerm(PTRef tr) {
...
-#endif
-
     if (PTRefERefPairVec.size() == 0) {
         return;
     }
 
     for (auto PTRefERefPair : PTRefERefPairVec) {
-        updateParentsVector(PTRefERefPair.tr);
+        updateUseVectors(PTRefERefPair.tr);
     }
```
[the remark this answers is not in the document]
blishko: I like this. Updated.

blishko:
I think we resolved everything except the possible interleaving of children and indices in Enode.
Let's leave that to separate PR, to see if it makes sense.
I am going to run SMT-LIB benchmarks again to check that nothing broke.

src/tsolvers/egraph/EgraphSolver.cc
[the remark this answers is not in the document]
blishko: Unfortunately, GCC 7.3.1 from our starexec CI considers `_` as unused variable and issues a warning. I will have to explicitly mark it to silence the warning.

blishko:
The results on SMT-LIB benchmarks are the same as the one I shared before your review, so I think this should be good to go.

aehyvari (approved):
Great!

Issue #315, opened 2021-10-28
Make use of ReadLine/EditLine optional


blishko:
It would be better to have a default build of OpenSMT without the dependency on ReadLine or EditLine, but still allow people to opt-in if they want to use this feature.

PR #318, opened 2021-11-03
Make compilation with line editing libraries optional

aehyvari:
Most users will probably be able to live without the possibility of scrolling interactively the history of smt-lib script commands while interacting with the solver.  Having libraries that enable this (`libedit` or `libreadline`) as dependencies causes occasional compilation problems in systems where the libraries are difficult to install.  This PR makes the compilation of line editing features optional, and the default is not to have them.  The continuous integration will still build one with `libedit` to maintain the option for the users who might benefit from it.

Addresses  #315


src/bin/opensmt.cc
```diff
@@ -155,7 +154,12 @@ int main( int argc, char * argv[] )
     return 0;
 }
 
-void interpretInteractive(Interpret & interpret, FILE* input) {
+#ifndef ENABLE_LINE_EDITING
+void interpretInteractive(Interpret & interpret) {
+    interpret.interpPipe();
+}
+#else
+void interpretInteractive(Interpret & interpret) {
```
blishko, line 162: Nice! So the `FILE*` parameter was not used at all?
This reminds me that we should try to enable more compiler warnings such as detection of unused parameters and variables :)

blishko (commented):
Looks good!
Have you tested this manually? I think all our tests took file as input so this feature is not really tested, no? Or would this be covered by our `regression_pipe`?

Also, please update the main `README` regarding the line editing libraries.

src/bin/opensmt.cc
aehyvari: Yes, looks like it was never implemented.  Maybe not needed.

blishko:
Adding a changelog entry is also a good idea, but I meant that we should also update our [documentation README](https://github.com/usi-verification-and-security/opensmt/blob/master/README.md).

aehyvari:
I've tested it only manually.  I don't think it's worth to try to write tests for the use of the line editing libraries.  They've been very stable anyway.  Famous last words.  The pipe mode tests hopefully catch their problems, even if they don't use the code.

I updated both `changelog` and `readme` now.

blishko (approved):
LGTM!

PR #320, opened 2021-11-03
Do not include headers of line editing library when disabled

blishko:
It looks like in #318 we forget to hide include of headers behind `ifdef`.
I got compilation error on the cluster.
It was most likely not detected because the headers could be found in the standard system paths, but that's not the case in the cluster.


aehyvari (approved):
True. Good that you tried it there!

PR #322, opened 2021-11-04
Cleanup after changes in data structures of Egraph

blishko:
After the recent changes in Egraph, some types and inclusions became unnecessary.
Especially, the inclusion of `Global.h` into `SSort.h` is really unnecessary and propagates all the dependencies from `Global.h` into almost every other file in the project.


aehyvari (commented):
This looks good!  I wonder if we should split Global.h further.  An easy(?) start would be the loads of interpolation-specific stuff that could probably go to a separate file.

blishko:
Yes, I think it can be split further.
At least the interpolation stuff could be moved, as you suggest.

aehyvari:
Ok, let's do it later.

aehyvari (approved):

PR #323, opened 2021-11-04
Enode cleanup

aehyvari:
This PR suggests mostly minor cleanups in the egraph system.  I tried to organise the commits thematically so that they're easier to check.


src/tsolvers/egraph/EnodeStore.cc
```diff
@@ -150,9 +150,7 @@ vec<PTRefERefPair> EnodeStore::constructTerm(PTRef tr) {
...
 
 bool EnodeStore::needsRecursiveDefinition(PTRef tr) const {
-    bool recdef = true;
-    for (auto ch : logic.getPterm(tr)) {
-        recdef &= needsEnode(ch);
-    }
-    return recdef;
+    Pterm const & t = logic.getPterm(tr);
+    opensmt::span const sp(t.begin(), t.size());
+    return std::all_of(sp.begin(), sp.end(), [&](PTRef ch) { return needsEnode(ch); });
```
blishko, line 155:
```suggestion
    return std::all_of(sp.begin(), sp.end(), [this](PTRef ch) { return needsEnode(ch); });
```
I like restricting the lambda's capture, if it is not too wordy. But I am also OK with your version.

src/tsolvers/egraph/EnodeStore.cc
```diff
@@ -150,9 +150,7 @@ vec<PTRefERefPair> EnodeStore::constructTerm(PTRef tr) {
 }
 
 bool EnodeStore::needsRecursiveDefinition(PTRef tr) const {
-    bool recdef = true;
-    for (auto ch : logic.getPterm(tr)) {
-        recdef &= needsEnode(ch);
-    }
-    return recdef;
+    Pterm const & t = logic.getPterm(tr);
+    opensmt::span const sp(t.begin(), t.size());
```
blishko, line 154: Do we actually need the span here? We could use `t.begin()` and `t.end()` directly in `std::all_of`, no?

src/tsolvers/egraph/Explainer.cc
```diff
@@ -48,10 +48,10 @@ void Explainer::storeExplanation(ERef x, ERef y, PtAsgn reason)
     // balanced (which is a requirement to keep the O(nlogn) bound
 
     // Make sure that x is the node with the larger number of edges to switch
-    const Enode& root_x = getEnode(getEnode(x).getRoot());
-    const Enode& root_y = getEnode(getEnode(y).getRoot());
+    Enode const & root_x = getEnode(getEnode(x).getRoot());
+    Enode const & root_y = getEnode(getEnode(y).getRoot());
     if ( root_x.getSize() < root_y.getSize() ) {
-        swap(x,y);
+        std::swap(x,y);
```
blishko, line 54: Seeing this now, I got a little worried. I thing we should not create the named references `root_x` and `root_y` and just inline them in the `if` check.
Because if the `swap` does occur, suddenly `root_x` is the root of `y` and `root_y` is the root of `x`, which is super confusing.
Hopefully we are not using these references later...

blishko: Also, this made me realize that I have inadvertently changed the meaning of this code in #314. Because `getSize()` now returns the number of children and this should now be `getEqSize()`.

blishko (commented):
You changes uncovered an unwanted change I made in #314. We need to take a closer look, but I have to go now and will return to this later.

src/tsolvers/egraph/Enode.h
```diff
@@ -33,6 +33,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "PtStructs.h"
 #include "SymRef.h"
 #include "TypeUtils.h"
+#include "Global.h"
```
blishko, line 36: Hopefully we do not need this include?

src/tsolvers/egraph/EnodeStore.cc
aehyvari: Fixed.

src/tsolvers/egraph/Explainer.cc
aehyvari: > Hopefully we are not using these references later...

They're not used afterwards.  I could make a scope just for this to make it clearer?

> Also, this made me realize that I have inadvertently changed the meaning of this code in #314. Because getSize() now returns the number of children and this should now be getEqSize().

I'll fix it in a follow-up commit.  If I understood you correctly, here it's enough to change `getSize()` => `getEqSize()`?

src/tsolvers/egraph/EnodeStore.cc
aehyvari: For me it's the same.  We can go with your suggestion.  I think the three extra characters will not cause major problems with modern disk capacity.

blishko (commented):
I think this looks good!

src/tsolvers/egraph/Enode.h
blishko: This is needed because of the `SET_BIT` thing, I guess. Could we avoid this somehow?

src/tsolvers/egraph/Explainer.cc
blishko: Yes, but I would like to measure the impact of this and also check that the other usages are correct.

blishko: > They're not used afterwards. I could make a scope just for this to make it clearer?

Yes, scope is also fine.

src/tsolvers/egraph/Enode.h
aehyvari: Fixed.

blishko (approved):
I think this is all good now!

PR #324, opened 2021-11-04
Theory propagation is always on, remove code for disabling it

aehyvari:


blishko (approved):
OK! Hopefully we will not need to disable it again in the future :)

PR #327, opened 2021-11-09
[logic] Moved LALogic::isNegated to LAVarMapper::isNegated

aehyvari:
`isNegated` implementation seems to fit better in the theory solver.  This PR suggests it to be implemented in `LAVarMapper` that holds the mapping between `PTRef`s and `LAVar`s and uses `isNegated` in an `assert`.

### DIFF
--- a/src/logics/LALogic.cc
+++ b/src/logics/LALogic.cc
@@ -10,28 +10,6 @@
 
 #include <memory>
 
-bool LALogic::isNegated(PTRef tr) const {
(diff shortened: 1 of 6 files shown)


blishko (approved):
This could also be a free function, since it does not really depend on `LaVarMapper`.
It would take `LALogic` as an additional parameter.
But I am also fine with this version.

aehyvari:
Ok, let's try with this and if there's need to use it in weird places we can always make it independent.

PR #330, opened 2021-11-10
Minor cleanup for terms and LA bounds

aehyvari:


src/symbols/Symbol.h
```diff
@@ -66,6 +66,16 @@ class Symbol {
...
     SRef     operator [] (int i) const   { return args[i+1].sort; }
+    /**
+     * @note The function is unsafe: if used in a loop, the loop should in *absolutely no case* build new terms in the same Pterm allocator
+     * @return A pointer to the first child of the term
+     */
+    SRef const * begin   ()      const   { return (SRef*)(args + 1); }
+    /**
+     * @note The function is unsafe: if used in a loop, the loop should in *absolutely no case* build new terms in the same Pterm allocator
+     * @return A pointer to the last child of the term
+     */
```
blishko, line 77: Isn't the comment a bit imprecise? Here it depends on creating new symbols, no? Creating new terms is safe if we don't create new symbols, no?

aehyvari: Right, that's a copy-paste error :)

blishko (approved):

PR #336, opened 2021-11-12
[logic] Type checking for mkFun when in debug mode

aehyvari:


src/logics/Logic.cc
```diff
@@ -1593,3 +1591,20 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+    if (symbol.left_assoc()) {
+
+    }
+    if (symbol.nargs() == args.size_()) {
+        for (unsigned int i = 0; i < symbol.nargs(); i++) {
+            if (symbol[i] != getSortRef(args[i]))
+                return false;
+        }
+    }
+    return true;
```
blishko, line 1609: This does not seem right.
I don't get the meaning of the two empty `if`s and for the last one, if the number of arguments do not match, it should definitely return `false`, no?

blishko (changes requested):

src/logics/Logic.cc
aehyvari: Yes, this PR clearly still has some placeholder code.  I'll need to polish it some.

src/logics/Logic.h
```diff
@@ -261,6 +261,12 @@ class Logic {
...
     SymRef      declareFun_Commutative_NoScoping_Chainable(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun_NoScoping_Chainable(s, rsort, args); sym_store[sr].setCommutes(); return sr; }
     SymRef      declareFun_Commutative_NoScoping_Pairwise(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun_NoScoping_Pairwise(s, rsort, args); sym_store[sr].setCommutes(); return sr; }
+
+    SymRef      declareFun_LeftAssoc(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun(s, rsort, args); sym_store[sr].setLeftAssoc(); return sr; }
+    SymRef      declareFun_RightAssoc(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun(s, rsort, args); sym_store[sr].setRightAssoc(); return sr; }
+    SymRef      declareFun_Chainable(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun(s, rsort, args); sym_store[sr].setChainable(); return sr; }
+    SymRef      declareFun_Pairwise(std::string const & s, SRef rsort, vec<SRef> const & args) { SymRef sr = declareFun(s, rsort, args); sym_store[sr].setPairwise(); return sr;}
```
aehyvari, line 268: These functions are not used anywhere in our code, but having them might make sense for the API, if somebody wants to declare his own special functions with these attributes.

src/logics/Logic.cc
```diff
@@ -1600,3 +1607,108 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+        }
+        SRef argSort = getSortRef(args[0]);
+        if (argSort != symbol[0]) {
+            why.assign(genSortMismatchString(sym, args));
+            return false;
+        }
+        if (not std::all_of(args.begin(), args.end(), [&](PTRef tr) { return argSort == getSortRef(tr); })) {
+            why.assign(genSortMismatchString(sym, args));
+            return false;
+        }
```
blishko, line 1663: These two ifs have the same body, can we merge them?

src/logics/Logic.cc
```diff
@@ -1600,3 +1607,108 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+        SRef firstSort = symbol[0];
+        SRef secondSort = symbol[1];
+        if (firstSort != getSortRef(args[0])) {
+            why.assign (genSortMismatchString(sym, args));
+            return false;
+        } else if (not std::all_of(args.begin()+1, args.end(),
+                            [&](PTRef tr) { return secondSort == getSortRef(tr); }
+                            )) {
+            why.assign(genSortMismatchString(sym, args));
+            return false;
```
blishko, line 1681: Also these two ifs could be merged.

src/logics/Logic.cc
```diff
@@ -1600,3 +1607,108 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+        }
+        SRef firstSort = symbol[0];
+        SRef secondSort = symbol[1];
+        if (secondSort != getSortRef(args.last())) {
+            why.assign(genSortMismatchString(sym, args));
+            return false;
+        } else if (not std::all_of(args.begin(), args.end()-1,
+                   [&](PTRef tr) { return firstSort == getSortRef(tr); })) {
+            why.assign(genSortMismatchString(sym, args));
+            return false;
```
blishko, line 1699: And these.

src/logics/Logic.cc
```diff
@@ -1600,3 +1607,108 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+bool Logic::typeCheck(SymRef sym, vec<PTRef> const & args, std::string & why) const {
+
+    auto genSortMismatchString = [&](SymRef sym, vec<PTRef> const & args) {
+        std::string symStr = getSymName(sym);
+        Symbol const & symbol = sym_store[sym];
+        if (symbol.chainable() or symbol.pairwise()) {
+            for (auto _ : args) {
+                (void)_;
+                symStr += " " + std::string(getSortName(symbol[0]));
+            }
```
blishko, line 1620: The raw loop seems better fitting in this case, since you don't want to do anything about the elements, just repeat some action a given number of times.

```
for (int i = 0; i < args.size(); ++i) {
    symStr += " " + std::string(getSortName(symbol[0]));
}
```

Especially for the later loops, where constructing the span seems unnecessary when a simple counter will do.

src/logics/Logic.cc
```diff
@@ -851,8 +851,17 @@ PTRef Logic::insertTerm(SymRef sym, vec<PTRef>&& terms)
...
+    assert(
+        [this](SymRef sym, vec<PTRef> const & terms) {
+            std::string why;
+            bool rval = typeCheck(sym, terms, why);
+            if (not rval) {
+                throw OsmtApiException(why);
+            }
+            return rval;
+        }(sym, terms)
+    );
```
blishko, line 863: Somehow, this `assert` does not feel like the best way, since it can never evaluate to false. It will either return true or throw an exception.
If you really want to throw an exception here, maybe we can just wrap this in `ifdef`?
```
#ifndef NDEBUG
std::string why;
if (not typeCheck(sym, terms, why)) {
   throw OsmtApiException(why);
}
#endif
```

test/unit/test_TermTypechecking.cc
```diff
@@ -0,0 +1,90 @@
...
+        , pairwiseFun(logic.declareFun_Pairwise("pf", logic.getSort_bool(), {usort, usort}))
+        , leftAssocFun(logic.declareFun_LeftAssoc("lf", usort, {usort, vsort}))
+        , rightAssocFun(logic.declareFun_RightAssoc("rf", usort, {vsort, usort}))
+        {}
+};
+
+TEST_F(TermTypecheckingTest, test_UnaryFun) {
+    std::string msg;
+    ASSERT_TRUE(logic.typeCheck(ufun1, {uvar}, msg));
+    ASSERT_TRUE(logic.typeCheck(ufun1, {uvar}, msg));
```
blishko, line 43: Duplicate lines?

blishko (commented):
I think this could be simplified a bit, but otherwise looks OK.

src/logics/Logic.cc
aehyvari: fixed

src/logics/Logic.cc
aehyvari: fixed

src/logics/Logic.cc
aehyvari: fixed

src/logics/Logic.cc
aehyvari: Well, it's kind of for each element.  But anyway, I'll change them to counters.

src/logics/Logic.cc
aehyvari: That sounds good.  I think it makes sense for now to throw until we figure out the way we want to type checking to happen and when.  Maybe an internal exception would make more sense for now?

test/unit/test_TermTypechecking.cc
aehyvari: Fixed.

src/logics/Logic.cc
```diff
@@ -1600,3 +1607,108 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+                symStr += " " + std::string(getSortName(symbol[0]));
+            }
+        } else if (symbol.left_assoc()) {
+            symStr += " " + std::string(getSortName(symbol[0]));
+            for (auto _ : opensmt::span(args.begin() + 1, args.size()-1)) {
+                (void)_;
+                symStr += " " + std::string(getSortName(symbol[1]));
+            }
+        } else if (symbol.right_assoc()) {
+            for (auto _ : opensmt::span(args.begin(), args.size()-1)) {
```
aehyvari, line 1628: This doesn't look correct.  I wanted it to go from begin to second-to-last element.  I'll fix it.

src/logics/Logic.cc
blishko: I am fine with either case.

src/logics/Logic.cc
```diff
@@ -1600,3 +1605,98 @@ char* Logic::printTerm        (PTRef tr, bool l, bool s) const { return printTer
...
+        } else if (symbol.left_assoc()) {
+            symStr += " " + std::string(getSortName(symbol[0]));
+            for (int i = 1; i < args.size(); i++) {
+                symStr += " " + std::string(getSortName(symbol[1]));
+            }
+        } else if (symbol.right_assoc()) {
+            for (int i = 0; i < args.size() - 1; i++) {
+                symStr += " " + std::string(getSortName(symbol[0]));
+            }
+            symStr += " " + std::string(getSortName(symbol[1]));
```
blishko, line 1627: We want the last argument of symbol, not the second one, no?
```suggestion
            symStr += " " + std::string(getSortName(symbol[symbol.size() - 1]));
```

blishko (commented):
I noticed one last thing, otherwise it looks good.

src/logics/Logic.cc
aehyvari: You're right, but the last element is always `symbol[1]`.  There's an internal invariant that the Left- and Right-Associative symbols must be binary in the sense that they have two arguments. This is not enforced anywhere though  For now I'll try to add an assert in `Symbol::set{Left,Right}Assoc()`.  

The `Symbol::setX()` methods should be methods in `SymbolStore`, and the chainable and pairwise arguments could be made unary with the same logic.   That's for another PR (that is yet to be written, not in my queue).

blishko: Yeah, I missed that the symbol has to be binary in this case.
Makes sense.

blishko (approved):

PR #340, opened 2021-11-13
Remove some unused parameters

blishko:
This represents a first batch of trivial cases where parameter is not used. The goal is to eventually enable -Wunused-parameter.


aehyvari (approved):
Looks good to me.

PR #343, opened 2021-11-16
CUFLogic: Remove apparently unused sort

blishko:
`CUFStr` sort is apparently not used anywhere.
Not sure what its original function was.


aehyvari (approved):
Yes, we can remove it.

PR #345, opened 2021-11-19
Tiny cleanup in Egraph

blishko:
The small changes are: 
* Passing `ERef`s by value instead of by reference.
* Do not insert constant terms into signature map. Signatures are only important for detecting congruence and constants cannot become congruent since they have no children.
* Early detection of signature-based equality by true equality.

I got 1% speedup on QF_UF, which may be just noise, but I'd like to think it's more than that :)


aehyvari (approved):
LGTM!

PR #346, opened 2021-11-19
[interpret] Give more descriptive error on unknown command

aehyvari:


src/api/smt2tokens.h
```diff
@@ -26,51 +26,96 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+        t_theory,
+        t_writestate,
+        t_readstate,
+        t_simplify,
+        t_writefuns,
+        t_let,
+        t_echo
+    };
+
+    inline const std::unordered_map<token, std::string> tokenToName = {
```
blishko, line 76: `constexpr` should also work here, no?
```suggestion
    inline constexpr std::unordered_map<token, std::string> tokenToName = {
```

blishko (commented):
Can we use `constexpr` instead of `const`?

src/api/smt2tokens.h
blishko: It seems `unordered_map` cannot be `constexpr` yet.

blishko (approved):
Nice!

PR #347, opened 2021-11-19
[Sorts] Avoid a copy of sort args when getting existing sort references

aehyvari:


blishko (approved):
I think this is good!

PR #348, opened 2021-11-19
[interpret] Minor cleanup

aehyvari:


blishko (approved):
I think this is good!

PR #349, opened 2021-11-19
[tsolvers] Unify LIA and LRA solvers

aehyvari:
I still want to make sure there's no performance regressions but this should be ready for review hopefully.


src/tsolvers/LATHandler.cc
```diff
@@ -0,0 +1,31 @@
+//
+// Created by Antti on 21.10.21.
```
blishko, line 2: Maybe use our new cool way of licensing? :)

test/unit/test_HNF.cc
```diff
@@ -0,0 +1,327 @@
...
+    MId U1 = MId_Undef;
+    MId V1 = MId_Undef;
+    ms.compute_hnf_v1(U, H, dim, U1, V1);
+    std::cout << "The hnf matrix:" << std::endl << ms.print(H) << std::endl;
+    for (int i = 0; i < rows; i++)
+        for (int j = 0; j < cols; j++)
+            ASSERT_EQ(ms.MM(H, i+1, j+1), hnf_ref[i][j]);
+}
+
+
```
blishko, line 327: There is something weird with the commit or how GitHub shows me the difference because it shows `test/unit/test_HNF.cc` once as a completly new file and once as with a single line changed.

test/unit/test_Matrix.cc
```diff
@@ -0,0 +1,566 @@
...
+
+    for (int i = 0; i < 4; i++)
+        for (int j = 0; j < 4; j++)
+            ms.MM(U, i+1, j+1) = u[i][j];
+
+    ASSERT_EQ(ms.least_in_row(U, 1, 1), 4);
+    ASSERT_EQ(ms.least_in_row(U, 2, 2), 3);
+    ASSERT_EQ(ms.least_in_row(U, 3, 3), 4);
+    ASSERT_EQ(ms.least_in_row(U, 4, 1), 0);
+}
```
blishko, line 566: And the same for `test/unit/test_Matrix.cc`

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -332,6 +352,78 @@ void LASolver::informNewSplit(PTRef tr)
...
+                } else {
+                    c = x_val.R();
+                }
+            }
+
+            // We might have this blocked already, and then the solver should essentially return "I don't know, please go ahead".
+            if (cuts[getVarId(x)].find(c) != cuts[getVarId(x)].end()) {
+                continue;
+            }
+            cuts[getVarId(x)][c] = true;
```
blishko, line 384: Do we need these `cuts`. They seem a bit weird. But that probably need more investigation not related to this PR.

blishko (commented):
I think this is pretty good!
I don't see any problem, but I have to look at this issue of duplicated files.

test/unit/test_HNF.cc
blishko: I see it now, the new file is `.cc` file, the old one is `.cpp`.
This can be done with `git mv`, so git automatically picks it up as a renaming operation.

src/tsolvers/LATHandler.cc
aehyvari: Fixed :)

src/tsolvers/lasolver/LASolver.cc
aehyvari: OK, let's look into that at a later time.  And not forget it.

test/unit/test_HNF.cc
aehyvari: It looks like both the `.cc` and `.cpp` exist in the repository, so either I forgot to remove the `.cpp` or I wanted to do this in a different commit.  Anyway, I think I managed to fix this.

test/unit/test_Matrix.cc
aehyvari: Fixed.

blishko (approved):
LGTM!

PR #350, opened 2021-11-19
Rewriting assertions

aehyvari:
I found these assertions descriptive and helpful when combining the arithmetic logics.  If we like them, we could put them to the code.


blishko (approved):
Looks good!

PR #351, opened 2021-11-22
[logic] Logic construction

aehyvari:
A suggestion for handling the different aspects of different logics without too many derived classes.


blishko (approved):
I think this is good for now. We will see if we need to adjust it in the future.
We should remove the `QF_CT` logic and maybe one of the `UNDEF`, `EMPTY`, `QF_BOOL` as well.

PR #352, opened 2021-11-22
Whitespaces and vectors

aehyvari:
Here's yet another tiny cleanup PR.  I think I implemented most of the changes in another PR and forgot to put this one in.

### DIFF
--- a/src/rewriters/Rewriter.h
+++ b/src/rewriters/Rewriter.h
@@ -14,7 +14,7 @@ class Rewriter {
 public:
     Rewriter(Logic & logic, TConfig & cfg) : logic(logic), cfg(cfg) {}
 
-    PTRef rewrite(PTRef root) {
+    virtual PTRef rewrite(PTRef root) {
         struct DFSEntry {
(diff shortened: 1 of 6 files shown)


src/rewriters/Rewriter.h
```diff
@@ -14,7 +14,7 @@ class Rewriter {
 public:
     Rewriter(Logic & logic, TConfig & cfg) : logic(logic), cfg(cfg) {}
 
-    PTRef rewrite(PTRef root) {
+    virtual PTRef rewrite(PTRef root) {
```
blishko, line 17: Why should this be virtual?

aehyvari: I believe that it's because `DivModRewriter` also has a method called `PTRef rewrite(PTRef root)`.  That method should be marked `override`.  I'll add it to the PR and let's see then if it's correct.

blishko: I see. OK, let's go with this.

blishko (approved):

PR #353, opened 2021-11-22
Changelog update

blishko:
Updates the changelog with the recent changes visible to the user.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,20 @@
 ### 2.2.1 (unreleased)
 
-API changes:
- - Interpret: the default build does not depend on line editing libraries.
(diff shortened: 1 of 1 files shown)


aehyvari:
We should add a mention on @MicTarHal's performance improvements on FastRationals.  Were there other big performance improvement recently?

blishko:
Added entries related to performance improvements.

aehyvari:
What if we put performance improvements as a first item so that it's more visible?  We don't have anything so far that would classify as a new feature, but once we have that, it'd be good to put that in a prominent place as well.

aehyvari (approved):
Good!

PR #355, opened 2021-11-23
LIA: Remove the extended signature flag

blishko:
For now we temporarily enable div and mod by default in LIA logic.
The goal is to add minimal support for non-linear arithmetic and then
allow these operations in non-linear logic only.


aehyvari (approved):
LGTM!

PR #356, opened 2021-11-23
Logic: Refactor functionality regarding interpreted/uninterpreted symbols

blishko:
Changes in this PR:
* Remove the information about interpreted symbols from `Logic`. This information is stored in `SymbolStore` and was duplicated in `Logic`.
* Pass the `interpreted` flag directly when requesting new symbol from `SymbolStore`, do not set it up afterwards.
* Clean up a few functions regarding inquiring whether something is interpreted or uninterpreted.

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -765,15 +765,19 @@ void Logic::markConstant(SymId id) {
     constants[id] = true;
 }
 
+SymRef Logic::newSymb(const char * name, const vec<SRef> & sort_args, bool isInterpreted) {
+    return sym_store.newSymb(name, sort_args, isInterpreted);
+}
+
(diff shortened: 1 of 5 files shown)


aehyvari (approved):
This looks good.  Maybe the name Logic::yieldsSortUF is a little strange, since UF doesn't sound like it should be a sort.  But I can't quickly think of a better name.

blishko:
> Maybe the name Logic::yieldsSortUF is a little strange, since UF doesn't sound like it should be a sort.

I agree. I was also thinking of `yieldsSortUninterpreted` or `yieldsUninterpretedSort`. The first one is more consistent with the arithmetic methods I guess. Should I rename it?

aehyvari:
If we both think so, maybe it's worth it.

PR #357, opened 2021-11-23
First proposal for theory combination, UFLRA specifically.

blishko:
This PR represents a prototype implementation of UFLRA theory based on delayed theory combination.
The formula is purified so that no mixed atoms are present in the formula.
Then, for every pair of interface variables, tautological clauses are added to the formula ensuring that `x = y iff x <= y and x >= y`.

Small adjustmens in `ArithLogic` and `Egraph` were necessary to properly determine which terms must be purified and which terms should be introduced to `Egraph`.


blishko:
The current version seems to be consistent, both with the previous version and with Z3 on the SMT-LIB QF_UFLRA benchmarks.

aehyvari (approved):
This looks great!  I added a note (maybe to myself?) on studying if the equality writing could be moved to a rewriter from `ArithLogic`.  I also made a PR on top of this for some small suggestions.

PR #361, opened 2021-11-30
computing interface variables

aehyvari:


aehyvari:
Here's an idea which should avoid an extra linear pass on the variables marked as uninterpreted when computing the interface variables.

blishko (approved):
Looks good!

PR #362, opened 2021-12-01
Logic: Small optimization in template instantiation

blishko:
If the template does not have any arguments, the only possible
instantiation is its body, there is no need to call the substitutor,
as it will never change the formula.


aehyvari (approved):
LGTM

PR #363, opened 2021-12-02
Remove unused debug code

blishko:
I believe the macro DEBUG_REASONS has not been used for a long time and
it seems that it requires an overload of `getReason` that does not exist
anymore. I suggest to just remove this debug code.


aehyvari (approved):
I agree, let's remove it!

PR #364, opened 2021-12-02
sorts: Fix problem where param sorts were not allocated enough memory

aehyvari:


blishko (approved):
Glad you found the problem.
I am sorry for causing it.

I am wondering if we should make it consistent with `ptermWord32Size`, which means it would be `(sizeof(Sort) + (sizeof(SRef) * size)) / sizeof(uint32_t)`.

aehyvari:
> Glad you found the problem. I am sorry for causing it.

Hehe no problem :)

> 
> I am wondering if we should make it consistent with `ptermWord32Size`, which means it would be `(sizeof(Sort) + (sizeof(SRef) * size)) / sizeof(uint32_t)`.

I it like this.

PR #367, opened 2021-12-04
Egraph: Do not backtrack on destruction

blishko:
I don't see any reason now why the `Egraph` should be backtracking in destructor.
QF_UF SMT-LIB benchmarks are fine with this change.


aehyvari (approved):
I agree.  Let's take it away.

PR #368, opened 2021-12-04
TSolver: Cleanup of reason computation and statistics 

blishko:
This PR contains two parts:
1. Introduce a separate method `getReason` in `TSolver`. This cleans up the method in `THandler` a bit and allows us to remove the boolean parameter from `TSolver::getConflict` (which was mostly ignored anyway).
2. Related to the first part, the collection and printing of statistics is cleaned up a bit as well. There are more things that could be done, but I hope this is a good start.


blishko:
Depends on #367.

PR #369, opened 2021-12-04
TSolvers: Remove useless method

blishko:
The method `TSolver::print` is not doing anything in any of the theory
solvers, I believe we should remove it.


aehyvari (approved):
Yes, this definitely can go!

PR #371, opened 2021-12-07
TSolver: Cleanup of reason computation and statistics 

blishko:
This PR contains two parts:

* Introduce a separate method `getReason` in `TSolver`. This cleans up the method in THandler a bit and allows us to remove the boolean parameter from `TSolver::getConflict` (which was mostly ignored anyway).
* Related to the first part, the collection and printing of statistics is cleaned up a bit as well. There are more things that could be done, but I hope this is a good start.


aehyvari (approved):
This looks good to me.  Sorry it took a while!

PR #372, opened 2021-12-09
UFLA: Simplify preprocessing

blishko:
Replaces the check `containsUF` with a much simpler check `isPureArithmeticEquality`.
The former check can be pretty expensive as it requires the instantiation and run of a `TermVisitor`, and it is run on every equality encountered.
The latter check is much simpler and directly checks the children of the equality. This is only possible because the subterms are purified before the equalities are considered.
The difference is not noticeable on small benchmarks, but makes a huge difference on the few extremely large benchmarks in `QF_UFLRA/cpachecker-bmc-svcomp14`. Unfortunately we still run into the blow-up when creating interface clauses in these examples. (but now we run into the blow-up much sooner! :) ).

### DIFF
--- a/src/logics/UFLATheory.cc
+++ b/src/logics/UFLATheory.cc
@@ -28,23 +28,16 @@ namespace {
         return logic.isUF(sym);
     }
 
-    bool containsUF(PTRef term, ArithLogic & logic) {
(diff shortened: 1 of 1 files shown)


aehyvari:
That's a really good catch!

aehyvari (approved):

PR #373, opened 2021-12-10
UF+LA: Model computation for the combination of theories

blishko:
The model computation for the theory combination is based on the following idea:

1. Compute variable values in `LASolver` first. In `Egraph` the equivalence classes with numeric sort needs to be assigned numeric value. 
2. In `Egraph`, Traverse the equivalence classes and if the class contains a numeric variable with known value from `LASolver`, this value is assigned to this class. 
3. If some numeric equivalence class does not have a value yet, assign it a new numeric value that is guaranteed to be different than all the numeric values assigned so far (This can be achieved by remembering maximum value seen so far and incrementing it on demand.). This ensures different `Egraph` equivalence classes have different numeric value.


aehyvari:
The smtlib scrambler used in the model validation tests doesn't understand the `(get-models)` command, so we should remove them from the scripts.

Also, the model validation regression scripts don't detect correctly the failure of the scrambler.  I have a small fix, but I could put it into a separate PR since it's not related to this.

blishko:
> Also, the model validation regression scripts don't detect correctly the failure of the scrambler. I have a small fix, but I could put it into a separate PR since it's not related to this.

OK, yes, put that to a separate PR, and we should merge that before this one, I think.

src/tsolvers/egraph/EgraphModelBuilder.cc
```diff
@@ -0,0 +1,136 @@
...
+
+        if (logic.hasSortBool(tr) and (not logic.isUP(tr))) {
+            continue; // The Boolean return sorted terms that are not uninterpreted predicates come from the SAT solver
+        }
+
+        // Check that enode_store's PTRef -> ERef conversion is consistent
+        assert(([&](PTRef tr) { ERef tmp; enode_store.peekERef(tr, tmp); return tmp == er; })(tr));
+
+        assert(not logic.isIte(tr));
+        if (((logic.isVar(tr) or logic.isConstant(tr))) and logic.yieldsSortUninterpreted(tr)) {
```
aehyvari, line 96: Extra parentheses

blishko: Fixed.

aehyvari (approved):
I think this is good now!

PR #374, opened 2021-12-14
regression: model validation script fixes

aehyvari:
Catch scrambler parsing errors and notify user that CI scripts run model regression tests.


blishko (approved):
Looks good!
(It looks like rebase is required?)

PR #376, opened 2021-12-16
UF+Arith: Enable substitutions at preprocessing

blishko:
We can run the same preprocessing of the formula as for the `UF` and `LA` theories.

It looks like the simplification code in `ArithLogic` can handle UF terms already, only a small fix was needed for the case where sum simplifies to an uninterpreted term.


src/logics/ArithLogic.cc
```diff
@@ -372,11 +372,12 @@ PTRef ArithLogic::mkPlus(vec<PTRef> && args)
     args.clear();
     SymRef s_new;
     simp.simplify(plusSym, flattened_args, s_new, args);
-    if (args.size() == 1)
-        return args[0];
     if (s_new != getPlusForSort(returnSort)) {
         return mkFun(s_new, std::move(args));
     }
+    if (args.size() == 1) {
```
blishko, line 378: Now I am thinking this might cause problems if sum simplifies to a multiplication?!

blishko: No, it seems the code handles this case correctly.

aehyvari:
Looks sufficiently innocent to me.

blishko:
I ran QF_LRA benchmarks, and QF_UFLRA all benchmarks and model validation of SAT benchmarks, everything looks good, so I think this should be good to go.

aehyvari (approved):

PR #377, opened 2021-12-17
Arith: Substitute variable by a constant when possible

blishko:
When computing substitutions from top-level arithmetic equalities, it
can happen that multiple substitutions for the same variable are
possible.
Before we just used the first one we encountered.
With this change, if we later encounter a different possible substition
with a constant, we use that one.
This should simplify the formula more.

Another possibility is to compare the complexity of the current and
alternative substitution and pick the less complex one (an example
metric would be the number of terms in the polynomial).
That I leave for future investigation.


aehyvari (approved):
Sounds good, we can do this first and increase sophistication later if we feel like it!

PR #378, opened 2021-12-18
Enable warnings on unused parameters

blishko:
This PR fixes a couple of places where compiler warns on unused parameter and enables the flag in CI runs.

There are a few places where it is not clear if the parameter should be removed completely or not.

### DIFF
--- a/src/logics/SubstLoopBreaker.h
+++ b/src/logics/SubstLoopBreaker.h
@@ -86,7 +86,7 @@ private:
     SNRef   parent;
     TVLRef  children;
 
-    SubstNode(PTRef tr, PTRef, TargetVarListAllocator& tvla)
+    SubstNode(PTRef tr, TargetVarListAllocator& tvla)
     : tvla(tvla)
     , procChild(0)
     , index(-1)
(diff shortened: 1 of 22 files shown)


src/logics/SubstLoopBreaker.h
```diff
@@ -86,7 +86,7 @@ class SubstNode {
     SNRef   parent;
     TVLRef  children;
 
-    SubstNode(PTRef tr, PTRef target, TargetVarListAllocator& tvla)
+    SubstNode(PTRef tr, PTRef, TargetVarListAllocator& tvla)
```
blishko, line 89: @aehyvari , this is one example where the parameter could be removed completely. but I don't know if this this a bug or some artifact of past.

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -2112,7 +2112,7 @@ bool CoreSMTSolver::scatterLevel()
 }
 
 
-bool CoreSMTSolver::createSplit_scatter(bool last)
+bool CoreSMTSolver::createSplit_scatter(bool)
```
blishko, line 2115: @aehyvari , and this is the second example where the parameter should be probably removed completely?!

blishko (commented):
@aehyvari Can you check the two cases I pointed out?

src/smtsolvers/CoreSMTSolver.cc
aehyvari: It looks like this can be removed.  It's not used in parallel-infra either, and there it should be tested at least.

src/logics/SubstLoopBreaker.h
aehyvari: I don't know.  I have no recollection, so probably it's just something that was refactored out but wasn't removed.

aehyvari (approved):
LGTM.  Go ahead and remove parameters also from the two ambiguous cases.

PR #380, opened 2021-12-22
Enable more compiler warnings

blishko:
`Wextra` subsumes `Wunused-parameter`.
I exclude `Wignored-qualifiers` for now, as we have quite a few of those.
We should discuss if we enable also that one (I vote "yes").

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
                 password: $DOCKERHUB_PASSWORD
             environment:
                 CMAKE_BUILD_TYPE: Release
-                FLAGS: -Wall -Wunused-parameter -Werror
+                FLAGS: -Wall -Wextra -Wno-ignored-qualifiers -Werror
                 INSTALL: ~/osmt-install
                 USE_READLINE: OFF
 
@@ -57,7 +57,7 @@ jobs:
                 command: ./ci/run_travis_commands.sh
                 environment:
                     CMAKE_BUILD_TYPE: Debug
-                    FLAGS: -Wall -Wunused-parameter -Werror -Werror=sign-compare -Wunused-lambda-capture -fsanitize=signed-integer-overflow
+                    FLAGS: -Wall -Wextra -Wno-ignored-qualifiers -Werror -Werror=sign-compare -Wunused-lambda-capture -fsanitize=signed-integer-overflow
 
     # To set up the centos-7 environment:
     #  - yum install centos-release-scl devtoolset-7 gmp-devel libedit-devel
@@ -79,7 +79,7 @@ jobs:
                 password: $DOCKERHUB_PASSWORD
               environment:
                 CMAKE_BUILD_TYPE: Release
-                FLAGS: -Wall -Wunused-parameter -Werror
+                FLAGS: -Wall -Wextra -Wno-ignored-qualifiers -Werror
                 INSTALL: ~/osmt-install
                 USE_READLINE: OFF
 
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
                 password: $DOCKERHUB_PASSWORD
             environment:
                 CMAKE_BUILD_TYPE: Release
-                FLAGS: -Wall -Wunused-parameter -Werror
+                FLAGS: -Wall -Wextra -Wno-ignored-qualifiers -Werror
```
blishko, line 11: Is there something like a YAML variable? So we could put the basic flag in one place and not repeat them in four different places...

.circleci/config.yml
```diff
@@ -57,7 +57,7 @@ jobs:
                 command: ./ci/run_travis_commands.sh
                 environment:
                     CMAKE_BUILD_TYPE: Debug
-                    FLAGS: -Wall -Wunused-parameter -Werror -Werror=sign-compare -Wunused-lambda-capture -fsanitize=signed-integer-overflow
+                    FLAGS: -Wall -Wextra -Wno-ignored-qualifiers -Werror -Werror=sign-compare -Wunused-lambda-capture -fsanitize=signed-integer-overflow
```
blishko, line 60: @aehyvari , do you know why we have the extra flags like `unused-lambda-capture` here but not anywhere else?

src/common/Timer.h
```diff
@@ -90,10 +90,7 @@ class OSMTTimeVal {
         out.systime = systime - subst.systime;
         return out;
     }
-    void operator= (const OSMTTimeVal& from) {
-        usrtime = from.usrtime;
-        systime = from.systime;
-    }
```
blishko, line 96: This is doing the same thing as default assignment operator, so it can be implicit.

src/tsolvers/bvsolver/Bvector.h
```diff
@@ -87,7 +87,7 @@ class Bvector {
         header.size      = 0;
     }
 
-    Bvector    operator=   (Bvector)         { assert(false); return *this; }
+    Bvector&    operator=   (Bvector const &) = delete;
```
blishko, line 90: @aehyvari , Do you know what is the purpose of this `Bvector` class? I have not tried, but from the search in IDE it looks like it is not used anywhere...

aehyvari: Yes it's a structure that holds the bits of a gate. I use it in the counting code that maybe goes to master in a relatively short time frame. Maybe we can keep it for now if it's not a big problem.

blishko: OK!

.circleci/config.yml
blishko: It looks like alias can help if we want to reuse the FLAGS node exactly, but not if we want to extend it...
I'll do this in another PR.

aehyvari (commented):
I fixed an error in benchmarks.  Should we also enable the warnings for ignored qualifiers?

blishko:
OK, it seems the ignored qualifiers were not that much work in the end :) Thanks!

.circleci/config.yml
blishko: OK, we could try to unify the flags later. It would be good to also have a CI run with sanitizers, I think. But all this can be done in some later PR.

aehyvari (approved):
Let's go with this then!

PR #381, opened 2021-12-27
ModelBuilder: Avoid name clash

blishko:
Model validator does not seem to like the case when function argument
has the same name as a normal variable defined in the same model.
With this change, names of the parameters are picked so the symbol with
the same name does not yet exist in the logic.

### DIFF
--- a/regression_models/instances/name_clash.smt2
+++ b/regression_models/instances/name_clash.smt2
@@ -1,0 +1,5 @@
+(set-logic QF_UFLRA)
+(declare-fun x0 () Real)
+(declare-fun f (Real Real) Real)
+(assert (distinct 0.0 (f 1.0 0.0)))
+(check-sat)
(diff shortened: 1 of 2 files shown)


blishko:
On the second thought, now I don't think this is the best solution.
Suppose that I am using OpenSMT as a library and using a single `Logic` for many queries and many models are built.
Then more and more names will be used for the symbols and this would take longer and longer to find a name that can be used.

We should probably find a different way how to deal with this problem...

aehyvari:
Yes, so this is more a problem of ModelValidator (and possibly other tools that make similar assumptions).  What if we have this as a configuration option that is on by default if we're running opensmt through `Interface`?

blishko:
> What if we have this as a configuration option that is on by default if we're running opensmt through Interface?

That is a possible solution, which should be easy to implement and sufficient for now.

PR #382, opened 2021-12-29
Interpret: Add model post-processing to avoid name clashes

blishko:
Model validator does not like when a formal argument of function
definition has the same name as a variable already defined in the model.
To work around this limitation, this patch adds a small post-processing
that checks for such name clashes in the model and renames the formal
parameters in problematic function definitions.

This is an alternative proposal to #381.

### DIFF
--- a/src/bin/Interpret.cc
+++ b/src/bin/Interpret.cc
@@ -700,7 +700,7 @@ public:
                     std::string name;
                     PTRef var = PTRef_Undef;
                     do {
-                        name = "x!" + std::to_string(num++);
+                        name = getSafePrefix(symRef) + std::to_string(num++);
                         var = logic.mkVar(logic.getSortRef(oldArg), name.c_str());
                     } while (forbiddenVars.find(var) != forbiddenVars.end());
                     newArgs.push(var);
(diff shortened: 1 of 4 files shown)


blishko:
This workaround for the limitation of the ModelValidator is a bit more verbose, but it is localized to `Interpret` so that it does not affect the use of models through API.
It is also easily reversible if it is no longer needed.

src/bin/Interpret.cc
```diff
@@ -668,27 +669,95 @@ void Interpret::getValue(const std::vector<ASTNode*>* terms)
...
+        forbiddenVars.insert(var);
+    }
+
+    void addFunction(SymRef sym) {
+        functionsToCheck.insert(sym);
+    }
+
+    std::vector<TemplateFunction> getSafeTemplates(Model const & model) {
+        std::vector<TemplateFunction> res;
+        unsigned num = 0;
```
aehyvari, line 692: It feels like by moving the counter to the top level we can avoid some traversals, since likely many functions coming from the model counter will have similarly named arguments.

aehyvari (commented):
What do you think about this, @blishko ?

blishko:
LGTM!

aehyvari (approved):
Then let's go with this.

PR #383, opened 2021-12-29
LASolver: Prepare for adding new atoms on-the-fly

blishko:
We aleady have a mechanism for adding new atoms that deals with LIA
splits. However, those atoms are special in the way that there are new
bounds on an existing variable.
This mechanism is generalized to allow bounds on arbitrary expressions
to be added dynamically.
The main issue is that new row being added to a tableaux must be
normalized, since he variables in the new expression might not be
nonbasic variables at the moment of addition.

Apart from that, we need to ensure that LASolver's data structure are
ready for dynamically added bounds of a new variable not seen at the
init stage.

The rest seems to be working nicely thanks for the mechanism of
QUASI_BASIC variables.

### DIFF
--- a/src/tsolvers/lasolver/LABounds.cc
+++ b/src/tsolvers/lasolver/LABounds.cc
@@ -85,6 +85,12 @@ void LABoundStore::clear() {
     this->bounds.clear();
 }
 
+void LABoundStore::ensureReadyFor(LVRef v) {
+    while (bounds.size() <= getVarId(v)) {
+        bounds.emplace_back();
+    }
+}
+
 void LABoundStore::updateBound(BoundInfo bi) {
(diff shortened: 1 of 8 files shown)


aehyvari (approved):
This looks good to me.  I'll push a little unit test here in a sec so I feel that I've run the code a little.

blishko:
The unit test looks reasonable, so I think it's a good idea!

I am still running the LIA benchmarks; the LRA benchmarks did not show any visible effect on the performance.

aehyvari:
Ok good, I guess we wait for those before merging!

blishko:
LIA results also do not show any change in performance, so I am going to merge this.

PR #384, opened 2021-12-31
Egraph: Introduce mechanism for dealing with dynamicly added terms

blishko:
Egraph is sensitive to backtracking and needs to undo the made changes
exactly. The important thing is to maintain the invariants of signatures
and use-vectors. Adding new terms to Egraph after initialization causes
problems for maintaining the invariants when backtracking happens.

The solution is inspired by the approach of Yices. For these newly added
terms, we remember on the backtracking stack that we need to re-analyze
them on backtracking to maintain the invariants.

Re-analysis means that we remove these terms from the signature and
use-vectors data structures if they are the representative. After the
solver has backtracked, these terms are re-introduced to these
data-structures, if they are still the representative of their class.

One problem is that it is still not entirely clear to me at which point
we can stop re-analyzing these terms.

### DIFF
--- a/src/tsolvers/egraph/EgraphSolver.cc
+++ b/src/tsolvers/egraph/EgraphSolver.cc
@@ -254,7 +254,16 @@ void Egraph::declareTerm(PTRef tr) {
     }
 
     for (auto [term, eref] : PTRefERefPairVec) {
-        updateUseVectors(term); (void)eref;
+        if (logic.getPterm(term).size() > 0) { // MB: No need to insert signatures of terms who cannot be parents
+            if (backtrack_points.size() > 0) {
+                undo_stack_main.push(Undo(REANALYZE, eref));
+            }
+            ensureUseVectorFor(eref);
+            if (not enode_store.containsSig(eref)) { // Check needed to allow dynamic addition of new terms
+                enode_store.insertSig(eref);
+                addToUseVectors(eref);
+            }
+        }
     }
 
     if (logic.hasSortBool(tr) and not logic.isDisequality(tr) and PTRefERefPairVec.size() == 2) {
@@ -682,6 +691,7 @@ void Egraph::backtrackToStackSize ( size_t size ) {
     // (might be empty, though, if boolean backtracking happens)
     explanation.clear();
     has_explanation = false;
+    vec<ERef> toReanalyze;
 
     //
     // Restore state at previous backtrack point
@@ -726,11 +736,24 @@ void Egraph::backtrackToStackSize ( size_t size ) {
                 clearPolarity(u.arg.ptr);
                 break;
             }
+            case REANALYZE: {
+                ERef term = u.arg.er;
+                assert(enode_store.lookupSig(term) != ERef_Undef);
+                if (enode_store.lookupSig(term) == term) {
+                    enode_store.removeSig(term);
+                    removeFromUseVectors(term);
+                }
+                toReanalyze.push(term);
+                break;
+            }
             default: {
                 opensmt_error("unknown action");
                 break;
             }
         }
+    }
+    for (ERef eref : toReanalyze) {
+        reanalyze(eref);
     }
 }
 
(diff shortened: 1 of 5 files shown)


blishko:
@aehyvari , this is my proposal for preparing Egraph for adding new atoms on-the-fly.
I will add a unit test as you did for the LA solver.
There is one FIXME on the code, if you have some idea about that, let me know. So far it did not cause problems for performance, so I did not think more about it.

blishko:
I have added the unit test and checked the SMT-LIB UF benchmarks, no problem there.

aehyvari:
I think I managed to convince myself that the `REANALYZE` trick is necessary, not expensive, and therefore a good way to go.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1575,3 +1596,14 @@ void Egraph::printStatistics(std::ostream & os) {
     TSolver::printStatistics(os);
     egraphStats.printStatistics(os);
 }
+
+void Egraph::reanalyze(ERef eref) {
+    // FIXME: This will probably never be false, we need to find a different condition
```
aehyvari, line 1601: Could it simply be `backtrack_points.size() > 1` ?  I also think that this has to be always true, since the backtrack point in `TSolver` is popped only after the call to `Egraph::backtrackToStackSize()` returns.  I don't have test cases that I could quickly use for checking the theory though.  If you don't have either, maybe we can postpone the discussion to once MBTC is available.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1575,3 +1596,14 @@ void Egraph::printStatistics(std::ostream & os) {
...
+    if (backtrack_points.size() > 0) {
+        undo_stack_main.push(Undo(REANALYZE, eref));
+    }
+    if (not enode_store.containsSig(eref)) {
```
aehyvari, line 1605: Is this condition always `false`?  I have the feeling that it might be, but again with no test cases I feel unsure.

blishko: I believe this can be true. Basically, when at this moment, this term is the root of its equivalence class. In our cases, these are only the newly added interface equalities and it might be there are not equivalent to other equalities at this point. And if you check the case of `REANALYZE` in backtracking, at that point I now remove these from the data structures, so here we need to re-add them.

This might be something that we can optimize here, but I think we can analyze that properly only when we have the full MBTC support.
At this moment, this change does not really have any effect on UF solving (I ran the benchmarks on the cluster).

src/tsolvers/egraph/EgraphSolver.cc
blishko: That would probably work, but I still think this can stop much sooner.

I agree that this discussion and analysis will be much easier when you can try some real examples and see how this behaves.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: Sorry, I meant the other way!  I noticed that the code in the `if` block is executed in the unit test.

I think it's fine to have it this way for now, but I'd like to not forget to figure whether this is the case, at least by the time we get some examples.  This way maybe I'll remember it better.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: > That would probably work, but I still think this can stop much sooner.

Maybe so.  In any case let's keep the `FIXME` there for now obviously.

aehyvari (approved):

src/tsolvers/egraph/EgraphSolver.cc
blishko: Right, OK. 
But I think also the other case should be possible: When this equality is *not* the root of its equivalence class and this does not change after the backtracking. Then we don't want to add it back, because it is not the representative.

I will try to add more unit tests later.

PR #387, opened 2022-01-08
Generalize split clauses

blishko:
In order to implement model-based theory combination, we need to generalize our current mechanism of split clauses used currently only for LIA splits.

Instead of a single binary clause, we need to allow solvers to communicate any clause (not just binary) and any number of them. In our proposed implementation we need 3 clauses of which 2 are binary and 1 is ternary.

However, we still need to properly maintain invariants of the SAT solver.
For example, if the new split clause should actually propagate one of its literals, we need to backtrack to the point where it would have been propagated if the clause had already existed.

### DIFF
--- a/src/smtsolvers/TheoryIF.cc
+++ b/src/smtsolvers/TheoryIF.cc
@@ -79,8 +79,23 @@ TPropRes CoreSMTSolver::handleNewSplitClauses(SplitClauses & splitClauses) {
     vec<LitLev> deds;
     deduceTheory(deds); // To remove possible theory deductions
     TPropRes res = TPropRes::Undef;
-    Lit toPropagate = lit_Undef;
-    CRef propagationReason = CRef_Undef;
+    struct PropagationData {
+        Lit lit;
+        CRef reason;
+    };
+    std::vector<PropagationData> propData;
+
+    auto processNewClause = [&] (auto const & clause) {
+        CRef cr = ca.alloc(clause, false);
+        attachClause(cr);
+        clauses.push(cr);
+        if (logsProofForInterpolation()) {
+            // MB: the proof needs to know about the new clause
+            proof->newSplitClause(cr);
+        }
+        return cr;
+    };
+
     for (auto & splitClause : splitClauses) {
         unsigned satisfied = 0;
         unsigned falsified = 0;
@@ -94,6 +109,7 @@ TPropRes CoreSMTSolver::handleNewSplitClauses(SplitClauses & splitClauses) {
             else if (value(l) == l_False) { ++falsified; }
             else { ++unknown; impliedIndex = i; }
         }
+        assert(satisfied != 0 or unknown != 0); // The clause cannot be falsified
         if (satisfied == 0 and unknown == 1) { // propagate
             // Find the lowest level where all the falsified literals are still falsified
             int backtrackLevel = 0;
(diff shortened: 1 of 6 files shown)


blishko:
@aehyvari , maybe we can have a call about this one next week.
Just writing the PR description now, I realize something.
In my original approach I only remembered the last literal that should be propagated.
However, this led to an incorrect models in some cases, because it might have happened that multiple literals were deduced  from new split clauses on level 0, but I only propagated one of them and forgot the value of the other ones.
I fixed this for level 0, but now I started thinking that I might be losing some propagated literals on levels >0 for the same reason. This does not lead to incorrect behaviour, but maybe to some inefficiencies.

It would be good to discuss this situation, once you have a look at the proposed changes.

blishko:
I was able to nicely separate this functionality into two phases as we talked about this morning.
But I thing I managed to cover all the corner cases.
Firstly, I now collect all implied literals from the level where we backtracked, instead of a single one.
Secondly, to prevent this weird situation where after backtracking a previously satisfied clause would become implying, I now sort the clauses before processing them. They are sorted according to the highest level where some of their literals was assigned. Processing the clauses in increasing order of this value ensures that this special situation with satisfied clauses cannot happen.

src/smtsolvers/TheoryIF.cc
```diff
@@ -75,75 +78,129 @@ void CoreSMTSolver::crashTest(int rounds, Var var_true, Var var_false)
...
-                // MB: the proof needs to know about the new clause
-                proof->newSplitClause(cr);
-            }
-            forced_split = ~l1;
-            return TPropRes::Decide;
+        int impliedIndex = -1;
+        for (int i = 0; i < splitClause.size(); ++i) {
+            Lit l = splitClause[i];
+            addVar_(var(l));
+            if (value(l) == l_True) { ++satisfied; }
```
aehyvari, line 141: `break`?

src/tsolvers/THandler.cc
```diff
@@ -97,32 +97,28 @@ TRes THandler::check(bool complete) {
...
-    for (int i = 0; i < logic.getPterm(tr).size(); i++) {
-        PTRef arg = logic.getPterm(tr)[i];
-        Lit l = tmap.getOrCreateLit(arg);
-        assert(getLogic().isAtom(arg)); // MB: Needs to be an atom, otherwise the declaration would not work.
-        declareAtom(arg);
-        informNewSplit(arg);
-        splits.push(l);
+    for (PTRef clause : newSplits) {
+        splitClauses.emplace_back();
+        Logic & logic = getLogic();
```
aehyvari, line 109: `const`

aehyvari (commented):
This looks really good to me.  I suppose there's no way of having a meaningful unit test at this point.

src/smtsolvers/TheoryIF.cc
blishko: I thought it would be OK, but then I spotted `addVar_`. I think in our uses cases, it would not make a difference, but in general, it's better to iterate over all the literals and ensure that the variables are known to the SAT solver.

src/tsolvers/THandler.cc
blishko: Added

blishko:
> I suppose there's no way of having a meaningful unit test at this point.

I can't think of an easy way to unit test this. We can wait till the cluster is up and run the LIA benchmarks.

src/smtsolvers/TheoryIF.cc
aehyvari: Yeah, I missed that one! Let's keep it as it is.

aehyvari (approved):

PR #388, opened 2022-01-11
UFLA: Switch to model-based theory combination

blishko:
This PR represents the last part needed to switch from delayed theory combination to the model-based theory combination.
The main part is that instead of introducing all interface equalities (and corresponding clauses) up front, they are lazily introduced when one of the theory solvers thinks they should be equal.

This requires a new functionality from TSolvers: Given a set of variables, they need to answer which of them they now consider equal in their current model. To avoid duplication, we remember which equalities has already been learned previously.


src/tsolvers/UFLATHandler.cc
```diff
@@ -31,3 +31,49 @@ lbool UFLATHandler::getPolaritySuggestion(PTRef pt) const {
...
+            continue;
+        }
+        // create clauses corresponding to "x = y iff x >= y and x <= y"
+        assert(logic.isNumEq(eq));
+        PTRef lhs = logic.getPterm(eq)[0];
+        PTRef rhs = logic.getPterm(eq)[1];
+        assert((logic.isNumVar(lhs) or logic.isNumConst(lhs)) and (logic.isNumVar(rhs) or logic.isNumConst(rhs)));
+        PTRef leq = logic.mkLeq(lhs, rhs);
+        PTRef geq = logic.mkGeq(lhs, rhs);
+        vec<PTRef> args = {eq, logic.mkNot(leq), logic.mkNot(geq)}; // trichotomy clause
```
blishko, line 67: I suspect this might cause a difference between GCC and Clang, I am not sure of the order if evaluation is fixed by the standard in branch initialization or not. To be sure, we should probably create the negation up front in a specified order.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -858,4 +858,79 @@ PTRef LASolver::getIntegerInterpolant(std::map<PTRef, icolor_t> const& labels) {
...
+            for (int j = i + 1; j < equivalentVars.size(); ++j) {
+                PTRef eq = logic.mkEq(equivalentVars[i], equivalentVars[j]);
+                if (knownEqualities.count(eq) == 0) {
+                    equalities.push(eq);
+                }
+            }
+        }
+    }
+
+    // Hack to ensure that when model is computed, delta is not picked so that it merges some of the equivalence
```
blishko, line 898: To improve this, we could lower the maximal value of Delta in Simplex from `1` to e.g. `1/2`.

Another possibility that came to my mind, would be to compute the real values first in the current model, and use the real values to find equalities. Not sure if this would be faster or slower...

blishko (commented):
Also, we need to check how this combines with basic LIA splits in QF_UFLIA.

regression/QF_UFLRA/regression_10.smt2
```diff
@@ -0,0 +1,7 @@
+(set-logic QF_UFLRA)
+(declare-const x Bool)
```
aehyvari, line 2: `(set-info :status sat)`

src/tsolvers/UFLATHandler.cc
blishko: I have read a bit about this and it seems than unlike order of evaluation for function call arguments, the order of evaluation for brace-initialization is sequenced, so this should not be a problem.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -985,4 +985,79 @@ TRes LASolver::cutFromProof() {
...
+                if (ratio < FastRational(-1)) { continue; } // MB: ratio is -delta; hence -1 <= ratio < 0
+
+                // They could be equal for the right value of delta, add equalities for cross-product
+                vec<PTRef> const & varsOfFirstVal = eqClasses.at(val);
+                vec<PTRef> const & varsOfSecondVal = entry.second;
+
+                for (PTRef var1 : varsOfFirstVal) {
+                    for (PTRef var2 : varsOfSecondVal) {
+                        PTRef eq = logic.mkEq(var1, var2);
+                        if (knownEqualities.count(eq) == 0) {
```
aehyvari, line 1053:
```suggestion
                        if (knownEqualities.find(eq) != knownEqualities.end()) {
```

aehyvari (commented):
Looks good! I left some minor comments.  I agree that the delta hack is a little annoying, but I can't think of anything better now.

src/tsolvers/lasolver/LASolver.cc
blishko: This should be `== knownEqualities.end()` and I have updated it accordingly.

aehyvari (approved):

PR #389, opened 2022-01-12
LIA: Use most-infeasible branching heuristic

blishko:
When Simplex model needs fixing because of integer domain, let's be a bit more clever than just branching on the first variable we find.

There are a couple of heuristics that could be used, most-infeasible branching seems like a good one.
It considers all integer variables that needs fixing and chooses the one with largest distance from an integer point (i.e., minimum from the two distances to the two closest points: floor and ceiling).

### DIFF
--- a/src/tsolvers/lasolver/LASolver.cc
+++ b/src/tsolvers/lasolver/LASolver.cc
@@ -346,7 +346,7 @@ void LASolver::declareAtom(PTRef leq_tr)
 
 LVRef LASolver::splitOnMostInfeasible(vec<LVRef> const & varsToFix) const {
     opensmt::Real maxDistance = 0;
-    LVRef chosen;
+    LVRef chosen = LVRef_Undef;
     for (LVRef x : varsToFix) {
         Delta val = simplex.getValuation(x);
         assert(not val.hasDelta());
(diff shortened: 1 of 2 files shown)


aehyvari (commented):
It looks good!  I tried to refactor the heuristic to a different function.  Which one is better?

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -348,69 +344,52 @@ void LASolver::declareAtom(PTRef leq_tr)
     setKnown(leq_tr);
 }
 
+LVRef LASolver::splitOnMostInfeasible(vec<LVRef> const & varsToFix) const {
+    opensmt::Real maxDistance = 0;
+    LVRef chosen;
```
blishko, line 349:
```suggestion
    LVRef chosen = LVRef_Undef;
```

blishko (commented):
I think we keep it in separate function, as you suggested.

src/tsolvers/lasolver/LASolver.cc
aehyvari: Thanks, I noticed!

aehyvari (approved):

PR #390, opened 2022-01-20
Tableau: Simplify normalization of rows

blishko:
Instead of remembering pointers to problematic variables, can can
remember the variables themselves. Then, instead of building a separate
polynomial and updating the row in one go, we can update the row for
each variable the same way we do in pivoting.

The new way avoids the need for a temporary polynomial which caused
unnecessary memory allocotations as we were gradually building it.
On the other hand, we are now updating the same row multiple times.
It is not immediatelly clear which approach is better, but the
experiments suggest this is a good idea.


aehyvari (approved):
LGTM!

PR #392, opened 2022-01-24
fix: bump minimum required flex to 2.6

guillep:
flex versions < 2.6 are not compatible with c++17


blishko (approved):

blishko:
Looks like we first need to update our CI machine that we use for checking the compatibility with SMT-COMP environment.

PR #393, opened 2022-01-24
CI: Update and cleanup the scripts

blishko:
This include the following changes:

CentOS specific:
- Use gcc-8 instead of gcc-7 on CentOS
- Add DEBUG build on CentOS
- Build newer version of `flex` manually
- Install newer version of `cmake` instead of compiling it from source

Others:
- Rename environmental variable used `INSTALL` -> `OSMT_INSTALL` (to prevent name clash in `flex` build)
- Use more sanitizers in Clang Debug build
- Create and delete `build` directory in the common script instead of having individual CI steps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,6 +113,12 @@ jobs:
                 name: Create the build directory
                 command: mkdir build
             - run:
+                name: Debug build gcc under devtoolset-8
+                command: |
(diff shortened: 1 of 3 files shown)


aehyvari (approved):
I think this is good to go now.  Good one to run the debug stuff first!

PR #396, opened 2022-01-26
Change printing methods to return string instead of char*

blishko:
This prevents memory leaks when the caller does not free the received char*. 
It also makes the methods easier to use.

### DIFF
--- a/examples/test-bv.cc
+++ b/examples/test-bv.cc
@@ -106,11 +106,10 @@ int main(int argc, char** argv)
         printf("sat\n");
         bbb.computeModel();
         PTRef v = bbb.getValue(d);
-        char * val = logic.pp(v);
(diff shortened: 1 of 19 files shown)


aehyvari (approved):
I changed also the signature of `hasQuotableCharacters`.  I think this is good.

blishko:
In this case, maybe we could simplify also the implementation of `hasQuotableCharacters` (the for loop)?

blishko:
I had my own version, but yours looks good. Let's go with it :)
I'll merge after the CI.

aehyvari:
Maybe something like this would be better?

aehyvari:
> I had my own version, but yours looks good. Let's go with it :)

Ah sorry, we were not in sync.

PR #397, opened 2022-01-27
UFInterpolator: Fix memory leak and cleanup

blishko:
`CGraph::edges` is a vector of raw pointers, but it owns the objects.
The objects were correctly deleted when all edges are removed, but not
when a single edge is to be removed.

Later we can come up with a better design, but for now we simply delete the edge
before we remove it from our vector.

I was originally worried that `UFInterpolator` accesses the removed edge after `splitEdge`, but that's not the case, since `splitEdge` correctly sets the new edge for the source node and this new edge is then used. So deleting the edge is fine.

While browsing the code, I noticed a couple of small problems with the code (mostly style), so I am including also a small cleanup here.

### DIFF
--- a/src/tsolvers/egraph/UFInterpolator.cc
+++ b/src/tsolvers/egraph/UFInterpolator.cc
@@ -75,6 +75,7 @@ CGraph::removeCEdge(CEdge * e) {
     if (e == nullptr) return;
     for (std::size_t i = 0; i < cedges.size(); ++i) {
         if (cedges[i] == e) {
+            delete e;
             cedges.erase(cedges.begin() + i);
             break;
         }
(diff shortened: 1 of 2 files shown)


src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -75,6 +77,7 @@ CGraph::removeCEdge(CEdge * e) {
     if (e == nullptr) return;
     for (std::size_t i = 0; i < cedges.size(); ++i) {
         if (cedges[i] == e) {
+            delete e;
```
blishko, line 80: This is the important change here.

src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -70,15 +71,17 @@ icolor_t UFInterpolator::colorNode(CNode * c) {
...
+            e = nullptr;
             cedges.erase(cedges.begin() + i);
             break;
         }
     }
+    assert(e == nullptr);
```
aehyvari, line 84: Do you think we can enforce this?

aehyvari:
I guess I got a little distracted, and made a pass on the style..  There are some changes that are less trivial, I tried to separate these by commits.

src/tsolvers/egraph/UFInterpolator.cc
blishko: I think so, yes.

aehyvari (approved):
LGTM!

PR #398, opened 2022-01-27
CoreSMTSolver: Fix possible access to index out of bounds

blishko:
Split clauses can contain new literals not known to the SAT solver.
When sorting split clauses, these literals are not yet introduced to the SAT solver, so we need to guard the query to the decision level.


MasoudAsadzade (approved):

aehyvari (approved):
LGTM!

PR #399, opened 2022-01-28
Everywhere: Fix several memory leaks

blishko:
This PR fixes a couple of memory leaks in the parser and other related to using `char *` for string representations.
More functions for string representations now return `std::string` instead of `char *`.


src/tsolvers/THandler.cc
```diff
@@ -445,7 +445,7 @@ PTRef   THandler::varToTerm          ( Var v ) const { return tmap.varToPTRef(v)
 Pterm&  THandler::varToPterm         ( Var v)        { return getLogic().getPterm(tmap.varToPTRef(v)); } // Return the term corresponding to a variable
 Lit     THandler::PTRefToLit         ( PTRef tr)     { return tmap.getLit(tr); }
 
-void    THandler::getVarName         ( Var v, char** name ) { *name = getLogic().printTerm(tmap.varToPTRef(v)); }
+std::string THandler::getVarName( Var v) const { return getLogic().printTerm(tmap.varToPTRef(v)); }
```
aehyvari, line 448:
```suggestion
std::string THandler::getVarName(Var v) const { return getLogic().printTerm(tmap.varToPTRef(v)); }
```

src/tsolvers/THandler.h
```diff
@@ -75,7 +75,7 @@ class THandler
     Pterm&  varToPterm         ( Var v) ;// { return getLogic().getPterm(tmap.varToPTRef(v)); } // Return the term corresponding to a variable
     Lit     PTRefToLit         ( PTRef tr);// { return tmap.getLit(tr); }
 
-    void    getVarName         ( Var v, char** name );// { *name = getLogic().printTerm(tmap.varToPTRef(v)); }
+    std::string getVarName     ( Var v) const;
```
aehyvari, line 78:
```suggestion
    std::string getVarName     (Var v) const;
```

aehyvari (approved):
LGTM!

PR #400, opened 2022-01-31
Matrix: Remove redundant member

blishko:
`ArithLogic` is not needed for `Matrix` class and can be removed.

Additionally I made small refactoring in the unit tests related to this class.


aehyvari (approved):
LGTM!

PR #402, opened 2022-02-04
Use v1.6.1 google microbenchmarks 

aehyvari:
Update needed because the old version has an initialized but unused variable causing compiler errors with the future stricter settings.


blishko (approved):
LGTM!

PR #403, opened 2022-02-04
build: Allow choosing the c++ compiler path

aehyvari:
In this PR the ci script reads the environment variable `CMAKE_CXX_COMPILER` and sets the corresponding `cmake` variable to the compiler.  This is useful at least when one wants to use on OSX a more recent compiler from `homebrew`.


blishko (approved):
LGTM!

PR #404, opened 2022-02-04
Lookahead memleak fix

aehyvari:
This PR fixes critical memory leaks in lookahead code.


blishko (approved):
I don't see any obvious problem with this, so I think it should be fine.
We could (should?) run more tests using lookahead solving.
If they show any problem we can fix it later :)

PR #409, opened 2022-02-09
Cuts-from-proofs: minor comments

aehyvari:
Here are some tiny things that I saw on the way.

### DIFF
--- a/src/tsolvers/lasolver/CutCreator.cc
+++ b/src/tsolvers/lasolver/CutCreator.cc
@@ -66,7 +66,7 @@ void swapColumns(ColMatrix & A, ColIndex pivotIndex, ColIndex otherIndex, ColMat
     U.swapCols(pivotIndex, otherIndex);
 }
 
-void addColumnMultiple(ColMatrix & A, ColIndex colFrom, opensmt::Real multiple, ColIndex colTo, ColMatrix & U) {
+void addColumnMultiple(ColMatrix & A, ColIndex colFrom, opensmt::Real const & multiple, ColIndex colTo, ColMatrix & U) {
     A[colTo].add(A[colFrom], multiple);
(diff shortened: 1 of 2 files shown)


src/tsolvers/lasolver/CutCreator.cc
```diff
@@ -100,10 +100,10 @@ bool normalizeRow(ColMatrix & A, RowIndex rowIndex, ColIndex pivotIndex, ColMatr
             return A[first].getFirstCoeff() < A[second].getFirstCoeff();
         });
         std::iter_swap(it, activeColumns.begin());
-        // Now the index of column with smallest value is first in activeColumns
+        // Now the index of column with smallest first non-zero value is first in activeColumns
```
blishko, line 103: Every column in `activeColumns` has non-zero value for the current row. This additional text is confusing for me.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -869,12 +869,13 @@ TRes LASolver::cutFromProof() {
         PTRef term = laVarMapper.getVarPTRef(var);
         auto val = isOnLower ? simplex.Lb(var) : simplex.Ub(var);
         assert(not val.hasDelta());
-        auto rhs = val.R();
+        auto const & rhs = val.R();
```
blishko, line 872: This is good. I think also `val` (2 lines above) can be a `const &`, no?

blishko (commented):
Not sure about the comment, otherwise it is good!

src/tsolvers/lasolver/CutCreator.cc
aehyvari: I can remove it... It's probably about what the reader has as implicit assumptions I suppose.  It helped me somehow.

blishko (approved):
LGTM!

PR #410, opened 2022-02-09
Cuts-from-proofs: Use uint32_t for matrix indexes

aehyvari:
This is what the cuts-from-proofs (after the minor comments) would look like if we used `uint32_t`.

### DIFF
--- a/src/tsolvers/lasolver/CutCreator.cc
+++ b/src/tsolvers/lasolver/CutCreator.cc
@@ -196,7 +196,7 @@ Representation initFromConstraints(std::vector<CutCreator::DefiningConstaint> co
         }
     };
     std::unordered_map<PTRef, unsigned, PTRefHash> varIndices;
-    std::size_t columns = 0;
+    uint32_t columns = 0;
     // First pass to assign indices and find out number of columns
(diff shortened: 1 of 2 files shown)


blishko (commented):
I am fine with this change, but the compiler requires more changes.

src/tsolvers/lasolver/CutCreator.cc
```diff
@@ -210,7 +210,7 @@ Representation initFromConstraints(std::vector<CutCreator::DefiningConstaint> co
         }
     }
 
-    auto rows = constraints.size();
+    auto rows = static_cast<uint32_t>(constraints.size());
```
blishko, line 213:
```suggestion
    uint32_t rows = constraints.size();
```

blishko (commented):
Looks good! I would drop the static_cast if compiler allows it.

src/tsolvers/lasolver/CutCreator.cc
aehyvari: Sounds good to me.  Hopefully also to all the compilers :)

blishko (approved):

PR #411, opened 2022-02-10
Cuts-from-proofs: refactor the interface

aehyvari:
Move sparse matrix representation to a separate file, and remove
dependency on ArithLogic.

This version still uses LVRef as the index type for Polynomial inside the matrix operations.


src/tsolvers/lasolver/SparseMatrix.cc
```diff
@@ -0,0 +1,168 @@
...
+    opensmt::Real sum = 0;
+    for (auto const & term : poly) {
+        uint32_t index = term.var.x;
+        assert(index < values.size());
+        sum += term.coeff * values[index];
+    }
+    return sum;
+}
+
+SparseColMatrix::TermVec SparseColMatrix::Col::toVector() const {
```
blishko, line 29: Not sure we need this `TermVec`, it's basically a restricted version of `Polynomial`.
I am a bit worried about the unnecessary copying, but it's probably not a big deal...

src/tsolvers/lasolver/CutCreator.cc
```diff
@@ -274,11 +26,11 @@ PTRef CutCreator::cut(std::vector<DefiningConstaint> const & constraints) {
     // an integer, the row represents an infeasible constraint
     for (uint32_t rowIndex = 0; rowIndex < dim; ++rowIndex) {
         auto const & row = matrixU[rowIndex];
-        auto product = crossProduct(row, varValues);
+        auto product = row.product(varValues);
         if (not product.isInteger()) {
-            return infeasibleRowToCut(row, columnMapping, logic, product);
+            return std::pair{row.toVector(), product};
```
blishko, line 31: Maybe we could use `opensmt::pair`? :)

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -855,11 +855,105 @@ bool LASolver::shouldTryCutFromProof() const {
...
+
+struct DefiningConstraint {
+    PTRef lhs;
+    opensmt::Real rhs;
+};
+
+/*
+ * Translates the set of defining constraints (linear equalities) into a suitable inner representation:
+ * - Matrix A of coefficients
+ * - a vector with right-hand-side values (as Rationals)
```
blishko, line 868: Maybe replace these two points with a combined one mentioning the `SparseLinearSystem`?

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -855,11 +855,105 @@ bool LASolver::shouldTryCutFromProof() const {
...
+        LVRef var = term.first;
+        auto const & coeff = term.second;
+        args.push(logic.mkTimes(toVarMap[var.x], logic.mkIntConst(coeff)));
+    }
+    return logic.mkPlus(std::move(args));
+}
+
+PTRef cutToSplit(CutCreator::Cut && cut, std::vector<PTRef> const & toVarMap, ArithLogic & logic) {
+    auto termVec = cut.first;
+    auto rhs = cut.second;
```
blishko, line 940:
```suggestion
    auto const & termVec = cut.first;
    auto const & rhs = cut.second;
```
To avoid copies.

blishko (commented):
I think this is good! Small suggestions to avoid copies, but otherwise it should be good to go.
You should also add yourself to the license headers now :)

src/tsolvers/lasolver/CutCreator.cc
aehyvari: Ok, absolutely!

src/tsolvers/lasolver/LASolver.cc
aehyvari: I'll try something..

src/tsolvers/lasolver/LASolver.cc
aehyvari: Ok.

src/tsolvers/lasolver/SparseMatrix.cc
aehyvari: As discussed, let's keep it as it is for now.

blishko (approved):
LGTM!

PR #412, opened 2022-02-10
Tests: Replace single hard test in our regression test suite

blishko:
This test takes long time to run under sanitizers and was causing the builds to fail sometimes in CI due to time limit.
By replacing this test with a few simple ones we still test the lookahead functionality but we can lower back the time limit and save computational resources.


aehyvari:
All these do some search, one at least up to dl 7.  They exercise back jumping, backtracking and top-level unsatisfiability in lookahead.  They only exercise unsatisfiability when building the path due to inconsistent path (the literal being currently assigned to false), but we can search later for a benchmark that would cover more.  I think this is ok now, since the timeout issue is annoying and expensive.

PR #414, opened 2022-02-11
Template polynomial on the type of variable

aehyvari:
This is an attempt towards generalising Polynomial.  At the moment the polynomial is used both in tableau and for representing columns in sparse matrices.  In the two cases the variables or indices in the polynomial have nothing to do with each other, and there is a potential risk of mixing them.  The template parameter should make this more explicit and provide some type checking.


blishko (commented):
I think this is OK, but GitHub included my commits from previous PR for some reason...

I think we should introduce a type alias for `Polynomial<LVRef>` somewhere in `Simplex/Tableau`.

blishko (approved):
LGTM!

PR #415, opened 2022-02-11
Polynomial: Add convenient overloads for merge method

blishko:


blishko:
@aehyvari , I took the idea of static temporary storage in `Polynomial` and applied it here in my PR. This way we don't need 4 overloads, just two.

aehyvari:
I added a way to provide the functions for `merge` as default arguments, based on this discussion https://stackoverflow.com/questions/9628974/why-cant-the-compiler-deduce-the-template-type-from-default-arguments

I wonder if we actually need the function types to be templated though...

aehyvari (approved):
So from my perspective this is good.

blishko:
As checked on [compiler explorer](https://godbolt.org/z/7ETqxoez4), the `std::function` is harder to deal with for the compiler than direct lambdas. For this reason, to be on the safe side, I suggest to avoid using the `std::function` as the default template argument.

aehyvari:
One last try, this time using a function pointer instead of std::function as the type for the default argumented functions.

PR #416, opened 2022-02-11
Polynomial: Implement merge using default parameters (CG F.51)

aehyvari:
This is another suggestion on how to write `PolynomialT::merge`.  Apparently default arguments are better than overloading.

### DIFF
--- a/src/tsolvers/lasolver/Polynomial.h
+++ b/src/tsolvers/lasolver/Polynomial.h
@@ -27,6 +27,7 @@ public:
     using poly_t = std::vector<Term>; // Terms are ordered by variable num
 private:
     poly_t poly;
+    static poly_t tmp_storage;
 public:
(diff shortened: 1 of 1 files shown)


blishko:
I like it, but I need to check something about the static member `tmp_storage`. I would give it `thread_local` storage.
Also, can it be defined like this in the header file? It does not cause any problems when this is included in multiple .cc files, instantiated for the same type?

blishko:
So, the the definition in the header file is correct. But it seems I actually cannot call your version with only polynomial and coefficient. The compiler says:
`No matching member function for call to 'merge' candidate template ignored: couldn't infer template argument 'ADD'`

aehyvari:
Right, maybe it was that that i was remembering when I claimed that the default arguments don't work... I don't know if there's a fundamental problem which prevents their use in this case.

PR #419, opened 2022-02-17
CoreSMTSolver: Remove forced_split field

blishko:
The `forced_split` literal does not seem to be necessary anymore.
Experiments on QF_LIA suggest this is even positive change for the LIA solver.


aehyvari (approved):
LGTM

Issue #216, opened 2021-01-21
Extend `get-interpolants` command to support tree interpolation


blishko:
Following #213 , our current implementation of `get-interpolants` command supports only path interpolation, but not tree interpolation format as defined in [the proposal](https://ultimate.informatik.uni-freiburg.de/smtinterpol/proposal.pdf).

PR #420, opened 2022-02-18
Interpolation: Start migrating functionality from ProofGraph to InterpolationContext

blishko:
This PR represents mostly the removal of interpolation methods for computing various generalizations of binary interpolation that don't work any longer, mainly because they rely on a method to compute multiple interpolants which assumes only propositional context.

We can bring these methods back afterwards. Especially, I would like to design a new API for tree interpolation that would work with the interpolation standard proposal (see https://github.com/usi-verification-and-security/opensmt/issues/216).

We also start moving the interpolation-related functionality from ProofGraph to InterpolationContext, in this PR it's `getPathInterpolants`.


aehyvari (approved):
I'm fine with removing the functionality for now, and indeed maybe in the future there will be a time to look into them again!

I'm adding some minor cleanup in a PR on top of this PR, at the risk of you having already fixed them elsewhere.

PR #424, opened 2022-02-28
Proof: Use const & in some places for ProofNode

aehyvari:


blishko:
I have tried to move most methods taking `ProofNode *` to `ProofNode const &` in #425. I will double check if I have not missed anything.

aehyvari:
Fixed in #450 .

PR #425, opened 2022-02-28
Interpolation: Separate interpolation related code to separate class and move it in InterpolationContext

blishko:
This PR continues the separation of interpolation and proof graph.
Interpolation-related functionality is moved to a separate class and hidden in `.cc` file of `InterpolationContext`.
The code of `ProofGraph` is now interpolation-agnostic.


src/proof/PGInterpolator.cc
```diff
@@ -101,44 +101,6 @@ void ProofGraph::produceSingleInterpolant ( vec<PTRef> &interpolants, const ipar
     topolSortingTopDown ( DFSv );
     size_t proof_size = DFSv.size();
 
-    // check leaves for False clause
-    // TODO do this in a better fucking way...
```
aehyvari, line 105: We're working on making the opensmt code suitable for minors.

src/proof/PG.h
```diff
@@ -251,6 +252,56 @@ struct ProofNode
...
+            }
+            else throw OsmtInternalException("Error in computing variable colors");
+        }
+    }
+
+    inline bool isColoredA(ProofNode & n, Var v) { return n.isColoredA(getSharedVarIndex(v)); }
+
+    inline bool isColoredB(ProofNode & n, Var v) { return n.isColoredB(getSharedVarIndex(v)); }
+
+    inline bool isColoredAB(ProofNode & n, Var v) { return n.isColoredAB(getSharedVarIndex(v)); }
```
aehyvari: `inline bool isColoredA(ProofNode const & n, Var v) const` etc?

src/proof/PG.h
```diff
@@ -289,11 +290,11 @@ class VarColorsCache {
...
     }
 
-    inline bool isColoredA(ProofNode & n, Var v) { return n.isColoredA(getSharedVarIndex(v)); }
+    inline bool isColoredA(ProofNode const & n, Var v) const { return n.isColoredA(getSharedVarIndex(v)); }
 
-    inline bool isColoredB(ProofNode & n, Var v) { return n.isColoredB(getSharedVarIndex(v)); }
+    inline bool isColoredB(ProofNode const & n, Var v) const { return n.isColoredB(getSharedVarIndex(v)); }
 
-    inline bool isColoredAB(ProofNode & n, Var v) { return n.isColoredAB(getSharedVarIndex(v)); }
+    inline bool isColoredAB(ProofNode const & n, Var v) const { return n.isColoredAB(getSharedVarIndex(v)); }
```
aehyvari: In fact, here they are :)

src/proof/PGInterAux.cc
```diff
@@ -77,19 +75,16 @@ icolor_t ProofGraph::getPivotColor(ProofNode & n) {
 	interpolationInfo.updateColoringfromAnts(n);
 
     // Determine if variable A-local, B-local or AB-common
-	icolor_t var_color = var_class == icolor_t::I_A || var_class == icolor_t::I_B ? var_class : icolor_t::I_UNDEF;
-	if (var_color == icolor_t::I_UNDEF) {
+	icolor_t var_color = var_class;
+	if (var_color != icolor_t::I_A and var_color != icolor_t::I_B) {
         assert(var_class == icolor_t::I_AB);
         var_color = getSharedVarColorInNode(v, n);
```
aehyvari: Assigning to `var_color` seems to be unnecessary at line 78.  Maybe we can declare it directly at line 81?

src/proof/InterpolationContext.cc
```diff
@@ -9,14 +9,832 @@
...
+        TermMapper & termMapper,
+        PartitionManager & pmanager,
+        const ProofGraph & proof,
+        const ipartitions_t & A_mask
+) : logic(theory.getLogic()), config(config), pmanager(pmanager), proofGraph(proof), thandler(new THandler(theory, termMapper)), A_mask(A_mask) {
+    auto const & vars = proof.getVariables();
+    std::size_t varCounts = (*std::max_element(vars.begin(), vars.end())) + 1;
+    nodeData.clear();
+    AB_vars_mapping.clear();
+    nodeData.resize(proof.getGraphSize());
```
blishko, line 502: It looks like the number of nodes in the proof might be exaggerated by this, since the proof stores the nodes in a vector and during reduction, some nodes are removed, but they leave a hole in this vector behind.
Compacting the vector seems to be problematic, since the nodes themselves remember their index as id, and this id is used in multiple other places.

The effect is that the interpolation data might be bigger than necessary (prepared for more nodes). It is not clear how big the effect on memory consumption is. It might be negligible.
We could store pointers to `InterpolationNodeData` instead of the objects themselves, which would mimic the old setup. But that also has some memory overhead (compared to the current version) for the nodes that are truly present in the proof.
Currently, I am inclined to keep the current version.

src/proof/InterpolationContext.cc
```diff
@@ -9,14 +9,832 @@
...
+        Theory & theory,
+        TermMapper & termMapper,
+        PartitionManager & pmanager,
+        const ProofGraph & proof,
+        const ipartitions_t & A_mask
+) : logic(theory.getLogic()), config(config), pmanager(pmanager), proofGraph(proof), thandler(new THandler(theory, termMapper)), A_mask(A_mask) {
+    auto const & vars = proof.getVariables();
+    std::size_t varCounts = (*std::max_element(vars.begin(), vars.end())) + 1;
+    nodeData.clear();
+    AB_vars_mapping.clear();
```
blishko, line 501: These calls to `clear` are not necessary. They are relicts from older version of the code where this was not a constructor but explicit `initialize` method.

aehyvari: Should they be removed?

src/proof/InterpolationContext.cc
aehyvari: I'm ok with going with your suggestion.  If it turns out in later measurements that it's a performance issue, or something that can be improved on, we can do it at that point.

aehyvari (approved):
This is a remarkable service for the community :)

PR #426, opened 2022-02-28
Return sort resolution

aehyvari:
This PR concentrates around issues related to homonymous terms with different return sorts.

- Currently the printing of similarly named terms that have different return sorts does not work in opensmt.  This PR fixes the issue by qualifying the return sorts with the `(as <symbol> <sort>)` construct when necessary.

- Currently parsing and resolving the qualified terms does not work.  This PR suggest a fix for handling these correctly.


src/pterms/PtStore.cc
```diff
@@ -67,49 +99,88 @@ SymRef PtStore::lookupSymbol(const char* s, const vec<PTRef>& args) {
...
+    assert(candidates.size() > 1);
+    if (sort == SRef_Undef) {
+        throw OsmtApiException("Ambiguous symbol: `" + std::string(s) + "'");
+    }
+
+    assert(sort != SRef_Undef);
+
+    vec<SymRef> retSortMatched;
+    retSortMatched.capacity(candidates.size());
+    for (SymRef sr : candidates) {
```
blishko, line 172: I think you can use `std::count_if` here and avoid the temporary vector. I think it will simplified the code a bit as well.

src/logics/Logic.cc
```diff
@@ -1394,7 +1403,7 @@ void
 Logic::dumpFunction(ostream& dump_out, const TemplateFunction& tpl_fun)
 {
     const std::string& name = tpl_fun.getName();
-    auto quoted_name = protectName(name, tpl_fun.getRetSort(), tpl_fun.getArgs().size() == 0);
+    auto quoted_name = protectName(name, tpl_fun.getRetSort(), tpl_fun.getArgs().size() == 0, tpl_fun.isInterpreted());
```
blishko, line 1406: In which situations is `TemplateFunction` interpreted and in which uninterpreted?

src/logics/Logic.cc
```diff
@@ -371,30 +383,28 @@ SRef Logic::declareUninterpretedSort(const std::string & name) {
...
+                argSortsString += printSort(getSortRef(tr)) + (i == args.size() - 1 ? "" : " ");
+            }
+            throw OsmtApiException("Unknown symbol `" + std::string(s) + ' ' + argSortsString + (sortRef != SRef_Undef ?  "/ " +printSort(sortRef) : "") + "'");
         }
     }
     assert(sref != SymRef_Undef);
-    PTRef rval;
+    PTRef rval = PTRef_Undef;
 
     rval = insertTerm(sref, std::move(args));
```
blishko, line 405:
```suggestion
    PTRef rval = insertTerm(sref, std::move(args));
```

src/logics/Logic.cc
```diff
@@ -371,30 +383,28 @@ SRef Logic::declareUninterpretedSort(const std::string & name) {
...
         else {
-            int written = asprintf(msg, "Unknown symbol `%s'", s);
-            assert(written >= 0); (void)written;
-            return PTRef_Undef;
+            std::string argSortsString;
+            for (int i = 0; i < args.size(); i++) {
+                PTRef tr = args[i];
+                argSortsString += printSort(getSortRef(tr)) + (i == args.size() - 1 ? "" : " ");
+            }
+            throw OsmtApiException("Unknown symbol `" + std::string(s) + ' ' + argSortsString + (sortRef != SRef_Undef ?  "/ " +printSort(sortRef) : "") + "'");
```
blishko, line 399:
```suggestion
            throw OsmtApiException("Unknown symbol `" + std::string(s) + ' ' + argSortsString + (sortRef != SRef_Undef ?  "/ " + printSort(sortRef) : "") + "'");
```
Just a missing space.

src/logics/Logic.cc
```diff
@@ -162,43 +162,59 @@ bool Logic::hasQuotableChars(std::string const & name) const
...
+// specify the return sort.
+std::string Logic::disambiguateName(std::string const & protectedName, SRef sortRef, bool isNullary) const {
+    assert(not protectedName.empty());
+    if (not isNullary) {
+        return protectedName;
+    }
+
+    auto isQuoted = [](std::string const &s) { return s.size() > 3 and *s.begin() == '|' and *(s.end()-1) == '|'; };
+    auto name = isQuoted(protectedName) ?
+            std::string_view(protectedName.data()+1, protectedName.size()-2) : std::string_view(protectedName);
```
blishko, line 176: I am a little confused why the name was protected before with `| |` and now we are removing the protection.
Isn't there some workflow when this would not happen?

src/pterms/PtStore.cc
```diff
@@ -26,26 +26,58 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
     if (values) {
-        const vec<SymRef>& trefs = *values;
+        vec<SymRef> const &trefs = *values;
         if (symstore[trefs[0]].noScoping()) {
             // No need to look forward, this is the only possible term
             // list
             for (int i = 0; i < trefs.size(); i++) {
                 SymRef ctr = trefs[i];
-                const Symbol& t = symstore[ctr];
+                Symbol const &t = symstore[ctr];
```
blishko, line 80:
```suggestion
                Symbol const & t = symstore[ctr];
```

src/pterms/PtStore.cc
```diff
@@ -26,26 +26,58 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
 // Resolves the SymRef for name s taking into account polymorphism
 // Returns SymRef_Undef if the name is not defined anywhere
 //
-SymRef PtStore::lookupSymbol(const char* s, const vec<PTRef>& args) {
+SymRef PtStore::lookupSymbol(const char* s, const vec<PTRef>& args, SRef sort) {
     auto* values = symstore.getRefOrNull(s);
+    vec<SymRef> candidates;
     if (values) {
-        const vec<SymRef>& trefs = *values;
+        vec<SymRef> const &trefs = *values;
```
blishko, line 74:
```suggestion
        vec<SymRef> const & trefs = *values;
```

src/logics/ArithLogic.cc
```diff
@@ -269,18 +269,24 @@ void ArithLogic::termSort(vec<PTRef>& v) const
     sort(v, LessThan_deepPTRef(*this));
 }
 
-std::string ArithLogic::protectName(std::string const & name, SRef retSort, bool isNullary) const {
+bool ArithLogic::isAmbiguousNullarySymbolName(std::string_view name) const {
+    int i = 0;
+    auto name_str = std::string(name);
+    for (auto sr : term_store.getHomonymousNullarySymbols(name)) {
+        if (not (sym_store[sr].isInterpreted() and (opensmt::isIntString(name_str.c_str()) or opensmt::isRealString(std::string(name_str).c_str())))) {
```
blishko, line 276:
```suggestion
        if (not (sym_store[sr].isInterpreted() and (opensmt::isIntString(name_str.c_str()) or opensmt::isRealString(name_str.c_str())))) {
```
This seems like a redundant copy.

blishko (commented):
Looks reasonable. Left a few  comments.
I want to have another look tomorrow to check if anything can be simplified further.

src/bin/Interpret.cc
```diff
@@ -389,6 +389,16 @@ PTRef Interpret::letNameResolve(const char* s, const LetRecords& letRecords) con
     return letRecords.getOrUndef(s);
 }
 
+PTRef Interpret::resolveQualifiedIdentifier(const char * name, ASTNode const * sort) {
+    SRef sr = sortFromASTNode(*sort);
```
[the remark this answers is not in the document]
aehyvari: Looks good!

src/bin/Interpret.cc
```diff
@@ -407,19 +417,30 @@ PTRef Interpret::parseTerm(const ASTNode& term, LetRecords& letRecords) {
...
 
     else if (t == QID_T) {
-        const char* name = (**(term.children->begin())).getValue();
+        if ((**(term.children->begin())).getType() == AS_T) {
+
+            auto const & as_node = **(term.children->begin());
+            const char * name = (*as_node.children)[0]->getValue();
+            const ASTNode * sortNode = (*as_node.children)[1];
+            assert(name != nullptr);
+            PTRef tr = resolveQualifiedIdentifier(name, sortNode);
```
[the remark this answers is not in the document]
aehyvari: I think I'm going to try to dereference `sortNode` already when I read it from the `as_node`.  At least it'll crash if it's a null pointer.

src/logics/ArithLogic.cc
aehyvari: Yes, I must have forgotten to fix the second.

src/logics/Logic.cc
aehyvari: I'll think about it...  Now the workflow is that 
1. protect the name
2. disambiguate the name

and it feels like it'd be better to
1. disambiguate the name
2. protect the name

However, syntactically the protection needs to be inside the disambiguation construct, so it makes more sense to have the protected name already available when disambiguating.  Otherwise we could add quotations inside the concrete string representing the disambiguation, which seems awkward.

More flexibility would be available if disambiguation and protection would be inside a single function, but in the current design protection needs to be called without disambiguation when producing formal parameters for `declare-fun`s, e.g., when printing models.

src/bin/Interpret.h
```diff
@@ -150,6 +150,7 @@ class Interpret {
 
     bool                        addLetFrame(const vec<char *> & names, vec<PTRef> const& args, LetRecords& letRecords);
     PTRef                       letNameResolve(const char* s, const LetRecords& letRecords) const;
+    PTRef                       resolveQualifiedIdentifier(const char * name, ASTNode const * sort);
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/models/FunctionTools.h
```diff
@@ -50,5 +53,6 @@ class TemplateFunction {
     PTRef getBody() const { return tr_body; }
     const vec<PTRef>& getArgs() const { return signature.getArgs(); }
     void updateBody(PTRef new_body) { tr_body = new_body; }
+    bool isInterpreted() const { return signature.getSymbolConfig().isInterpreted; }
```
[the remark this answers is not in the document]
aehyvari: You're right.  It feels silly to hard-code that all concretisations of template functions are non-interpreted, but I don't really have a use case when this would not be so.  I'll remove the functionality for now.

src/pterms/PtStore.cc
aehyvari: I tried to come up with something simpler.

src/bin/Interpret.cc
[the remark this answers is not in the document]
blishko: OK!

src/logics/ArithLogic.cc
```diff
@@ -269,18 +269,17 @@ void ArithLogic::termSort(vec<PTRef>& v) const
...
-    assert(not name.empty());
-    // TODO: We cannot use isNumConst here since it takes a SymRef.  We cannot take a SymRef because this
-    // function needs to be callable from TemplateFunction, which does not have a SymRef.
-    if (isSortNum(retSort) and isNullary and std::all_of(name.begin(), name.end(), [](unsigned char c) { return std::isdigit(c); })) {
-        return name; // Is a number, no escaping
-    } else if (hasQuotableChars(name) or std::isdigit(name[0]) or isReservedWord(name)) {
-        std::stringstream ss;
-        ss << '|' << name << '|';
-        return ss.str();
+bool ArithLogic::isAmbiguousNullarySymbolName(std::string_view name) const {
```
blishko, line 272: I am thinking that this override can also be removed, maybe?
In the `Logic::disambiguateName`, it only considers `nullary` symbols as candidates for disambiguation. What if it restricted that further to `nullary not interpreted`?

src/pterms/PtStore.cc
```diff
@@ -67,49 +99,91 @@ SymRef PtStore::lookupSymbol(const char* s, const vec<PTRef>& args) {
...
+    candidates.shrink(i-j);
 
+    if (candidates.size() == 0) {
+        return SymRef_Undef;
     }
-    // Not found
-    return SymRef_Undef;
+    assert(candidates.size() == 1);
+    assert(symstore[candidates[0]].rsort() == sort);
+    return candidates[0];
```
blishko, line 186: This looks OK, although, I would still prefer to use an STL algorithm for this.

blishko (commented):
I am wondering still, if we can use the reasoning for disambiguation as we did for quoting, and avoid the override in `ArithLogic`...

src/logics/Logic.cc
```diff
@@ -162,43 +162,59 @@ bool Logic::hasQuotableChars(std::string const & name) const
...
+    } else {
+        return protectedName;
+    }
+}
+
 //
 // Quote the name if it contains illegal characters
 //
-std::string Logic::protectName(std::string const & name, SRef, bool) const {
+std::string Logic::protectName(std::string const & name, SRef, bool isInterpreted) const {
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/logics/ArithLogic.cc
aehyvari: I think that should work. I'll commit a suggestion.

src/pterms/PtStore.cc
aehyvari: Yes, maybe so..  But at least with std::remove_if the exception throwing would mean that the predicate would have to maintain a state, which takes away at least some of its appeal.

aehyvari: As discussed, this cannot be a performance overhead in working code.

PR #427, opened 2022-02-28
CI: optional compilation with line editing

aehyvari:
Tiny PR for allowing the user to specify as an environment variable to compile opensmt with line editing.  Useful for people who want to quickly test corner case behaviour on opensmt when running locally ci scripts.


ci/run_travis_commands.sh
```diff
@@ -11,6 +11,7 @@ fi
 
 cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
       -DCMAKE_CXX_FLAGS="${FLAGS}" \
+      -DENABLE_LINE_EDITING="${ENABLE_LINE_EDITING}" \
```
blishko, line 14: Isn't this doing the same as the `ENABLE_LINE_EDITING` two lines below?

aehyvari: Yes, looks like that :)

I guess we won't need this one!

PR #431, opened 2022-03-01
Quoted symbols

aehyvari:
This PR allows Interpret and certain functions in Logic that Intepret uses for term resolution to be aware whether the user quoted an input symbol.

This is useful, for instance, in the case that a user wishes to declare an integer variable with all numeric names. 

The term resolution is implemented by a new matching system which allows specifying an element that matches a property of a symbol.  For now matchers exist for `Uninterpreted`, `Interpreted` and `Any`

In practice this system is used by `ArithLogic::lookupSymbol`, which in the case that the input symbol is essentially a numeric string that has been quoted, provides a matcher that picks only uninterpreted symbols and therefore avoids matching with the real numeric symbol.


blishko (approved):
LGTM!

PR #433, opened 2022-03-03
UFInterpolator: Fix interpolation in the present of distincts

### LINKED ISSUE #432: Interpolation error for tests in QF_UF benchmark
During interpolation test generation a [new](https://drive.google.com/file/d/1XZZZAvu6q7H7vmoZQFS-r75x0Vs8Ad5u/view?usp=sharing) file was generated, calling an `./opensmt ./2018-Goel-hwbench_QF_UF_exit.2.prop1_ab_reg_max.smt2` on this file results in an error:

`libc++abi: terminating with uncaught exception of type std::logic_error: Error in UF interpolator, could not determine the color of the conflict equality`

blishko:
UF theory conflict is caused by a disequality and a chain of (possibly
derived) equalities.
Previously, the code was assuming that it will find positive version of
the disequality in the labels map (unless the conflict is between two
interpreted constants). This assumption is not true if the disequality
is actually derived from a distinct term.
The fix is to consider also the possibility of presence of distinct
terms.

Fixes #432

### DIFF
--- a/src/tsolvers/egraph/UFInterpolator.cc
+++ b/src/tsolvers/egraph/UFInterpolator.cc
@@ -295,6 +295,42 @@ void UFInterpolator::colorCongruenceEdge(CEdge * edge) {
     }
 }
 
+icolor_t UFInterpolator::determineDisequalityColor(PTRef t1, PTRef t2, std::map<PTRef, icolor_t> const & conflictColors) const {
+    icolor_t conf_color = icolor_t::I_UNDEF;
+    PTRef eq = logic.mkEq(t1, t2);
+    if (conflictColors.find(eq) != conflictColors.end()) {
+        conf_color = conflictColors.at(eq);
+        if (conf_color == icolor_t::I_AB) {
+            conf_color = resolveABColor();
+        }
+    } else {
+        auto it = std::find_if(conflictColors.begin(), conflictColors.end(), [this](auto const & entry) { return logic.isDisequality(entry.first); });
+        if (it != conflictColors.end()) {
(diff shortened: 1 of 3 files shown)


src/tsolvers/egraph/UFInterpolator.cc
```diff
@@ -295,6 +295,42 @@ void UFInterpolator::colorCongruenceEdge(CEdge * edge) {
...
+    } else {
+        auto it = std::find_if(conflictColors.begin(), conflictColors.end(), [this](auto const & entry) { return logic.isDisequality(entry.first); });
+        if (it != conflictColors.end()) {
+            Pterm const & distinctTerm = logic.getPterm(it->first);
+            bool t1Present = std::find(distinctTerm.begin(), distinctTerm.end(), t1) != distinctTerm.end();
+            bool t2Present = std::find(distinctTerm.begin(), distinctTerm.end(), t2) != distinctTerm.end();
+            assert(t1Present and t2Present);
+            if (not(t1Present and t2Present)) {
+                throw OsmtInternalException("Error in UF interpolator, could not determine the color of the conflict equality");
+            } else {
```
aehyvari, line 315: This is essentially debug code but it could have an impact on performance, right?  Would it make sense to extract it as a lambda and put inside an assert?

aehyvari: The performance shouldn't be an issue.  The remaining question would be whether putting the lambda and its invocation inside the if statement.  Whether this is more clear seems to be a matter of taste.

PR #436, opened 2022-03-04
Splitting regression: more space-efficient solution

aehyvari:
Currently the partitioning-related regression tests repeat the same instances with only a couple of lines of changes.  This PR suggests a system where the base instance is bz2 compressed and the changes are stored as uncompressed patches obtained from `diff`.  The PR contains a script that uncompresses the base instance, applies the patch, and then proceeds with the regression test.


regression_splitting/bin/process-instance.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ -z $output ]; then
+    echo "Outputfile not provided"
+    output=/dev/stdout
+    echo "Using $output"
+fi
+
+uncompressed=${TMPDIR}file.smt2
+
+TMPDIR=$(mktemp -d)
+trap "rm -rf ${uncompressed} ${tmpdir}" EXIT
```
blishko, line 53: Not sure I am reading the shell script correctly, but it seems you are using `${TMPDIR}` before you assigned to it, no? And then in the `trap` you are using lowercase `tmpdir` instead of `TMPDIR`?

blishko (commented):
There seems to be an empty file `regression_splitting/base-instances/p2-zenonumeric_s6`?
I guess it should not be added?

regression_splitting/bin/process-instance.sh
aehyvari: That's correct.  I'l fix it.

blishko (approved):
LGTM!

PR #438, opened 2022-03-05
Interpol: Fix interpolation for split clauses

blishko:
Previously, we assumed the literals in a split clause have the same
color. This assumption was incorrect, it might happen from the
propositional point of view, these literals are assigned different
colors.
I believe split clause does not appear in the proof in single query
mode, but in incremental mode, the learnt split clause can be re-used by
the solver, in which case it will appear in the proof.

A correct solution (maybe not the only one) is to treat these split
clauses as original clauses, we just need to correctly assign color to
the clause, so that the "common variables" condition of interpolant will
be fulfilled.


aehyvari (approved):
LGTM!

PR #439, opened 2022-03-08
Async Printer and Time Checker

MasoudAsadzade:
This pull request consists of: 
1) An async printer to output coloured text which can be used in multi-threading context.
  use case: 
     A shared cross-thread object of opensmt::synced_stream, synced_stream, can be used as follow:
      synced_stream(std::clog);
      synced_stream.println(Color::FG_BrightCyan, "status:", status, ...);
    to have a thread-safe coloured stream.

2) A manual time checker which can start, stop, accumulate and reset the time
    use case:
        opensmt::StoppableWatch timer;
        timer.start();
        timer.elapsed_time_milliseconds()

3) An async time checker/printer which triggers on contructor/destructor, automatically measures the time
  use case:
     opensmt::PrintStopWatch timer("wait time ", synced_stream, Color::Code:: FG_BrightCyan );


MasoudAsadzade:
I closed this PR because it was moved to a new repository, partition-channel.

PR #441, opened 2022-03-08
Enable interpolant validation in regression tests

blishko:
This PR suggests to switch from external to internal interpolant validation with the aim of enabling interpolant validations in our regression tests.
The successful interpolant validation is asserted, so any violation in our regression test suite will be caught in CI.

This already caught a problem in how we determine how the input formula is split into A and B part.
For now, the validation is enabled only on some benchmarks in our suite, as there are still problems with validation of path interpolants or interpolants in incremental mode. This I plan to address in other PRs.


aehyvari (approved):
LGTM!

PR #445, opened 2022-03-10
Tests: Update exact output tests to have unique model

blishko:
In regression tests where we check exact match to expected output we
should try to use instances with unique model when model is queried.
This would prevent test failures when we change how a model is computed.


aehyvari (approved):
LGTM!

PR #446, opened 2022-03-10
LASolver: Fix performance regression related to thread-local storage

blishko:
Apparently, TLS is not for free and using it for helper storage in the
performance critical Polynomial::merge seems to have noticeable
performance penalty.

We revert to the previous behaviour where we keep the temporary storage
in the Tableau instance and pass it to `merge` as an argument.


blishko:
Here is the difference on LRA benchmarks: 3% improvement over master.
[comparison.pdf](https://github.com/usi-verification-and-security/opensmt/files/8225322/comparison.pdf)

blishko:
It seems that the performance can be improved while keeping the TLS variable.
The solution is to immediately store its address in the local variable in `Polynomial::merge` and access only this local variable afterwards.
It seems it is still a little bit slower than the version in this PR (~1%). It might also be noise, but on my local, it seems to be a tiny bit slower.

So it is possible to recover most of the performance back while keeping the code nice with `thread_local` temporary storage, but if we want to squeeze the most out of it, it seems the solution in this PR is better.

aehyvari (approved):
This looks like the most elegant solution.

PR #451, opened 2022-03-15
Substitutor: Optimize substitution application to variables

blishko:
The machinery of Rewriter is general, but also not optimal for simple
rewritings. Most importantly, it allocates a boolean mask (represented with
vector) to keep track of processed terms. This allocation is often
wasteful a should probably be re-thought.

Here we optimize Substitutor (a sub-class of Rewriter) to avoid the
general machinery when the term to process is a variable.
In this case, the general algorithm can be replaced by a simple lookup
in the substitutions map.

### DIFF
--- a/src/rewriters/Rewriter.h
+++ b/src/rewriters/Rewriter.h
@@ -16,6 +16,11 @@ public:
     Rewriter(Logic & logic, TConfig & cfg) : logic(logic), cfg(cfg) {}
 
     virtual PTRef rewrite(PTRef root) {
+        // MB: If term has no children then single call to config is enough; this can save memory allocations if successful
+        if (logic.isVar(root)) {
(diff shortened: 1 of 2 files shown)


blishko:
I have been adding some preprocessing and Golem and I noticed that a large amount of time is spent in my trivial quantifier elimination, particularly in `Logic::substitutionsTransitiveClosure` that I transitively use. In my case, I have a lot of substitutions of a variable for another variable and this was causing a lot of time spent in the `Substitutor` which was allocating and destroying the helper vector, which is totally unnecessary when trying to rewrite a variable.

I believe the proposed change should be beneficial in general to all parts of OpenSMT that performs substitutions.

aehyvari (approved):
LGTM!

aehyvari:
Ok, good!

PR #453, opened 2022-03-15
Enable validation of path interpolants

blishko:
This PR continues the work of #441 and enables internal validation of path interpolation property.
It also implements missing functionality in `PartitionManager` to return any subset of top-level asserted formulas, which is needed to validate path interpolants.


aehyvari (approved):
LGTM!

PR #454, opened 2022-03-15
ArithLogic: Use direct simplification for sum of constants

blishko:
In ArithLogic::mkPlus, when combining linear terms with the same variable, we collect the constants from all such terms. Then we need to sum these constants together. Instead of calling mkPlus again, we can directly call the constant simplification procedure, as we know that all the terms we are trying to sum must be numeric constants.
This saves some unnecessary vector copying.

I also suggest to move the simplification helpers into the `.cc` file, so we don't expose them unnecessarily.


aehyvari (approved):
LGTM!

PR #456, opened 2022-03-16
CMake: Add an option to prefer static GMP libraries

blishko:
Here is my take on how to force CMake to prefer static over shared GMP library, if we want to.
This is useful for creating statically linked executables.


aehyvari (approved):
Looks good to me, and great work on diving into how to find the static libs with cmake!  I got inspired by this and wrote some stuff which I'll put on another PR though.

PR #458, opened 2022-03-19
Rewriter: Optimize memory (re-)usage

blishko:
This PR suggests 2 ways to optimize memory usage in Rewriter:
1. The auxiliary vector for marking processed terms can be reused. I suggest to put it in `Logic` as it can be re-used for other purposes as well. Using `nat_set` instead of `vec` has the advantage that it can be cleared in constant time.
2. When building a new version of the term with rewriting already applied to its children, we need to collect the new children. Previously we were constructing new vector for every term, we can re-use a single vector for the whole rewriting process.


aehyvari (approved):
LGTM!  We should maybe do the same for `TermVisitor`?

blishko:
Added the same for `TermVisitor` as well.

PR #461, opened 2022-03-22
UFInterpolator: Add handling of a special case

### LINKED ISSUE #459: Assertion faliure during opensmt interpolant construction
## Status Quo

During the interpolant construction using this [file](https://drive.google.com/file/d/1W1rpwkTvzph85cW63A3qMwo1Vy199Bjp/view?usp=sharing) the assertion fails. Here is the received message:
`Assertion failed: (from->next == nullptr), function addCEdge, file /Users/konstantin.britikov/Documents/opensmt/src/tsolvers/egraph/UFInterpolator.cc, line 101.`

## Acceptance criteria

Interpolation search doesn't fail.

blishko:
When UF interpolation encounters an edge that is not colorable, it must split the edge. 
So given an edge U -> V, it will find an AB-colorable term W that is congruent with both U and V, and it will replace the edge with two edges: U -> W -> V.
However, suppose that W had an edge W -> Z. Then in general, the new path should be U -> W -> V -> Z.
But, there are two special cases: when Z = U and when Z = V.
The second special case has been handled in the code before. The first one was not.
In this case, the last edge would complete a triangle loop U -> W -> V -> U.
Instead, the path should be just U -> W -> V, without the last edge added.

Fixes #459.


aehyvari (approved):
Great, looks good!  The naming that `CNode::next` is a `CEdge` is super confusing.  I had to rename it to be able to review, so I'll make a quick PR with the change.

blishko:
OK, I'll merge this one, and I'll await your naming (and possibly style) changes. I think that's a good idea!

PR #465, opened 2022-03-24
Logic: Use wrapper with nicer interface for term marking

blishko:
We can hide the implementation details of using nat_set and provide a nicer interface (especially the constructor takes care of setting up the  low-level data structure) for term markings.

We can also mark the nat_set mutable (basically not considering it a logical part of the Logic state) and thus allowing the method for creating TermMarks helper constant on Logic.

### DIFF
--- a/src/common/TreeOps.h
+++ b/src/common/TreeOps.h
@@ -20,10 +20,10 @@
 
 template<typename TConfig>
 class TermVisitor {
-    Logic & logic;
+    Logic const & logic;
     TConfig & cfg;
 public:
-    TermVisitor(Logic & logic, TConfig & cfg) : logic(logic), cfg(cfg) {}
+    TermVisitor(Logic const & logic, TConfig & cfg) : logic(logic), cfg(cfg) {}
 
     void visit(PTRef root) {
         struct DFSEntry {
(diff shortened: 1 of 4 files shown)


aehyvari:
Maybe not so much related to this PR (only) but anyway: should we remove the special case handling for variables in [src/rewriters/Rewriter.h](https://github.com/usi-verification-and-security/opensmt/blob/6880db0a352c7fc7978eecc8f8fcf03468a63f25/src/rewriters/Rewriter.h#L24)?

blishko:
I don't know, it still seems to me like a reasonable optimization. There is still a bunch of work that is done which can be avoided (even though the memory allocation of bool marks will most likely not happen). Maybe the comment should be updated, but I would keep the check.

aehyvari (approved):
This is great!

src/common/TreeOps.h
```diff
@@ -32,37 +32,34 @@ class TermVisitor {
             unsigned int nextChild = 0;
         };
         // MB: Relies on an invariant that id of a child is lower than id of a parent.
```
aehyvari, line 34: We should remove this comment, right?

blishko: Good point, yes, we should remove it, I'll do it.

src/rewriters/Rewriter.h
```diff
@@ -31,31 +31,28 @@ class Rewriter {
             unsigned int nextChild = 0;
         };
         // MB: Relies on an invariant that id of a child is lower than id of a parent.
```
blishko, line 33: And the same comment is here, so this should also be removed.

blishko: It turns out this comment is still relevant (something to discuss, perhaps), but I move it to the description of the method instead. Feels like a better place for it now.

PR #467, opened 2022-03-31
Logic: make abstract value counter member of logic

aehyvari:
Right now I see no reason why the abstract value counter, used for getting unique names for values in uninterpreted function models, should be a global variable.  This PR suggests to turn it into a member of `Logic`.


blishko (approved):
I agree. I don't see a problem of this being member of Logic.

BTW, I think we should start using in-class member initializers whenever possible (C++ Core Guidelines C.45). Maybe we can start here? :)

aehyvari:
OK, we can start small :)

PR #474, opened 2022-04-17
Terms: Add generic subterms collector

blishko:
This PR adds a generic visitor for collecting subterms of a term, templated by a predicate the subterms need to satisfy.
Two helper methods are added for collecting all subterms and for collecting all variables.

This follows the pattern of uniting all operations on terms on our base `Visitor` and `Rewriter`.

Some existing methods are marked as deprecated and I suggest to remove them in the future. This is subject for discussion.

### DIFF
--- a/src/common/TreeOps.h
+++ b/src/common/TreeOps.h
@@ -270,18 +270,18 @@ public:
 };
 
 template<typename TPred>
-static vec<PTRef> subTerms(Logic & logic, PTRef term, TPred predicate) {
+static vec<PTRef> subTerms(Logic const & logic, PTRef term, TPred predicate) {
     TermCollectorConfig<TPred> config(predicate);
     TermVisitor<decltype(config)>(logic, config).visit(term);
     return config.extractCollectedTerms();
 }
 
-inline vec<PTRef> subTerms(Logic & logic, PTRef term) {
+inline vec<PTRef> subTerms(Logic const & logic, PTRef term) {
     return subTerms(logic, term, [](PTRef){ return true; });
 }
 
 /* Returns all variables present in the given term */
-inline vec<PTRef> variables(Logic & logic, PTRef term) {
+inline vec<PTRef> variables(Logic const & logic, PTRef term) {
     return subTerms(logic, term, [&](PTRef subTerm) { return logic.isVar(subTerm); });
 }
 
--- a/src/common/TreeOps.h
+++ b/src/common/TreeOps.h
@@ -221,6 +221,7 @@ getVars(PTRef tr, Logic& logic, MapWithKeys<PTRef,bool,PTRefHash>& vars)
     }
 }
 
+[[deprecated]]
 inline std::vector<PTRef>
 getAtoms(PTRef tr, Logic & logic)
 {
--- a/src/common/TreeOps.h
+++ b/src/common/TreeOps.h
@@ -270,19 +270,19 @@ public:
 };
 
 template<typename TPred>
-static vec<PTRef> subTerms(Logic const & logic, PTRef term, TPred predicate) {
+static vec<PTRef> matchingSubTerms(Logic const & logic, PTRef term, TPred predicate) {
     TermCollectorConfig<TPred> config(predicate);
     TermVisitor<decltype(config)>(logic, config).visit(term);
     return config.extractCollectedTerms();
 }
 
 inline vec<PTRef> subTerms(Logic const & logic, PTRef term) {
-    return subTerms(logic, term, [](PTRef){ return true; });
+    return matchingSubTerms(logic, term, [](PTRef) { return true; });
 }
 
 /* Returns all variables present in the given term */
 inline vec<PTRef> variables(Logic const & logic, PTRef term) {
-    return subTerms(logic, term, [&](PTRef subTerm) { return logic.isVar(subTerm); });
+    return matchingSubTerms(logic, term, [&](PTRef subTerm) { return logic.isVar(subTerm); });
 }
 
 #endif
--- a/src/common/TreeOps.h
+++ b/src/common/TreeOps.h
@@ -185,6 +185,7 @@ void getTermList(PTRef tr, vec<T>& list_out, Logic& logic) {
 
 // Get variables starting from the root
 //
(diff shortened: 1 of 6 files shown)


src/common/TreeOps.h
```diff
@@ -255,4 +259,30 @@ getAtoms(PTRef tr, Logic & logic)
     return atoms;
 }
 
+template<typename TPred>
+class TermCollectorConfig : public DefaultVisitorConfig {
+    TPred predicate;
+    vec<PTRef> gatheredTerms;
+public:
+    TermCollectorConfig(TPred predicate) : predicate(std::move(predicate)) {}
```
aehyvari, line 267:
```suggestion
    TermCollectorConfig(TPred && predicate) : predicate(std::move(predicate)) {}
```

aehyvari (commented):
This looks really good!  I left some comments and ideas. 

Can we convert also `getAtoms` in TreeOps.h?

src/common/TreeOps.h
blishko: I prefer passing the predicate by value.
Here is some discussion on this topic: https://stackoverflow.com/questions/48803927/why-does-stdfind-iffirst-last-p-not-take-predicate-by-reference

src/common/TreeOps.h
```diff
@@ -255,4 +259,30 @@ getAtoms(PTRef tr, Logic & logic)
...
+    TermVisitor<decltype(config)>(logic, config).visit(term);
+    return config.extractCollectedTerms();
+}
+
+inline vec<PTRef> subTerms(Logic & logic, PTRef term) {
+    return subTerms(logic, term, [](PTRef){ return true; });
+}
+
+/* Returns all variables present in the given term */
+inline vec<PTRef> variables(Logic & logic, PTRef term) {
```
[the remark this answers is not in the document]
blishko: Fixed

blishko:
> Can we convert also getAtoms in TreeOps.h?

Probably yes, although it seems to have some non-trivial logic, does not examining terms beyond children of `BooleanOperator`.
We would have to extend `TermCollectorConfig` to accept override for `previsit` as well, I guess...
Seems a bit more complicated than what I wanted to achieve at the moment.

I am also thinking of having a general method that wouldn't collect the subterms, but it would instead accept an action that would be executed on every subterm (satisfying given predicate). Then we could maybe avoid **collecting** subterms altogether.

aehyvari (commented):
I had these comments that I didn't have time to finish yesterday.

src/common/TreeOps.h
```diff
@@ -255,4 +259,30 @@ getAtoms(PTRef tr, Logic & logic)
...
+template<typename TPred>
+static vec<PTRef> subTerms(Logic & logic, PTRef term, TPred predicate) {
+    TermCollectorConfig<TPred> config(predicate);
+    TermVisitor<decltype(config)>(logic, config).visit(term);
+    return config.extractCollectedTerms();
+}
+
+inline vec<PTRef> subTerms(Logic & logic, PTRef term) {
+    return subTerms(logic, term, [](PTRef){ return true; });
+}
```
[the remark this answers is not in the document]
blishko: `matchingSubTerms` sounds good to me. I can rename it.

src/common/TreeOps.h
aehyvari: > > Can we convert also getAtoms in TreeOps.h?
> 
> Probably yes, although it seems to have some non-trivial logic, does not examining terms beyond children of `BooleanOperator`. We would have to extend `TermCollectorConfig` to accept override for `previsit` as well, I guess... Seems a bit more complicated than what I wanted to achieve at the moment.

Maybe so...  Do you know if it is currently used outside opensmt?  The only use within opensmt is in a unit test.

> 
> I am also thinking of having a general method that wouldn't collect the subterms, but it would instead accept an action that would be executed on every subterm (satisfying given predicate). Then we could maybe avoid **collecting** subterms altogether.

Ok, sounds like defining the semantics of that operation is not trivial (e.g., can it change the tree?), but also good if we can make it work for something.

aehyvari:
> > Can we convert also getAtoms in TreeOps.h?
> 
> Probably yes, although it seems to have some non-trivial logic, does not examining terms beyond children of `BooleanOperator`. We would have to extend `TermCollectorConfig` to accept override for `previsit` as well, I guess... Seems a bit more complicated than what I wanted to achieve at the moment.

Maybe so...  Do you know if it is currently used outside opensmt?  The only use within opensmt is in a unit test.

> 
> I am also thinking of having a general method that wouldn't collect the subterms, but it would instead accept an action that would be executed on every subterm (satisfying given predicate). Then we could maybe avoid **collecting** subterms altogether.

Ok, sounds like defining the semantics of that operation is not trivial (e.g., can it change the tree?), but also good if we can make it work for something.

src/common/TreeOps.h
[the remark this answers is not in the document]
blishko: Done

blishko:
> Maybe so... Do you know if it is currently used outside opensmt? The only use within opensmt is in a unit test.

Not in Golem. Otherwise, I don't know. We can also just deprecate it, without a replacement.

aehyvari:
> > Maybe so... Do you know if it is currently used outside opensmt? The only use within opensmt is in a unit test.
> 
> Not in Golem. Otherwise, I don't know. We can also just deprecate it, without a replacement.

Let's do that.  The unit test can use a simple implementation that applies `matchingSubTerms` with the existing logic, I think that I checked yesterday and it worked.

blishko:
> Let's do that. The unit test can use a simple implementation that applies matchingSubTerms with the existing logic, I think that I checked yesterday and it worked.

It turns out it is possible to replace the calls to `getAtoms` in unit tests with calls to `variables`, for the terms in the tests, it is checking the same thing.

aehyvari:
> It turns out it is possible to replace the calls to getAtoms in unit tests with calls to variables, for the terms in the tests, it is checking the same thing.

Looks like a good solution.

src/common/TreeOps.h
aehyvari: Interesting 🤔

Ok, let's go with the value.

PR #478, opened 2022-04-21
CI: Make build script build-system-agnostic

blishko:
CMake can be instructed to build the project with the build system that has been used in configuration.
For example, I now use ninja as the default build system in CMake, so
the script did not work for me before.


aehyvari (approved):
LGTM!

PR #480, opened 2022-04-27
SAT improvements

aehyvari:
This PR introduces three improvements for opensmt's SAT solver:
 * Polarity saving
 * Luby restarts
 * Glue heuristic for removing learned clauses

Together they improve in general the performance in smt-lib benchmarks in the theories that opensmt supports.


src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -1197,18 +1184,22 @@ struct reduceDB_lt
...
+    // and clauses with high glue score
+    int extra = 0;
     for (i = j = 0; i < learnts.size(); i++)
     {
         Clause& c = ca[learnts[i]];
-        if (c.size() > 2 && !locked(c) && (i < learnts.size() / 2 || c.activity() < extra_lim))
+        if (c.getGlue() <= 3) {
+            extra++;
+        }
+        if (c.getGlue() > 3 and c.size() > 2 and not locked(c) and (i+extra < learnts.size() / 2)) {
```
blishko, line 1198: From what I understand, glue is always `<= 3`, no? That would make this condition always false (and the one above always true). What is the intention here?

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -231,11 +206,10 @@ Var CoreSMTSolver::newVar(bool sign, bool dvar)
...
     seen     .push(0);
-    polarity .push(sign);
     decision .push();
     trail    .capacity(v+1);
     setDecisionVar(v, dvar);
-    polarity    .push((char)sign);
+    savedPolarity.push(true);
 
 #if CACHE_POLARITY
     prev_polarity.push(toInt(l_Undef));
```
blishko, line 215: I saw in `CoreSMTSolver.h` that member `prev_polarity` (together with the macro) was deleted, so I believe this should also be removed from here, no?

blishko (commented):
First observations from quick scanning of the changes.

src/smtsolvers/CoreSMTSolver.cc
aehyvari: That's true.  Previously glue was always `<= 10`, but the test was the same as above.  I seem to have gotten confused with the bits.  I'll commit a fix for this one to see if I get it right now.

src/minisat/core/SolverTypes.h
```diff
@@ -140,7 +140,8 @@ class Clause {
         unsigned learnt    : 1;
         unsigned has_extra : 1;
         unsigned reloced   : 1;
-        unsigned size      : 27; }                            header;
+        unsigned glue      : 3;
```
msoos, line 143: I have had issues before where having too few bits for glues was causing issues. Some instances caused very high glues for all clauses, leading to the max (in this case, 7), assigned to all clauses, which consequently lead to extremely bad performance. This may not hit you at all, depending on the problems you are solving. So this is just a heads-up in case some instance is suddenly very slow/impossible to solve, it could be due to this.

I would suggest a few more bits. Even better, not bitstuffing it and using 16b/32b may be safer. But maybe I'm just paranoid :S

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -608,6 +566,30 @@ Lit CoreSMTSolver::pickBranchLit()
...
+uint32_t CoreSMTSolver::computeGlue(vec<Lit> const & ps) {
+    levelsInClause.reset();
+    uint32_t numLevels = 0;
+    for (Lit lit: ps) {
+        int level = vardata[var(lit)].level;
+        if (level != 0 and not levelsInClause.contains(level)) {
+            levelsInClause.insert(level);
+            ++ numLevels;
+            if (numLevels >= 4) {
+                break;
```
msoos, line 586: I think you can go to 7 here, so maybe you want to up this `4` to `7`? More resolution does seem to help. 4 seems very low, even 7 may be too low IMHO.

src/smtsolvers/CoreSMTSolver.cc
msoos: 3 is the cutoff in CMS to "keep forever". Some solvers use 2. It may be interesting to experiment.

src/minisat/core/SolverTypes.h
msoos, line 143: I have had issues before where having too few bits for glues was causing issues. Some instances caused very high glues for all clauses, leading to the max (in this case, 7), assigned to all clauses, which consequently lead to extremely bad performance. This may not hit you at all, depending on the problems you are solving. So this is just a heads-up in case some instance is suddenly very slow/impossible to solve, it could be due to this.

I would suggest a few more bits. Even better, not bitstuffing it and using 16b/32b may be safer. But maybe I'm just paranoid :S

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -1465,6 +1444,10 @@ lbool CoreSMTSolver::search(int nof_conflicts, int nof_learnts)
 
         CRef confl = propagate();
         if (confl != CRef_Undef) {
+            if (conflicts > conflictsUntilFlip) {
+                conflictsUntilFlip += flipIncrement;
+                flipState = not flipState;
```
msoos, line 1449: Ahh cool! I like this fix to have inverse stable polarities too :)

msoos (commented):
Sounds great! I'd consider improving the resolution of glue, I think 3 bits may not be enough. I'd go for 10 bits or more :) Just my 2 cents!

Great to see that it's helping with instances!

msoos:
Ah wait, while taking a shower I understood your rationale for the low resolution glues! Sorry! Yeah, you are right, we don't need so many bits for the glues. I was still stuck in the stone age when we used to use glues to sort clause usefulness and then the resolution really mattered. But as far as the modern systems go, we just check whether it's low to lock it in, otherwise, we don't care. So your system works like a charm. Sorry! Let me close my review :)

src/smtsolvers/CoreSMTSolver.cc
aehyvari: This was already in your original PR.  I messed up the updating of `conflictsUntilFlip`, this is brought "back" in commit 45711aad.

src/smtsolvers/CoreSMTSolver.cc
aehyvari: We could try the cutoff of 2 I guess.  We can then free one bit again for future use, if any :)

blishko (approved):
I think this is good from my side.
@aehyvari , do you want to re-run the experiments to check for the difference with computing glue in learnt clauses? Or do we trust that this will not be that much of a difference (as in LRA) and just go with it right now?

aehyvari:
> do you want to re-run the experiments to check for the difference with computing glue in learnt clauses? Or do we trust that this will not be that much of a difference (as in LRA) and just go with it right now?

Maybe we can run them, it's not that huge an overhead considering how much we've used so far and will give us the final, overall picture of this set of improvements.  And we did have this surprise slowdown with me messing up the branching heuristic already while rebasing.

PR #482, opened 2022-05-03
Adding vivification + glue recalc on analyze

msooseth:
Well, here goes another try. The vilification should be limited, to be honest, by mems (as Donald Knuth puts it), but I didn't bother because this is just a quick hack to see what gives. Also, I added the `-v` option to `opensmt` because I needed visibility :)

Recomputing the glue during analyze is standard. All systems I know do this, and they update the glue only in case it's lower. This way, as we progress the search, and reach a part of the search where the clause would be low-glue, we lock the clause in.

The code actually looks cleaner than I imagined when I was writing it... Huh. Maybe I'm not the worst coder in the universe, only in the bottom 25%? :smiley:

### DIFF
--- a/src/bin/opensmt.cc
+++ b/src/bin/opensmt.cc
@@ -94,7 +94,7 @@ int main( int argc, char * argv[] )
 
     SMTConfig c;
     bool pipe = false;
-    while ((opt = getopt(argc, argv, "hdpir:")) != -1) {
+    while ((opt = getopt(argc, argv, "hdpir:v")) != -1) {
         switch (opt) {
 
             case 'h':
@@ -112,6 +112,9 @@ int main( int argc, char * argv[] )
                 break;
             case 'i':
                 c.setOption(SMTConfig::o_produce_inter, SMTOption(true), msg);
+                break;
+            case 'v':
+                c.setOption(SMTConfig::o_verbosity, SMTOption(true), msg);
                 break;
             case 'p':
                 pipe = true;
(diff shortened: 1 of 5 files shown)


aehyvari:
Wow, thanks @msooseth !  I'm going to check soon, right now we've a busy cluster...

aehyvari:
I'll start to post the results here for discussion.

This is `QF_LIA` against the current master (on x-axis).

non-incremental-QF_LIA-b8597d1e-2022-04-30_vs_fc0cd483-2022-05-07

msooseth:
Hmmm it's supposed to be much clearer than this... Can you please send me
the files you are using as input? You mentioned you are using SMTLIB, but
not a single one of them takes more than 100s on my machine and here you
have 1000+ solvings, so something is off :S. Can you please send a link to
the input files you are using? I'm a bit confused, probably did something
wrong in downloading them :(

aehyvari:
They're from QF_LIA category of the official smt-lib repository.

https://clc-gitlab.cs.uiowa.edu:2443/SMT-LIB-benchmarks/QF_LIA

blishko:
I would split the fix for the glue recomputation from the vivification. These seem to be two independent things, no?
The command line option for verbosity seems useful on its own, I think we should put that into a separate pull request.

msooseth:
Yeah, I'll split it to 3. Sorry!

blishko:
I am closing this one, since the work here has been handled in separate PRs.

PR #484, opened 2022-05-07
Lia random splitting

aehyvari:
The LIA solver, upon finding a solution to the linear system assigning non-integer values to integer variables, needs to communicate to the SAT solver how to fix a value to an integer.  Previously this was done by choosing the variable to be fixed by choosing the one furthest away from an integer.  It turns out that a completely random heuristic is better than the previous heuristic.  While we are trying to find a better heuristic, this PR suggests to switch to a completely random heuristic.

### DIFF
--- a/src/common/Random.h
+++ b/src/common/Random.h
@@ -1,0 +1,22 @@
+//
+// Created by Antti Hyvarinen on 07.05.22.
+//
+
+#ifndef OPENSMT_RANDOM_H
+#define OPENSMT_RANDOM_H
+
+namespace opensmt {
+    // Returns a random float 0 <= x < 1. Seed must never be 0.
+    static inline double drand(double &seed) {
+        seed *= 1389796;
+        int q = (int) (seed / 2147483647);
+        seed -= (double) q * 2147483647;
+        return seed / 2147483647;
+    }
+
+    // Returns a random integer 0 <= x < size. Seed must never be 0.
+    static inline int irand(double &seed, int size) {
+        return (int) (drand(seed) * size);
+    }
+}
+#endif //OPENSMT_RANDOM_H
--- a/src/common/Random.h
+++ b/src/common/Random.h
@@ -1,6 +1,9 @@
-//
-// Created by Antti Hyvarinen on 07.05.22.
-//
+/*
+ * Copyright (c) 2022, Antti Hyvarinen <antti.hyvarinen@gmail.com>
+ * Copyright (c) 2003-2006, Niklas Een, Niklas Sorensson
+ *
+ * SPDX-License-Identifier: MIT
+ */
 
 #ifndef OPENSMT_RANDOM_H
 #define OPENSMT_RANDOM_H
--- a/src/common/Random.h
+++ b/src/common/Random.h
@@ -8,6 +8,7 @@
 namespace opensmt {
     // Returns a random float 0 <= x < 1. Seed must never be 0.
     static inline double drand(double &seed) {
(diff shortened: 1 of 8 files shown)


src/common/Random.h
```diff
@@ -0,0 +1,22 @@
+//
+// Created by Antti Hyvarinen on 07.05.22.
+//
```
blishko, line 3: Use new license header :)

src/common/Random.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifndef OPENSMT_RANDOM_H
+#define OPENSMT_RANDOM_H
+
+namespace opensmt {
+    // Returns a random float 0 <= x < 1. Seed must never be 0.
+    static inline double drand(double &seed) {
+        seed *= 1389796;
```
blishko, line 11: Maybe add `assert(seed != 0)`?

src/tsolvers/lasolver/LASolver.h
```diff
@@ -151,8 +152,8 @@ class LASolver: public TSolver
     LVRef getVarForTerm(PTRef ref) const  { return laVarMapper.getVarByPTId(logic.getPterm(ref).getId()); }
     void notifyVar(LVRef);                             // Notify the solver of the existence of the var. This is so that LIA can add it to integer vars list.
 
-    // Most-infeasible branching heuristic
-    LVRef splitOnMostInfeasible(vec<LVRef> const &) const;
+    // Random splitting heuristic
+    LVRef splitOnRandom(vec<LVRef> const &);
```
blishko, line 156:
```suggestion
    LVRef splitOnRandom(vec<LVRef> const &) const;
```
This can be `const`, right?

blishko (commented):
LGTM!
Pulling the random functions to a separate header is a good thing!
There are just a few tiny things.

src/common/Random.h
aehyvari: Technically it's not my code, it's probably from MiniSat.  I'm not sure what to say there...  Should I put Minisat's license and Een & Sorenson as authors, addition to me?

src/common/Random.h
aehyvari: OK.

src/tsolvers/lasolver/LASolver.h
aehyvari: It changes `LASolver::seed`.

blishko:
@aehyvari, how about we use some standard C/C++ utilities for the random number generation? Is there any reason why we should stick to the functions from MiniSAT?

blishko:
Or if that is too complicated, maybe we can stick to this for now. I don't know about the license in this case. If you want you can attribute it to the authors of MiniSAT. Although, then you could argue that we should do it for more files, such as `CoreSMTSolver`...

blishko (changes requested):
The `#include Random.h` has been left in the `CoreSMTSolver.h`

src/smtsolvers/CoreSMTSolver.h
```diff
@@ -62,6 +62,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #include "SolverTypes.h"
 
 #include "Timer.h"
+#include "Random.h"
```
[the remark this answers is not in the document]
aehyvari: Ok.

blishko (approved):
LGTM!

PR #485, opened 2022-05-09
Adding verbose option

msooseth:
Separate PR for verbose option adding

### DIFF
--- a/src/smtsolvers/CoreSMTSolver.cc
+++ b/src/smtsolvers/CoreSMTSolver.cc
@@ -52,6 +52,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 
 #include <cmath>
 #include <iostream>
+#include <iomanip>
 #include <algorithm>
 
 #include "Proof.h"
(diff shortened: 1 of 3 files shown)


src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -1449,6 +1450,19 @@ lbool CoreSMTSolver::search(int nof_conflicts)
...
+                else units = trail_lim[0];
+
+                if (verbosity)
+                    std::cout << "; conflicts: " << std::setw(5) << std::round(conflicts/1000.0) << "K"
+                    << " learnts: " << std::setw(5) << std::round(learnts.size()/1000.0) << "K"
+                    << " clauses: " << std::setw(5) << std::round(clauses.size()/1000.0) << "K"
+                    << " units: " << std::setw(5) << units
+                    << std::endl;
+            }
+
```
blishko, line 1465:
```suggestion
            if (verbosity and conflicts % 1000 == 999) {
                uint64_t units = trail_lim.size() == 0 ?  trail.size() :  trail_lim[0];

                    std::cout << "; conflicts: " << std::setw(5) << std::round(conflicts/1000.0) << "K"
                    << " learnts: " << std::setw(5) << std::round(learnts.size()/1000.0) << "K"
                    << " clauses: " << std::setw(5) << std::round(clauses.size()/1000.0) << "K"
                    << " units: " << std::setw(5) << units
                    << std::endl;
            }

```

If I am reading this correctly, if the verbosity is off then the whole block does not really do anything, so the check can be moved to the first `if`.
Also, the logic for computing `units` can be simplified.

blishko (commented):
I think this is useful. What do you think @aehyvari ?

aehyvari:
Yes, I also think this is good.

msooseth:
BTW, visibility into the workings of solvers is one of the most important things to understand & potentially improve it. If you take a look at [kissat](https://github.com/arminbiere/kissat), it has ridiculous amounts of code just for statistics generation & printing. While I try to preach this, I am far-far away from kissat in my own solver. However, I'd suggest checking some of the printing & statistics techniques that Armin is using. It's quite good.

PR #486, opened 2022-05-09
Update glue on analyze

msooseth:
Updating glue during analysis. This is default for most SAT solvers.

### DIFF
--- a/src/minisat/core/SolverTypes.h
+++ b/src/minisat/core/SolverTypes.h
@@ -167,11 +167,6 @@ class Clause {
                 calcAbstraction(); }
     }
 
-    void setGlue(const uint32_t glue) {
-        assert(glue < 8);
-        header.glue = glue;
-    }
-
 public:
     void calcAbstraction() {
         assert(header.has_extra);
@@ -208,6 +203,10 @@ public:
     void         strengthen  (Lit p);
     uint32_t     getGlue() const {
         return header.glue;
+    }
+    void setGlue(const uint32_t glue) {
+        assert(glue < 8);
+        header.glue = glue;
     }
 };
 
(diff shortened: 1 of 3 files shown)


aehyvari:
I'll update the pictures here again.

`QF_UF`

non-incremental-QF_UF-b8597d1e-2022-05-01_vs_d8850475-2022-05-09

aehyvari:
We discussed this PR a little with @blishko.  We decided to rebase this on `master`, run the once again against the current master using `QF_LIA` to see if there are surprises, and if nothing shows up, we'll put it in!

blishko (approved):
Yes, I think this is good. If `QF_LIA` on the current master looks good, we should merge this.

aehyvari:
Here's the result against a version that is for the relevant parts equal to the current master, on `QF_LIA`

non-incremental-QF_LIA-2ecc5464-2022-05-06_vs_83f11053-2022-05-11

The x-axis branch is [here](https://github.com/aehyvari/OpenSMT2/commit/2ecc54643d5b145ff5fa6b2698b5fc4185a4f160)

blishko:
Looks good! I like that now we do not lose the UNSAT instance which master can solve in 10 seconds anymore :)
Even though it now needs more than 300 seconds.

But overall the results are good, so I think we can merge this. I have already approved it.

msooseth:
BTW, instead of going piecemeal, I could try to just replace the whole SAT solver you got going with CMS :laughing: I mean, it would be a bit crazy but also probably quite a bit faster... Maybe I'll take a look over the weekend, see what changes you have made to MiniSat and how they could be ported to CMS. Note that CMS is already a CDCL(T) framework due to Gauss-Jordan elimination, so it's not supposed to be radically different...

msooseth:
`I like that now we do not lose the UNSAT instance which master can solve in 10 seconds anymore` -- that points to some more fundamental problem, IMHO. UNSAT instances are not supposed to be too variable in solving time. With correct heuristics, they tend to be within 20% solving time, even when you change things quite a bit. Such huge variance, in my opinion, points to some other issue, perhaps to do with the theory part, perhaps some other unexpected interaction between CDCL <-> T.

blishko:
I believe this big difference is definitely related to our Arithmetic solver, which is still not perfect and there are cases where it will diverge, or at least take a long time to decide the theory consistency.

PR #487, opened 2022-05-09
Vivify regularly

msooseth:
Regularly vivify learnt clauses. This follows: https://www.ijcai.org/Proceedings/2017/0098.pdf This is the paper usually referred to by Armin Biere (there are some... others, but they shall be banished from the face of the Earth :laughing: )

### DIFF
--- a/src/minisat/core/SolverTypes.h
+++ b/src/minisat/core/SolverTypes.h
@@ -142,7 +142,8 @@ class Clause {
         unsigned has_extra : 1;
         unsigned reloced   : 1;
         unsigned glue      : 3;
-        unsigned size      : 24; }                            header;
+        unsigned vivifed   : 1;
+        unsigned size      : 23; }                            header;
     union { Lit lit; float act; uint32_t abs; CRef rel; } data[0];
 
     friend class ClauseAllocator;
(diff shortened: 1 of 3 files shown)


msooseth:
From the paper: ```e evaluated our approach on five of the best perform-
ing CDCL SAT solvers: Glucose [Audemard and Simon,
2009], COMiniSatPS [Oh, 2016], MapleCOMSPS [Liang et
al., 2016c], MapleCOMSPS LRB [Liang et al., 2016b] and
TC Glucose [Moon and Mary, 2016]. The experimental re-
sults show that the proposed approach allows them to solve a
remarkable number of additional instances coming from the
hard combinatorial and application categories of the 2014 and
2016 SAT competitions```

Note that this is actually really true, I'm getting way more instances solved with vivification in CryptoMiniSat. In fact, Armin used to consider vivification the first inprocessing technique adopted by SAT solvers, before kissat destroyed all other SAT solvers.

msooseth:
Note, I'm not doing line 8 on Algorithm 2 in that paper -- analysis can be tricky with theory clauses: reasons may need to be computed. But it could be done, and it may be slightly (but only slightly) more effective, if (and likely only if) the theory reasons are cheap to compute.

blishko:
Unlike the other changes in the SAT solver, which were very helpful, vivification does not seem to positive impact for OpenSMT on SMT-LIB benchmarks.

blishko:
OK, I agree we should definitely revisit this, but I also don't have time at the moment.
Thanks for the all the suggested changes and pointers to the relevant papers!

PR #488, opened 2022-05-09
Clause iterators

aehyvari:


test/unit/test_SATSolverTypes.cc
```diff
@@ -0,0 +1,25 @@
+//
+// Created by prova on 09.05.22.
+//
```
blishko, line 3: Use new license? :)

blishko (approved):
LGTM!

test/unit/test_SATSolverTypes.cc
aehyvari: ok 🤓

PR #489, opened 2022-05-09
ArithLogic: Fix the LANonlinearException logic

MasoudAsadzade:
LANonlinearException was previously written weirdly.  This commit
suggests an improvement that is hopefully more standard.

Co-authored-by: Antti Hyvarinen <antti.hyvarinen@gmail.com>

### DIFF
--- a/src/logics/ArithLogic.h
+++ b/src/logics/ArithLogic.h
@@ -6,18 +6,14 @@
 
 class LANonLinearException : std::runtime_error
 {
+    std::string msg;
 public:
     LANonLinearException(const char* reason_) : runtime_error(reason_) {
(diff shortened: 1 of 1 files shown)


blishko (approved):
LGTM!

Reading the documentation on `std::runtime_error`, maybe `LANonLinearException` (and `ArithDivisionByZeroException` should not be derived from `std::runtime_error`.
But I don't have much experience with exceptions.

aehyvari:
Me neither... Let's go with this one and fix them later. This was the first and necessary step towards a preprocessor header which would allow us to throw instead of crashing with asserts. Once @MasoudAsadzade has it tested on smts he'll prepare a PR for that one. It  could also include the `release_assert` from #487 (if we decide to include it).

PR #490, opened 2022-05-10
Assert throws

MasoudAsadzade:
This PR includes:

A common wrapper for assert
This commit introduces a layer that either takes `assert` from `cassert`, or depending on a new `cmake` option `OSMT_ASSERT_THROWS`, optionally defines `assert` as something that throws.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -19,12 +19,13 @@ target_sources(common
     "${CMAKE_CURRENT_LIST_DIR}/PartitionInfo.cc"
     "${CMAKE_CURRENT_LIST_DIR}/VerificationUtils.cc"
     "${CMAKE_CURRENT_LIST_DIR}/ReportUtils.h"
+    "${CMAKE_CURRENT_LIST_DIR}/OpensmtAssert.h"
 )
 
 install(FILES
         Integer.h Number.h FastRational.h XAlloc.h Alloc.h StringMap.h Timer.h osmtinttypes.h
-        TreeOps.h Real.h FlaPartitionMap.h PartitionInfo.h OsmtApiException.h TypeUtils.h 
-        NumberUtils.h NatSet.h
+        TreeOps.h Real.h FlaPartitionMap.h PartitionInfo.h OsmtApiException.h TypeUtils.h
+        NumberUtils.h NatSet.h OpensmtAssert.h
 DESTINATION ${INSTALL_HEADERS_DIR})
 
 
--- a/src/common/OsmtAssertException.h
+++ b/src/common/OsmtAssertException.h
@@ -1,6 +1,9 @@
-//
-// Created by prova on 13.05.20.
-//
+/*
+ * Copyright (c) 2022, Antti Hyvarinen <antti.hyvarinen@gmail.com>
+ * Copyright (c) 2022, Masoud Asadzade <masoud.asadzade@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
 
 #ifdef OSMT_ASSERT_EXCEPTION_H
 #else
--- a/src/common/OsmtAssertException.h
+++ b/src/common/OsmtAssertException.h
@@ -1,0 +1,18 @@
+//
+// Created by prova on 13.05.20.
+//
+
+#ifdef OSMT_ASSERT_EXCEPTION_H
+#else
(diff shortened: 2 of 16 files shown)


aehyvari:
One of the commits was already in master (since a few minutes...) so I removed that one.

I also fixed the license :)

src/common/OsmtAssertException.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2022, Antti Hyvarinen <antti.hyvarinen@gmail.com>
+ * Copyright (c) 2022, Masoud Asadzade <masoud.asadzade@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#ifdef OSMT_ASSERT_EXCEPTION_H
+#else
```
blishko, line 9:
```suggestion
#ifndef OSMT_ASSERT_EXCEPTION_H
```

src/common/Timer.h
```diff
@@ -114,14 +114,12 @@ class StopWatch {
...
             time_start = OSMTTimeVal(tmp_rusage);
         }
-        else assert(false);
     }
     ~StopWatch()
     {
         if (getrusage(RUSAGE_SELF, &tmp_rusage) == 0) {
             timer += OSMTTimeVal(tmp_rusage) - time_start;
         }
-        else assert(false);
```
blishko, line 124: Why remove these `asserts`? If you are removing them you do not need to include the header, no?

src/minisat/mtl/BoxedVec.h
```diff
@@ -21,7 +21,7 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 #define BoxedVec_h
 
 #include <cstdlib>
-#include <cassert>
```
blishko, line 24: I don't think we are using this file, no? So it should probably not be changed.

src/common/CMakeLists.txt
```diff
@@ -19,12 +19,13 @@ target_sources(common
...
     "${CMAKE_CURRENT_LIST_DIR}/ReportUtils.h"
+    "${CMAKE_CURRENT_LIST_DIR}/OpensmtAssert.h"
 )
 
 install(FILES
         Integer.h Number.h FastRational.h XAlloc.h Alloc.h StringMap.h Timer.h osmtinttypes.h
-        TreeOps.h Real.h FlaPartitionMap.h PartitionInfo.h OsmtApiException.h TypeUtils.h 
-        NumberUtils.h NatSet.h
+        TreeOps.h Real.h FlaPartitionMap.h PartitionInfo.h OsmtApiException.h TypeUtils.h
+        NumberUtils.h NatSet.h OpensmtAssert.h
```
blishko, line 28: I think you need to include `OpenSMTAssertException.h` and `OpenSMTInternalException.h` otherwise OpenSMT might not compile when included as a library in other projects.

blishko (commented):
I am not sure about this design. It seems that things can go wrong if OpenSMT is build with this option and installed as a library, but client code that will include OpenSMT headers will not build using this option.

I am not sure, but it seems there are some lurking dangers in this.

src/common/OsmtAssertException.h
MasoudAsadzade: From SMTS perspective,  It's working. Actually, SMTS'client just relies on OpenSMT to throw exceptions at asserts, so it catches them by base exception, so SMTS's client wouldn't need to include asserts.

src/common/CMakeLists.txt
MasoudAsadzade: SMTS's client just catches the base exception and reports the message to the SMTS server

src/common/Timer.h
MasoudAsadzade: It causes a warning to the destructor to have a implicit non-throwing exception specification

src/minisat/mtl/BoxedVec.h
MasoudAsadzade: Done.

src/common/CMakeLists.txt
aehyvari: Yes, I also think like @MasoudAsadzade that in this particular use case we're aiming at, there's no need to install the header.

PR #492, opened 2022-05-11
LASolver: Remove redundant storage of information

blishko:
`LAMapper` stores two mappings:
1. For each `Pterm` that corresponds to arithmetic term the corresponding LVRef (Simplex variable)
2. For each `Pterm` that corresponds to arithmetic inequality the LVRef of the arithmetic term that is compared to a constant.

The second mapping is redundant. Given the inequality we can extract the term from the inequality and just ask for the representation of that.

This does not makes a big difference in single query context, but in scenarios such as in Golem, where new terms are created all the time, the initialization of the LASolver can actually spend a lot of time in creating the second mapping. The reason is that the mapping is represented by a vector and the initialization must `push` as many times as is the `id` of the inequality. In long runs there can be many new `Pterms` created and the `id` of inequalities can be quite high which means a lot of `push`es.

The proposed change is to completely remove this vector and thus save potentially significant amount of time on initializing `LASolver`.

### DIFF
--- a/src/tsolvers/lasolver/LASolver.cc
+++ b/src/tsolvers/lasolver/LASolver.cc
@@ -279,8 +279,9 @@ std::unique_ptr<Tableau::Polynomial> LASolver::expressionToLVarPoly(PTRef term) 
 
 
 //
-// Get a possibly new LAVar for a PTRef term.  We may assume that the term is of one of the following forms,
-// where x is a real variable or ite, and p_i are products of a real variable or ite and a real constant
+// Registers an arithmetic Pterm (polynomial) with the solver.
+// We may assume that the term is of one of the following forms,
+// where x is a variable or ite, and p_i are products of a variable or ite and a constant
 //
 // (1) x
 // (2a) (* x -1)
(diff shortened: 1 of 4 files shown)


aehyvari:
Looks good! Can we change the signature from `LVRef LASolver::exprToLVar(PTRef expr)` -> `void LASolver::exprToLVar(PTRef expr)` and maybe use a more descriptive name, like `makeLVarForLeq`, or `mkLVar`?

blishko:
It looks like we can now, it does not look like the method is used anywhere else, beside the two changed places.
I was thinking if returning `LVRef` might be useful in some context, but I don't see a case against removing the returned value at the moment.
But I agree we need a good name for it.
One thing is that this method must be called with LA expressions (polynomial terms), not with LEQ inequalities.
Another is that if it starts with `mk` or `make`, then I think it should still return the `LVRef`.
`LASolver` has a method `getVarForTerm` which can be used once the term has been registered.
So I was thinking something like `registerTerm` or `registerExpression`.
I don't have a favorite at the moment.

blishko:
I changed the name to `registerArithmeticTerm`, still don't know what would be the best name.
I would also like to examine why this is called both when an atom is declared and when the solver is being initialized.
Feels like one of these is redundant. But I would leave this investigation for the future.

blishko:
And I would keep the returned LVRef. Somehow it feels better to do that (more consistent with the name of the method), even if we do not need it at the moment.
It costs almost nothing. Even if the mapping already exists, we need to do the lookup to check if simplex already knows about it. There may be a way that we would not need to do that, but I think that requires to go over and understand the whole process of initializing the LASolver, which I would postpone for later.

aehyvari (approved):
LGTM!

PR #494, opened 2022-05-15
Vivification

aehyvari:
I'm moving the discussion from #487 to here for now since it looks like managing the commit history is a little limited for me there.

### DIFF
--- a/src/smtsolvers/CoreSMTSolver.cc
+++ b/src/smtsolvers/CoreSMTSolver.cc
@@ -61,6 +61,29 @@ OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWA
 
 #include "Random.h"
 
+#if defined(_MSC_VER)
+#define release_assert(a) \
+    do { \
+    __pragma(warning(push)) \
+    __pragma(warning(disable:4127)) \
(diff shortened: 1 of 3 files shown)


src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -1364,7 +1387,118 @@ void CoreSMTSolver::learntSizeAdjust() {
...
+    //std::random_shuffle(vivif_tmp_lits.begin(), vivif_tmp_lits.end());
+    vivif_new.clear();
 
+    newDecisionLevel();
+
+    for (Lit l : vivif_tmp_lits) {
+        if (value(l) == l_True) {
+            cancelUntil(0);
+            detachClause(offs);
+            return false;
```
aehyvari, line 1461: I switched this from `true` to `false`, since it made sense to me that the clause would also be removed from the clause database if it gets detached.

aehyvari:
Here's a comparison between a recent master and [branch that is equal to the one in the PR](https://github.com/aehyvari/OpenSMT2/tree/vivification) for `QF_LIA`.  It looks like at least here vivification is not useful, in the form it is presented in this PR.

aehyvari:
I'm closing this one for now, since we have no evidence that it works on opensmt in the form suggested here.

PR #498, opened 2022-05-19
Adds linter to the CI

### LINKED ISSUE #479: Add linter so pushed code will be automatically checked on the codestyle rules
It would be goodto automatically check/fix any troubles with the codestyle in the repository
Preferrably standard Google [linter](https://github.com/cpplint/cpplint#readme)

BritikovKI:
Closes #479

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,17 +24,6 @@ jobs:
                   name: install model validation environment
                   command: ./regression_models/bin/setup-env.sh
             - run:
-                  name: Debug build gcc
-                  command: ./ci/run_travis_commands.sh
-                  environment:
-                      CMAKE_BUILD_TYPE: Debug
-                      ENABLE_LINE_EDITING: ON
-            - run:
-                  name: Release build gcc
-                  command: ./ci/run_travis_commands.sh
-                  environment:
-                      CMAKE_BUILD_TYPE: Release
-            - run:
                   name: Install clang
                   command: |
                       sudo apt-get install clang
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
             - run:
                   name: Run linter
                   command: |
-                      cpplint --recursive --filter=-build,-cmake-build-debug --help ./*
+                      cpplint --recursive --filter=-build,-cmake-build-debug ./*
 
     build-recent:
         docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,6 +193,10 @@ jobs:
 workflows:
   build-test:
       jobs:
+        - lint:
+          filters: &filters-build-test
+            tags:
+              only: /^v.*/
         - build-starexec:
           filters: &filters-build-test
             tags:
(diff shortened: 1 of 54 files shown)


BritikovKI:
@aehyvari @blishko Hey guys, here is a linter as a part of the CI, but it would be good if we define what exact rools do we want. Right now it founds 34000 problems with the code...

blishko:
Cool!
I would disable everything that prevents the CI from passing :D.
Then we can start fixing the issues one by one and gradually enable more and more checks.

.circleci/config.yml
```diff
@@ -1,5 +1,42 @@
...
+                  name: install dependencies
+                  command: sudo apt-get install libgmp-dev libedit-dev bison flex cmake libubsan0 python3-pip cython
+            - run:
+                  name: install model validation environment
+                  command: ./regression_models/bin/setup-env.sh
+            - run:
+                  name: Install clang
+                  command: |
+                      sudo apt-get install clang
+                      sudo update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++ 60
```
blishko, line 30: I think all these steps are not necessary and can be removed, no?

random_lookahead.smt2
```diff
@@ -0,0 +1,4 @@
+(set-option :pure-lookahead true)
```
blishko, line 1: These `smt2` files should probably not be part of the PR, no?

blishko (commented):
Left some comments.

.circleci/config.yml
BritikovKI: Actually I think that's right, not sure about clang, but we may check

random_lookahead.smt2
BritikovKI: Sure

aehyvari:
Yes, I wonder if there's an option for doing this gradually, for instance so that the linter could be run for new changes only...

Alternatively we could start small, doing this one directory or even file at a time, to get a feeling of what we want.

blishko:
Yes, my idea would be to disable everything and then start enabling files/directories one at a time as we clean the code.
An orthogonal problem is to settle on the set of rules we want to apply. I am pretty sure we don't want 100% of the default rules, we will need to customize them.
For the files/directories I understand we can create filters in the config file. Is there a similar mechanism for the rules?

MasoudAsadzade:
This is really great!
Maybe we can pass "git diff --name-only HEAD" to filter to only be applied on edited files for now.
BTW, in the future we can write a hook script to verify what is about to be committed. 
like following:
https://gist.github.com/shawndooley/1aee8217e75a323299cf6bd86ee5ac90

aehyvari:
So with all respect to the fine people behind `cpplint`, I'm not sure if this project is ready for production.  How about we try to use `clang-tidy`?  I found it from [this reddit thread](https://www.reddit.com/r/cpp/comments/5b397d/what_c_linter_do_you_use/).  I didn't have time yet to dig into it, but I wanted to bring this up asap.  Maybe there are other options around as well, for instance.

If you want to try it out, it's available through `homebrew`
```
$ brew install fmenezes/tap/clang-tidy
```

Let me know what you think!

aehyvari:
After reading a little more maybe the first step could be to use `clang-format`.

blishko:
If we can customize the rules, I would be fine with using `clang-*` family of tools.

PR #500, opened 2022-05-24
Do not use vectors indexed by Pterms' ids in LASolver

blishko:
This PR suggests to replace the inner representation of how LASolver maintains mappings from Pterms to its inner representations (LVRefs and bounds).
Instead of vectors indexed by PTId, I suggest to use just normal HashMap.
The disadvantage of current representation is its memory inefficiency, because the size of the vectors possibly depends on the number of terms in Logic. But Logic may contain a huge number of terms compared to a few atoms that need to be represented in LASolver.

For OpenSMT itself this does not make much difference, but for Golem, it can have noticeable impact, as `LASolver` is re-initialized on every check-sat call, and Golem makes quite a lot of them.

### DIFF
--- a/src/tsolvers/lasolver/LASolver.cc
+++ b/src/tsolvers/lasolver/LASolver.cc
@@ -210,12 +210,6 @@ const Delta LASolver::overBound(LVRef v) const
 }
 */
 
-void LASolver::setBound(PTRef leq_tr)
-{
-//    printf("Setting bound for %s\n", logic.printTerm(leq_tr));
-
-    addBound(leq_tr);
-}
 
 opensmt::Number LASolver::getNum(PTRef r) {
     return logic.getNumConst(r);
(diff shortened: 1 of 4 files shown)


blishko:
No, I wanted to try, but the cluster was not working today. I'll try later.
I also am a little worried, but this is not really critical code, so I would not expect the performance on single-queries to be much different. But I will try at least QF_LRA and post the results.
I started considering writing a specialized hashmap, but I hope it will not be necessary :)

I have a benchmark for Golem where this change meant more than 10% improvement.

aehyvari:
OK let's wait for the results.  One option is to run them in mantella, there are some scripts to do that and it might be faster than waiting for the cluster to come up.

blishko:
[comparison_master_maps.pdf](https://github.com/usi-verification-and-security/opensmt/files/8770741/comparison_master_maps.pdf)

It seems there is some slowdown associated with this.
Maybe I need implement custom hashmap :)

PR #501, opened 2022-05-24
ArithLogic: Avoid expensive call to mkTimes in mkNeg

blishko:
The method mkTimes is general a does a lot of work to ensure the term is properly
normalized. When we wnat to negate en xisting term, we know that it is
already normalized, so we can compute the proper result much cheaper
than passing multiplying it by -1 and letting mkTimes do the
simplifications.


aehyvari (approved):
LGTM!

PR #505, opened 2022-05-28
Benchmark: Add benchmarking of simple term creation

blishko:
This PR adds benchmarking tests for creation of conjunctions.
There are tests for
1. Creating a lot of binary conjunctions
2. Creating one larger conjunction

In both cases, I added tests evaluating the impact of negated terms and in the second case comparing an ordered arguments and randomly shuffled arguments.

### DIFF
--- a/benchmark/performance/perf_mkTerms.cc
+++ b/benchmark/performance/perf_mkTerms.cc
@@ -1,0 +1,135 @@
+/*
+ *  Copyright (c) 2022, Martin Blicha <martin.blicha@gmail.com>
+ *
+ *  SPDX-License-Identifier: MIT
+ *
+ */
+#include <benchmark/benchmark.h>
+#include <Logic.h>
+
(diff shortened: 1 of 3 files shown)


benchmark/performance/perf_mkTerms.cc
```diff
@@ -0,0 +1,135 @@
...
+ *  SPDX-License-Identifier: MIT
+ *
+ */
+#include <benchmark/benchmark.h>
+#include <Logic.h>
+
+#include <algorithm>
+#include <random>
+
+class RationalEfficiencyFixture : public ::benchmark::Fixture {
```
aehyvari, line 13: mkTerms

blishko:
I fixed the name of the Fixture. But now I see that here I am not using it at all, so I am not sure if it makes sense to keep it...

aehyvari (approved):
LGTM!

PR #506, opened 2022-06-02
benchmarking: first version of the scripts

aehyvari:
Scripts for comparing benchmarks.  I'll write later a readme, and we can think of something nicer.  The problem is that I want to store the benchmark binaries to be able to run them later on one go, but the benchmarks link on the opensmt library that is currently in the `build` directory.

### DIFF
--- a/benchmarking/compare.sh
+++ b/benchmarking/compare.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 workdir=$(cd $(dirname $0); pwd)/work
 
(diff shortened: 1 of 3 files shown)


benchmarking/compare.sh
```diff
@@ -0,0 +1,100 @@
...
+
+echo "Running benchmark $benchmark2 three times and saving the last" >&2
+for ((i=0; i < 3; i++)); do
+    LD_LIBRARY_PATH=$lib2 $benchmark2 --benchmark_format=csv \
+        |csvcut -c 'name,real_time' > $TMPDIR/bm2.csv
+done
+
+echo "set term pngcairo color"
+
+echo -n "set xtics ("
```
blishko, line 73: Apparently,
```
echo -n "text"
```

prints `-n text` on Mac Os. Can this be done without the option?

blishko: But not in the command line, only in the script?? Maybe I am doing something wrong...

aehyvari: The shell is wrong.  It should be `/bin/bash`.  I'll fix it.

aehyvari:
I added the readme (and a few other fixes), so if we're ok with this philosophy I think I don't have anything more in mind right now.

PR #509, opened 2022-06-04
Rewriting: Fix missing rewrites in rewriteMaxArity

blishko:
Previously, if rewriteMaxArity encountered a connective different than
conjunction or disjunction, it would automatically keep the term as it
was. This is not correct, because some rewriting might have already
happened in the subterms, so this term should also be rewritten
accordingly.

### DIFF
--- a/src/simplifiers/BoolRewriting.h
+++ b/src/simplifiers/BoolRewriting.h
@@ -77,19 +77,34 @@ PTRef rewriteMaxArity(Logic & logic, const PTRef root, T doNotRewrite) {
             else if (logic.isAtom(t[i])) {
                 assert(!cache.has(t[i]));
                 cache.insert(t[i], t[i]);
+            } else {
+                assert(false);
             }
         }
         if (unprocessed_children)
             continue;
 
         unprocessed_ptrefs.pop();
-        PTRef result = PTRef_Undef;
-        assert(logic.isBooleanOperator(tr));
+        SymRef symRef = t.symb();
+        assert(logic.isBooleanOperator(symRef));
 
-        if (logic.isAnd(tr) || logic.isOr(tr)) {
+        PTRef result = PTRef_Undef;
+        if (logic.isAnd(symRef) or logic.isOr(symRef)) {
             result = ::mergeAndOrArgs(logic, tr, cache, doNotRewrite);
-        } else {
-            result = tr;
+        } else if (logic.isNot(symRef)) {
+            PTRef child = t[0];
+            PTRef newChild = cache[child];
+            result = child == newChild ? tr : logic.mkNot(newChild);
+        } else { // general connective
+            vec<PTRef> newArgs;
+            newArgs.capacity(t.size());
+            bool changed = false;
+            for (PTRef child : t) {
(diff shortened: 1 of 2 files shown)


src/simplifiers/BoolRewriting.h
```diff
@@ -77,19 +77,34 @@ PTRef rewriteMaxArity(Logic & logic, const PTRef root, T doNotRewrite) {
...
-        if (logic.isAnd(tr) || logic.isOr(tr)) {
+        PTRef result = PTRef_Undef;
+        if (logic.isAnd(symRef) or logic.isOr(symRef)) {
             result = ::mergeAndOrArgs(logic, tr, cache, doNotRewrite);
-        } else {
-            result = tr;
+        } else if (logic.isNot(symRef)) {
+            PTRef child = t[0];
+            PTRef newChild = cache[child];
+            result = child == newChild ? tr : logic.mkNot(newChild);
```
blishko, line 97: I separated negation as the most common case (other than conjunction and disjunction) to try to avoid the temporary vector.

aehyvari:
Looks good!  Did you check if this improves the performance?

blishko:
I don't remember :)
Some time has passed since I created this, I'll have to refresh my memory what exactly is the purpose here :)

aehyvari:
Right, I think the competition took all my energy at that time...

blishko:
I don't think I ran performance test, I will try to run the benchmark and see if there is any difference.

blishko:
And here is LRA:
non-incremental-QF_LRA-master-2022-08-30_vs_rewrite-2022-08-30

blishko:
It does not seem to produce performance benefit. There is even ~1% slowdown in UF, which could also be a noise...

blishko:
QF_LIA is much more affected, but the effect still seems quite random.
This is the results with 300s timeout.
Maybe I should re-run it with 1200s timeout?

non-incremental-QF_LIA-master-2022-08-30_vs_rewrite-2022-08-30

blishko:
Here is QF_LIA with 20 minutes timeout. The overall impression is the same as from 5 minutes timeout.
The effect of this PR seems like a random noice.
I think it should be OK to merge.
Although it does not seem to produce any benefit on SMT-LIB benchmarks, conceptually we should make this change.

non-incremental-QF_LIA-master-2022-08-31_vs_rewrite-2022-09-02

aehyvari:
I tend to agree.  We can merge.

aehyvari (approved):
LGTM!

blishko:
Just a note:
One benchmark that was lost is `QF_LIA/convert/convert-jpg2gif-query-1491.smt2`.
Previously it was solved under 1 second.
Testing random seeds from 1 to 20, this benchmark is solved, it was not solved 3 times out of 20 with time limit 10 seconds. In most cases it is still solved under 1 second.
With the default seed, it seems the tableau representation gets complex (the coefficients grow large enough that they must be represented with GMP) and checking the satisfiability of the relaxed problem takes a lot of time.

We may try to normalize the tableau, see if we can maybe keep the better representation of the coefficients.
But it is not clear if that will have the desired effect.

PR #510, opened 2022-06-07
Linter clang

BritikovKI:


.clang-format
```diff
@@ -0,0 +1,174 @@
...
+BreakBeforeConceptDeclarations: true
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit: 80
```
blishko, line 59: I would suggest to be more generous and allow 100 or even 120 line width.

blishko (commented):
Nice!
Should we maybe start with something smaller than `CoreSMTSolver` before we tune the rules to our needs? `TermMapper` seems OK to me.

.clang-format
```diff
@@ -0,0 +1,174 @@
+---
+Language: Cpp
+# BasedOnStyle:  LLVM
```
blishko, line 3: Reading the documentation, we could actually activate this entry and then we do not need most of the other entries. Instead we can specify those values that we want to change. It seems to me that would be more understandable.

.clang-format
BritikovKI: Done

.clang-format
BritikovKI: Fixed

blishko: What I meant to say is that now it should be possible to remove all the other entries in the `.clang-format` file, no?
Except the changed `ColumnLimit`. And whatever else we want to customize.

BritikovKI: Hmm, understood, I think done

MasoudAsadzade:
It would be great if we provide some brief description to our pull requests, so we can get/catch up to the subject fast.

aehyvari (commented):
As discussed: 
 * let's use 4 spaces for indentation
 * remove `CoreSMTSolver.cc` for now as a formatted thing (it's a bit too big...), 
 * `lookahead.smt2` doesn't have anything to do with this so should be removed.

aehyvari (approved):
LGTM!

PR #513, opened 2022-06-10
Fix handling of scoped define-funs

### LINKED ISSUE #512: Function definition not cleared after pop
Consider the following example:
```
(set-logic QF_UF)
(declare-sort U 0)
(declare-fun x () U)
(declare-fun f (U) U)
(push 1)
(define-fun a1 () Bool (not (= x (f x))))
(assert a1)
(set-info :status sat)
(check-sat)
(pop 1)
(define-fun a1 () Bool false)
(assert a1)
(set-info :status unsat)
(check-sat)
(exit)
```

The function `a1` is defined in each frame. This is OK, because the first definition should be removed with the `pop`. However, OpenSMT does not clear the defined functions, reports error on the second `define-fun` and uses the original definition which leads to an incorrect answer.

blishko:
This PR proposes to move the handling of defined functions from `Logic` to `Interpret` and treat them basically as macros when interpreting SMT-LIB2 file.
In the `Interpret`, we can tie their lifetime to the push frame where they were introduced and remove them on `pop`.

Fixes #512.


src/bin/Interpret.cc
```diff
@@ -971,6 +972,13 @@ bool Interpret::defineFun(const ASTNode& n)
     return rval;
 }
 
+bool Interpret::storeDefinedFun(std::string const & fname, const vec<PTRef> & args, SRef ret_sort, const PTRef tr) {
+    if (defined_functions.has(fname)) { return false; }
+
+    defined_functions.insert(fname, TemplateFunction(fname, args, ret_sort, tr), true); // TODO: ask config if defined are scoped
```
blishko, line 978: The standard introduced option `:global-declarations`, but I don't think we have it yet. We need to add it to the config.

src/bin/Interpret.h
```diff
@@ -32,9 +32,36 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
 #include <unordered_map>
 
+class DefinedFunctions {
+    std::unordered_map<std::string,TemplateFunction> defined_functions;
+    opensmt::ScopedVector<std::string> scopedNames;
+
+public:
+    bool has(const std::string& name) const { return defined_functions.find(name) != defined_functions.end(); }
+
+    void insert(const std::string& name, TemplateFunction && templ, bool scoped = false) {
```
aehyvari, line 46: Check for west-constness

src/bin/Interpret.h
```diff
@@ -32,9 +32,36 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+        assert(not has(name));
+        defined_functions[name] = std::move(templ);
+        if (scoped) { scopedNames.push(name); }
+    }
+
+    TemplateFunction & operator[](const char* name) {
```
aehyvari, line 52: Check for west-constness

aehyvari (approved):
LGTM.  I left two small style comments.

blishko:
I fixed the style and added also the new option `:global-declarations`.
There is also a new test to exercise this option.
@aehyvari , do you want to check again?

src/bin/Interpret.h
blishko: Fixed

src/bin/Interpret.h
blishko: Fixed

aehyvari:
Ok I'll have a quick look!

aehyvari (approved):
Looks good!  For now we don't have the problem with the declarations being global but let's think about that later then...

Issue #518, opened 2022-06-17
Performance improvement on QF_LIA benchmark


blishko:
Performance profile of OpenSMT on `QF_LIA/20220307-SMPT/SharedMemory-PT-000100/RC-08.smt2` shows several bottlenecks on this benchmark, both in preprocessing and solving.

I believe the core of the problem is that this benchmark encodes all intermediate sums of big sums with helper variables and OpenSMT processes them gradually, computing the substitutions, creating the final large sums, but then adding all the intermediate sums back to the formula, because we add all the substitutions back to the formula (to not miss inconsistencies of LIA equalities and to easily compute model even for variables that are substituted away).
This unfortunately duplicates a lot of work OpenSMT is doing both in preprocessing and in solving.

The encoding is unfortunate for OpenSMT (I believe`let`s would be sufficient and better for OpenSMT), but we can try to improve the handling of such benchmarks nevertheless.

One idea that I have is that we are probably not computing the substitutions in an optimal way, which leads to more iterations and more rewriting in the substitution code. We could try to improve that.

PR #519, opened 2022-06-17
ArithLogic: Handle multiplication by -1 as negation

blishko:
Since OpenSMT uses multiplication by -1 to express negated arithmetic
terms, this operation can occur quite often and is repeated for example
when applying substitutions.
Multiplication by -1 can be done more efficiently by negating the second
argument (as opposed to full mkTimes run).

### DIFF
--- a/src/logics/ArithLogic.cc
+++ b/src/logics/ArithLogic.cc
@@ -356,9 +356,8 @@ PTRef ArithLogic::mkNeg(PTRef tr)
         auto sortRef = getSortRef(symref);
         return mkFun(getTimesForSort(sortRef), {tr, getMinusOneForSort(sortRef)});
     }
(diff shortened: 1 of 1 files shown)


blishko:
@aehyvari , this shaved off a minute and a half from the computation of substitution in the benchmark from #518.

It would be good to test at least QF_LRA benchmark suite to check that this does not have any undesirable side effect.

blishko:
[comparison_run2.pdf](https://github.com/usi-verification-and-security/opensmt/files/8940527/comparison_run2.pdf)

Here is the comparison with master on QF_LRA.
Should be a safe change.

As I mentioned before, this does help with preprocessing on #518 somewhat, even though I think much better improvement can be achieved if we re-design the substitution code.

aehyvari (approved):
LGTM!

PR #523, opened 2022-06-20
TermVisitor: make top-level visit virtual

aehyvari:
In some use cases it is useful to perform special actions at top-level
visit, such as resetting the state of the config.  This commit suggests
to turn TermVisitor's top-level visit function into a virtual function
so that it can be fluidly overridden if needed.


aehyvari:
I'll bundle the use case together with the PR.

MasoudAsadzade:
I can only see one commit with only a "virtual " added ?

PR #525, opened 2022-06-22
CMakeLists-parallel: Print stats of the PTPLib in color

MasoudAsadzade:
This PR,
print stats of the PTPLib in color, it'd help to track compilation process easily on ci, aws through docker,...

BTW, I was just tracking compilation process and here is how it looks like:

Screenshot 2022-06-22 at 16 01 07


blishko (approved):
Looks good!

Please rebase on top of current master.

PR #526, opened 2022-06-23
Interpret: Fix printing of models for functions

blishko:
Logic::pp should not be used for output to the user.
It prints rationals in infix format, but models should be printed in SMT-LIB format (using Logic::printTerm).


aehyvari (approved):
LGTM!

PR #528, opened 2022-07-07
Array solver style fixes

aehyvari:
Some suggestions on the array PR

### DIFF
--- a/src/tsolvers/axdiffsolver/ArraySolver.cc
+++ b/src/tsolvers/axdiffsolver/ArraySolver.cc
@@ -220,9 +220,7 @@ void ArraySolver::merge(ERef storeTerm) {
     if (arrayNode == storeNode) { return; }
     makeWeakRepresentative(arrayNode);
     if (getRepresentative(storeNode) == arrayNode) {
-        Map<ERef, bool, ERefHash> forbiddenIndices;
-        forbiddenIndices.insert(indexTerm, true);
-        mergeSecondary(storeNode, arrayNode, storeTerm, forbiddenIndices);
+        mergeSecondary(storeNode, arrayNode, storeTerm, std::unordered_set<ERef,ERefHash>({indexTerm}));
     } else {
         // new primary edge
         getNode(arrayNode).primaryEdge = storeNode;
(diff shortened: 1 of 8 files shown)


src/tsolvers/axdiffsolver/ArraySolver.cc
```diff
@@ -432,104 +424,89 @@ PTRef ArraySolver::computeExtensionalityClause(NodeRef n1, NodeRef n2) {
...
-                        PTRef equalityOfIndices = getEquality(index, storeIndex);
+                        PTRef equalityOfIndices = getEquality(index, storeIndex, logic);
                         if (not isFalsified(equalityOfIndices)) {
                             assert(not isSatisfied(equalityOfIndices));
-                            lemmaConditions.undecidedEqualities.insert(equalityOfIndices);
+                            undecidedEqualities.insert(equalityOfIndices);
                         }
                     }
-                    this->lemmas.push_back(std::move(lemmaConditions));
+                    lemmas.emplace_back(LemmaConditions{equalityOfSelects, std::move(undecidedEqualities)});
```
blishko, line 470:
```suggestion
                    lemmas.emplace_back(equalityOfSelects, std::move(undecidedEqualities));
```

blishko (commented):
LGTM!
I spotted one thing that I believe can be improved.

Also, I would prefer the collection for collecting indices and explanations to be past as the last parameter. What is the reason to have it as the first parameter?

src/tsolvers/axdiffsolver/ArraySolver.cc
blishko: OK, it looks like this does not work.
Then probably revert this and I'll accept your previous version.

aehyvari:
> Also, I would prefer the collection for collecting indices and explanations to be past as the last parameter. What is the reason to have it as the first parameter?

I was reading the discussion in [Beautiful c++](https://www.oreilly.com/library/view/beautiful-c-30/9780137647767/) about the CG F.21 on "out" values.  The book seems to argue that this construct is in some sense an extension of the API of the class being passed as a reference argument.  While it doesn't make sense in this case (and the example case in the book) to actually extend the class API, it's still logical to pass the object as the first argument, as it would be done in the low level, in `python`, or, say, if we were trying to simulate object-oriented coding in `C`.  I thought this was a nice way of seeing it, so I included it into the suggestions.  But I'm not fundamentalist-religious about it obviously :)

src/tsolvers/axdiffsolver/ArraySolver.cc
aehyvari: Damn, that didn't seem to work...

blishko:
> I was reading the discussion in [Beautiful c++](https://www.oreilly.com/library/view/beautiful-c-30/9780137647767/) about the CG F.21 on "out" values. The book seems to argue that this construct is in some sense an extension of the API of the class being passed as a reference argument

OK, I am familiar with this, but I am not sure it matches this situation. Anyway, we can go with your changes, and get back to this discussion someday in the future :)

blishko (approved):
LGTM!

PR #532, opened 2022-07-19
Arrays: Fix missing literals from extensionality lemma

blishko:
When explaining weak congruence between two array terms, we were previously missing an explanation why the considered index is equivalent to its root (if it is not the root itself).
This resulted in producing invalid theory lemmas, which were missing some literals. These literals are required to express that that the lemma is valid only under the circumstance that some indices are equivalent.


aehyvari (approved):
LGTM!

PR #534, opened 2022-07-22
Extensionality lemmas for arrays without selects

aehyvari:
The current code might crashes on an assertion if an instance contains array symbols that do not appear in `select`s.  This seems to be fixable by turning the assertion into a conditional case.  This PR contains the suggested change and related regression tests.


blishko (approved):
LGTM!

PR #536, opened 2022-07-25
Logic: Fix unintialized sort symbol

blishko:
After updating OpenSMT to latest release in Golem, I got some unit tests failing at a very weird place.
I tracked down the problem to uninitialized value of sort symbol `sym_ArraySort` (when `Logic` does *not* support arrays).

The code was not consistent before because it only creates array sort if the Logic does not have arrays, but the check for array sort in `Logic::getSort` is unconditional.
Thus, it is a check against uninitialized value.

This PR fixes to problem by initializing the symbol sort to `SSymRef_Undef` if the logic does not support arrays.

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -82,9 +82,7 @@ Logic::Logic(opensmt::Logic_t _logicType) :
     sortToEquality.insert(sort_BOOL, sym_EQ);
     sortToDisequality.insert(sort_BOOL, sym_DISTINCT);
     sortToIte.insert(sort_BOOL, sym_ITE);
-    if (hasArrays()) {
-        sym_ArraySort = sort_store.newSortSymbol(SortSymbol("Array", 2, SortSymbol::INTERNAL));
-    }
+    sym_ArraySort = sort_store.newSortSymbol(SortSymbol("Array", 2, SortSymbol::INTERNAL));
 }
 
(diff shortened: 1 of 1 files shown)


src/logics/Logic.cc
```diff
@@ -82,9 +82,7 @@ Logic::Logic(opensmt::Logic_t _logicType) :
     sortToEquality.insert(sort_BOOL, sym_EQ);
     sortToDisequality.insert(sort_BOOL, sym_DISTINCT);
     sortToIte.insert(sort_BOOL, sym_ITE);
-    if (hasArrays()) {
-        sym_ArraySort = sort_store.newSortSymbol(SortSymbol("Array", 2, SortSymbol::INTERNAL));
-    }
+    sym_ArraySort = hasArrays() ? sort_store.newSortSymbol(SortSymbol("Array", 2, SortSymbol::INTERNAL)) : SSymRef_Undef;
```
blishko, line 85: I would actually prefer to always create this sort properly, even though this suggested fix is more consistent with the philosophy that SMT-LIB logic type dictates which terms are possible to create.

aehyvari (approved):
Good!  I don't mind if we always initialise it correctly, it makes the code simpler as well.

blishko:
OK, I'll make it unconditional, then!

PR #537, opened 2022-07-26
ITE: Use appropriate memory in NodeAllocator

blishko:
I believe previously `iteNodeWord32Size` was computing the size in bytes, not size in 32-bit chunks.


aehyvari (approved):
Good that you noticed!

PR #540, opened 2022-07-30
Fix problems in splitting

aehyvari:
This PR contains fixes for the splitting system used in the build for parallel opensmt, a.k.a. SMTS:
- The current splitting does not work as intended.  This PR suggests a fix that correctly excludes the already covered search space in the splits.  It also fixes some related regression tests.
- The splitting is unsound.  The ite terms are shared using auxiliary variables, whose names constructed from PTRefs.  These are not guaranteed to be same even within the same build of opensmt.  As a result, SMTS can show satisfiable instances unsatisfiable.  The fix suggested in the PR is to rewrite the auxiliary variables using the original terms.


aehyvari:
I will split this PR into smaller parts.

PR #541, opened 2022-08-01
Rewriters: Apply clang-format

blishko:
Reformat more files in our codebase.


blishko:
@aehyvari , please check if you agree with these formatting changes, or we should adjust some of the rules.

aehyvari (approved):
LGTM!

PR #543, opened 2022-08-18
STPSolver: hack to support incrementality

MasoudAsadzade:
This PR will fix an Assertion failed: (e != EdgeRef_Undef), function assertLit, file src/tsolvers/stpsolver/STPSolver_implementations.hpp in QF_RDL and QF_IDL instances.
Co-authored-by: Antti Hyvärinen <antti.hyvarinen@gmail.com>

### DIFF
--- a/src/tsolvers/TSolver.h
+++ b/src/tsolvers/TSolver.h
@@ -185,7 +185,7 @@ public:
     virtual void  informNewSplit(PTRef) { };
     virtual Logic& getLogic() = 0;
     virtual bool isValid(PTRef tr) = 0;
-    bool         isKnown(PTRef tr);
+    virtual bool isKnown(PTRef tr);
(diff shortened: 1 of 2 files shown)


blishko:
What problem is this fixing exactly?
This is breaking the invariant we have been maintaining so far between the methods `setKnown` and `isKnown` in `TSolver`.
It does not look like a good idea to me...

aehyvari:
See issue #544

aehyvari:
This fix is addressed in #546, now in master.  As soon as the experiments do not depend on this PR, we should close this.

blishko:
We can close the PR, no? It is sufficient to keep the branch alive, if you need it for the experiments.

aehyvari:
Right, we can close this.

PR #545, opened 2022-08-19
STPSolver: Use 'informed' data structure in declareAtom

### LINKED ISSUE #544: Difference logic solver does not support incrementality
It seems the difference logic solver (QF_RDL, QF_IDL) does not support incrementality correctly.
This was not needed before, since they are no incremental benchmarks for these logics in SMT-LIB.

Here is a small example:
```
(set-logic QF_IDL)
(declare-fun a () Int)
(declare-fun b () Int)
(declare-fun c () Int)
(set-info :status sat)
(assert (> (- a b) 10))
(check-sat)
(assert (> (- b c) 5))
(set-info :status sat)
(check-sat)

(exit)
```

blishko:
This unifies the behaviour with LASolver.
Since 'known' PTRefs are not being cleared, STPSolver did not work in
incremental mode, as it was ignoring the re-declaration of atoms.

The whole design of 'informed' and 'known' PTRefs should be improved.

Fixes #544.


blishko:
This is better dealt with in #546.

PR #546, opened 2022-08-22
Clarify how theory solvers register what terms they know about

### LINKED ISSUE #544: Difference logic solver does not support incrementality

aehyvari:
Here's yet another suggestion on how to address issue #544.  The idea, as suggested in PR #545, is to clarify the use of `isInformed` and remove `isKnown`.

The relevant interface now has two methods: `isValid` which returns `true` if the term can be declared to the theory solver, and `isInformed` which returns `true` if the term has been declared to the theory solver.

Fixes #544.

The commit touches `egraph` solver in a non-trivial way, but doesn't seem to break anything.  I'm attaching the relevant plot.
non-incremental-QF_UF-82f50f87-2022-08-19_vs_3089764c-2022-08-19


src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -152,7 +152,6 @@ lbool Egraph::getPolaritySuggestion(PTRef p)
     if (term.size() > 2) { return l_Undef; } // For now focus on 2 arguments
     PTRef lhs = term[0];
     PTRef rhs = term[1];
-    assert(enode_store.has(lhs) && enode_store.has(rhs));
```
blishko, line 155: Is there a reason to remove this assertion? Shouldn't we keep it?

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -353,7 +353,7 @@ void LASolver::declareAtom(PTRef leq_tr)
     // DEBUG check
     isProperLeq(leq_tr);
 
-    setKnown(leq_tr);
+    setInformed(leq_tr);
```
blishko, line 356: This can be removed. There is a call to `setInformed` on line 342.

blishko (changes requested):
Looks good!
I was working on something similar, but was not sure about some details.

I think this should work, I would  just keep the assertion and remove the duplicate call.

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: It fails.

blishko: Wait, why does it suddenly fail?
It means we now do not create some enodes we were creating before?

aehyvari: No, it doesn't look like that.  The failure happens in the regression test suite when SAT solver asks for a polarity suggestion.  I believe that it must have been some weird artefact of using inconsistent information.

src/tsolvers/lasolver/LASolver.cc
aehyvari: Removed.

src/tsolvers/egraph/EgraphSolver.cc
blishko: I am still not sure about this.
How can it happen that there is an equality (or disequality) in the Egraph such that the there is no Enode for one of the child terms?

blishko: So, for the purpose of a documentation:
There is no change in what Enodes are created. This happens for equalities inside an uninterpreted function.
The change is that before such equalities were marked as `known` but not as `informed`.
The code is `Egraph::getPolaritySuggestion` checks `informed`. Thus previously the execution stopped at this check, but now it gets past it.

aehyvari: Great, thanks!  I get the feeling that you agree that the new flow makes more sense?

It might be worth investigating if `isInformed` could be made virtual and the relevant information be obtained directly from the theory solver, possibly more efficiently.

blishko: Can you please add a comment in the source code about when this condition can be violated?
I don't have a better solution now, so I think we can use your suggestion.

blishko (approved):
LGTM!

src/tsolvers/egraph/EgraphSolver.cc
aehyvari: Ok, I added a comment there and separated the assertion removal + comment to a distinct commit for maximal visibility.

PR #547, opened 2022-08-24
Implement "reverse operators" for auxiliary variables

aehyvari:
Currently when we print terms there is no support for avoiding printing internal auxiliary expansion variables (currently `.ite`, `.div`, `.mod`).  This is a problem when terms need to be communicated to other solvers, and seems to cause (soundess) problems also if the terms are communicated within different opensmt instances.

This PR introduces reverse operators that allow expressing terms using the original terms defined over the input vocabulary.  This is done by syntactically obtaining the PTRef number from the symbol name and running a rewriter-based dag-traversal.

### DIFF
--- a/src/itehandler/IteHandler.cc
+++ b/src/itehandler/IteHandler.cc
@@ -15,10 +15,28 @@ PTRef IteHandler::rewrite(PTRef root) {
 
 PTRef IteHandler::getAuxVarFor(PTRef ite) {
     assert(logic.isIte(ite));
-    std::string name(".ite");
+    std::string name(itePrefix);
     name += std::to_string(ite.x);
     name += suffix;
     return logic.mkVar(logic.getSortRef(ite), name.c_str());
+}
+
+PTRef IteHandler::getIteTermFor(Logic const & logic, PTRef auxVar) {
+
+    std::string const & name = logic.getSymName(auxVar);
+    assert(name.compare(0, itePrefix.size(), itePrefix) == 0);
+    std::string numberStr;
+    for (auto it = name.begin() + itePrefix.size(); it != name.end(); ++it) {
+        if ('0' <= *it and *it <= '9') {
+            numberStr += *it;
+        } else {
+            break;
+        }
+    }
+    auto number = static_cast<uint32_t>(std::stoi(numberStr));
+    PTRef ite = { number };
+    assert(logic.isIte(ite));
+    return ite;
 }
 
 PTRef IteHandler::replaceItes(PTRef root) {
--- a/src/itehandler/IteHandler.cc
+++ b/src/itehandler/IteHandler.cc
@@ -33,7 +33,7 @@ PTRef IteHandler::getIteTermFor(Logic const & logic, PTRef auxVar) {
             break;
         }
     }
-    auto number = static_cast<uint32_t>(std::stoi(numberStr));
+    auto number = static_cast<uint32_t>(std::stoul(numberStr));
     PTRef ite = { number };
     assert(logic.isIte(ite));
     return ite;
--- a/src/logics/ArithLogic.cc
+++ b/src/logics/ArithLogic.cc
@@ -634,7 +634,6 @@ PTRef ArithLogic::mkMod(vec<PTRef> && args) {
         assert(intMod.sign() >= 0 and intMod < abs(divisorValue));
         return mkIntConst(intMod);
     }
-    divsAndModsSeen = true;
     return mkFun(sym_Int_MOD, {dividend, divisor});
 }
 
@@ -655,7 +654,6 @@ PTRef ArithLogic::mkIntDiv(vec<PTRef> && args) {
         auto intDiv = divisorValue.sign() > 0 ? realDiv.floor() : realDiv.ceil();
         return mkIntConst(intDiv);
     }
-    divsAndModsSeen = true;
     return mkFun(sym_Int_DIV, std::move(args));
 }
 
@@ -980,10 +978,9 @@ PTRef ArithLogic::removeAuxVars(PTRef tr) {
             return tr;
         }
     };
-    if (divsAndModsSeen) {
-        AuxSymbolMatcherConfig config(*this);
-        tr = Rewriter(*this, config).rewrite(tr);
-    }
+    // Note: this has negligible impact on performance, no need to check if there are divs or mods
+    AuxSymbolMatcherConfig config(*this);
+    tr = Rewriter(*this, config).rewrite(tr);
     return Logic::removeAuxVars(tr);
 }
 
--- a/src/logics/ArithLogic.cc
+++ b/src/logics/ArithLogic.cc
@@ -965,11 +965,10 @@ ArithLogic::getDefaultValuePTRef(const SRef sref) const
 
 PTRef ArithLogic::removeAuxVars(PTRef tr) {
     // Note: Since ites are removed first and div/mod's then, it is important to first reintroduce div/mod's and then ites
-    class AuxSymbolMatcherConfig : DefaultRewriterConfig {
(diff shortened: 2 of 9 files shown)


src/logics/ArithLogic.cc
```diff
@@ -959,6 +963,30 @@ ArithLogic::getDefaultValuePTRef(const SRef sref) const
...
+            auto symName = std::string_view(logic.getSymName(tr));
+            if (symName.compare(0, DivModConfig::divPrefix.size(), DivModConfig::divPrefix) == 0) {
+                return DivModConfig::getDivTermFor(logic, tr);
+            } else if (symName.compare(0, DivModConfig::divPrefix.size(), DivModConfig::modPrefix) == 0){
+                return DivModConfig::getModTermFor(logic, tr);
+            }
+            return tr;
+        }
+    };
+    if (divsAndModsSeen) {
```
blishko, line 984: Can we just *not* use this flag? It seems unnecessary to try to track this information.
Just run the rewriter everytime. We do the same in `LATheory`.
You can guard with `logic.hasIntegers()`, the same as we do in `LATheory`.

src/logics/ArithLogic.cc
```diff
@@ -959,6 +963,30 @@ ArithLogic::getDefaultValuePTRef(const SRef sref) const
         return Logic::getDefaultValuePTRef(sref);
 }
 
+PTRef ArithLogic::removeAuxVars(PTRef tr) {
+    // Note: Since ites are removed first and div/mod's then, it is important to first reintroduce div/mod's and then ites
+    class AuxSymbolMatcherConfig : DefaultRewriterConfig {
+        ArithLogic & logic;
+    public:
+        AuxSymbolMatcherConfig(ArithLogic & logic) : logic(logic) {}
+        bool previsit(PTRef) override { return true; }
```
blishko, line 972: This is unnecessary, this is the default behaviour, no need for `override`.

src/logics/Logic.cc
```diff
@@ -1265,6 +1265,24 @@ PTRef Logic::learnEqTransitivity(PTRef formula)
...
+    class AuxSymbolMatcherConfig : DefaultRewriterConfig {
+        Logic const & logic;
+        bool match(PTRef tr) const {
+            if (not logic.isVar(tr)) return false; // Only variables can match
+            auto symName = std::string_view(logic.getSymName(tr));
+            return symName.compare(0, IteHandler::itePrefix.size(), IteHandler::itePrefix) == 0;
+        }
+    public:
+        AuxSymbolMatcherConfig(Logic const & logic) : logic(logic) {}
+        bool previsit(PTRef) override { return true; }
```
blishko, line 1278: Unnecessary, this is the default behaviour.

src/logics/Logic.h
```diff
@@ -383,6 +383,8 @@ class Logic {
 
     PTRef learnEqTransitivity(PTRef); // Learn limited transitivity information
 
+public:
```
blishko, line 386: `public` is unnecessary here

src/rewriters/DivModRewriter.h
```diff
@@ -29,17 +31,45 @@ class DivModConfig : public DefaultRewriterConfig {
...
+                        throw OsmtInternalException("Parse error in auxiliary variable symbol: " + std::string(name));
+                        return false;
+                    } else {
+                        return true;
+                    }
+                }(parsingDividend, name));
+                parsingDividend = false;
+            }
+        }
+        return {{static_cast<uint32_t>(std::stoi(dividendNumberStr))},
```
blishko, line 65: Would `std::stoul` work here?

src/rewriters/DivModRewriter.h
```diff
@@ -29,17 +31,45 @@ class DivModConfig : public DefaultRewriterConfig {
...
+                divisorNumberStr += *it;
+            } else if (*it == '_') {
+                assert([](bool parsingDividend, std::string_view const name) {
+                    if (not parsingDividend) {
+                        throw OsmtInternalException("Parse error in auxiliary variable symbol: " + std::string(name));
+                        return false;
+                    } else {
+                        return true;
+                    }
+                }(parsingDividend, name));
```
blishko, line 61: Seems like unnecessary complicated piece of code. Can we just do
```suggestion
                    if (not parsingDividend) { throw OsmtInternalException("Parse error in auxiliary variable symbol: " + std::string(name)); }
```

blishko (changes requested):
Looks reasonable, left some comments.

I am wondering if the `removeAuxVars` must be part of `Logic`, if it cannot be somewhere else...

But I am also fine with the methods where they are now.

aehyvari:
> I am wondering if the removeAuxVars must be part of Logic, if it cannot be somewhere else...

I tried this, but it got a little unnatural (downcasts from `Logic` to `ArithLogic` because of `DivModRewriter`, in order to to construct the `div`s and `mod`s), so I settled to the solution of having it as part of `Logic`.

src/logics/ArithLogic.cc
aehyvari: This is a good question.  It looked like this rewriting can cause a performance bottleneck.  I didn't specifically check the effect of this optimisation though.  Checking `logic.hasIntegers()` would only help in case of `QF_LRA`.  The `QF_LIA` logic doesn't have `div`s & `mod`s, and therefore all the smtlib benchmarks would have this extra traversal for no reason.

One solution could be to record this in `ArithProperty`...

src/itehandler/IteHandler.cc
```diff
@@ -15,12 +15,30 @@ PTRef IteHandler::rewrite(PTRef root) {
...
+    assert(name.compare(0, itePrefix.size(), itePrefix) == 0);
+    std::string numberStr;
+    for (auto it = name.begin() + itePrefix.size(); it != name.end(); ++it) {
+        if ('0' <= *it and *it <= '9') {
+            numberStr += *it;
+        } else {
+            break;
+        }
+    }
+    auto number = static_cast<uint32_t>(std::stoi(numberStr));
```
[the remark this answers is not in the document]
aehyvari: Fixed.

src/logics/Logic.cc
aehyvari: Fixed.

src/logics/Logic.h
aehyvari: Fixed.

src/rewriters/DivModRewriter.h
aehyvari: Fixed.

src/logics/ArithLogic.cc
blishko: I would be surprised if this one additional traversal would be the bottleneck.
Which is not to say that that is not case, but I would like to see some evidence.

aehyvari: It feels wrong not to handle also the div&mod case, and I don't want to see this bug showing up again.

src/rewriters/DivModRewriter.h
aehyvari: Looks ok to me.

blishko (approved):
LGTM!

I would still prefer a version without `divsAndModsSeen`, but I am not going to block it because of this.

aehyvari:
> I would still prefer a version without divsAndModsSeen, but I am not going to block it because of this.

I'm also curious, I'm going to run the experiment to see if we can get away without.

aehyvari:
> I'm also curious, I'm going to run the experiment to see if we can get away without.

I could not show significant speedup with the optimisation, so I will remove it.

For the record: in extreme cases, where the run time of three calls to `ArithLogic::removeAuxVars` was ~20 seconds, the optimisation speed-up was in the order of 0.5% (measuring only `ArithLogic::removeAuxVars`).

PR #552, opened 2022-09-10
Remove FPU control flags

blishko:
It came to my attention that these flags are not present in all Linux environments, which could result in compilation errors.

It seems this was encountered also in MiniSAT some time ago: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=590254

I don't know what is the way to fix it, so here I am proposing to just remove this piece of code.


aehyvari:
OK.  Let's hope that this is some historic fix that isn't needed in the fix in modern Linuxes.

aehyvari (approved):
LGTM!

Issue #520, opened 2022-06-17
Bad performance of substitution code


blishko:
[This example](https://gist.github.com/blishko/52738162e5ac3f300c075bf29b5f3a38) shows pathological behaviour of OpenSMT's code for computing substitutions.

It contains 1000 variables and 1000 assertions that each variable is equal to the last one.
With the right order of assertions, this leads to OpenSMT running 1000 iterations of the main substitution loop. On every iteration only a single substitution is selected from all the existing facts. The problem is in `ArithLogic::arithmeticElimination` which computes substitutions from arithmetic equalities. There, is the same variable is picked for substitution from multiple equations, only the first one is used and others are ignored.
Better solution is to try to pick a different variable if the first choice has been picked before.

All the substitutions in this example could be computed in one iteration of the main substitution loop.
OpenSMT currently requires ~7 seconds to solve this example, Z3 needs ~0.03s. OpenSMT should be able to do the same.

blishko:
This problem has been fixed by the changes in #553.

PR #553, opened 2022-09-18
ArithLogic: Rewrite arithmetic substitutions

blishko:
This PR suggests to use a different mechanism to compute substitutions from arithmetic equalities.

In the old mechanism, if there already existed a substitution for a chosen variable, the equality was ignored (unless the RHS was a constant). It delegated more responsibility to the main substitution loop, which could be very costly, see #520.

The new mechanism tries to be a bit more clever before falling back to the previous behaviour.

- In the first phase, it tries to apply as much _substitutions by a constant_ as possible, iterating until fixed point. This is cheaper that doing this in multiple iterations of the main substitution loop.

- In the second phase, it searches for variables present only in a single equality. This equality is then used to derive a substitution for such variable. This prevents the bad performance seen in #520.

- Finally, it falls back to the previous behaviour.

Performance-wise, this change seems to be positive and crucial for some benchmarks in `QF_LIA/20220307-SMPT` category (SharedMemory-* subcategories).

**Side note:** This PR eliminates the use of `LAExpression` in substitution code. It has been replaced by our `Polynomial` class.
The final use of `LAExpression` is in `FarkasInterpolator`. After refactoring that code, `LAExpression` can removed completely.


blishko:
Here is the comparison on QF_LRA and QF_LIA to the master (after #509 was merged):
QF_LRA
non-incremental-QF_LRA-rewrite-2022-08-30_vs_subst-2022-09-16
QF_LIA
non-incremental-QF_LIA-rewrite-2022-09-02_vs_subst-2022-09-16

blishko:
Side note: The category `QF_LIA/convert` seems to be the most sensitive to changes. It would be good to investigate how we can stabilize the performance there.

src/logics/ArithLogic.cc
```diff
@@ -243,35 +244,197 @@ PTRef ArithLogic::normalizeMul(PTRef mul)
...
+                    poly_t constantPoly;
+                    constantPoly.addTerm(PTRef_Undef, coeff * logic.getNumConst(val));
+                    poly.merge(constantPoly, 1);
+                }
+                changed = true;
+            }
+        }
+        // remove used polynomials
+        auto lastValid = zeroPolynomials.size() - 1;
+        for (auto rit = processedIndices.rbegin(); rit != processedIndices.rend(); ++rit) {
```
blishko, line 314: Could this be achieved by some algorithm in STL?

src/logics/ArithLogic.cc
```diff
@@ -243,35 +244,197 @@ PTRef ArithLogic::normalizeMul(PTRef mul)
...
+        assert(logic.isEquality(eq));
+        poly_t poly;
+        PTRef polyTerm = logic.mkMinus(logic.getPterm(eq)[0], logic.getPterm(eq)[1]);
+        assert(logic.isLinearTerm(polyTerm));
+        if (logic.isLinearFactor(polyTerm)) {
+            auto [var,c] = logic.splitTermToVarAndConst(polyTerm);
+            auto coeff = logic.getNumConst(c);
+            poly.addTerm(var, std::move(coeff));
+        } else {
+            for (PTRef factor : logic.getPterm(polyTerm)) {
```
blishko, line 395: `assert(logic.isPlus(polyTerm))`

blishko:
@aehyvari , do you have time to review this PR? I would still like to polish the code, if possible. Maybe you will also spot something that should be changed.

aehyvari:
Yes sure!

aehyvari:
Looks really good!  I had some suggestions that I was trying out so I thought I'll write them as a PR to this one.  They're in  #554

src/logics/ArithLogic.cc
```diff
@@ -9,6 +9,7 @@
 #include "OsmtApiException.h"
 #include "IteHandler.h"
 #include "DivModRewriter.h"
+#include "Polynomial.h"
```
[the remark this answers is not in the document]
aehyvari: I wouldn't mind.  The logics easily get a little bloated apparently.

src/logics/ArithLogic.cc
```diff
@@ -243,35 +245,196 @@ PTRef ArithLogic::normalizeMul(PTRef mul)
...
+                changed = true;
+            }
+        }
+        // remove used polynomials
+        auto lastValid = zeroPolynomials.size() - 1;
+        for (auto rit = processedIndices.rbegin(); rit != processedIndices.rend(); ++rit) {
+            auto index = *rit;
+            zeroPolynomials[index] = std::move(zeroPolynomials[lastValid]);
+            zeroPolynomials.pop_back();
+            --lastValid;
```
aehyvari, line 318:
```suggestion
        for (auto rit = processedIndices.rbegin(); rit != processedIndices.rend(); ++rit) {
            zeroPolynomials[*rit] = std::move(zeroPolynomials.back());
            zeroPolynomials.pop_back();
```

src/logics/ArithLogic.cc
```diff
@@ -243,35 +245,196 @@ PTRef ArithLogic::normalizeMul(PTRef mul)
...
+    }
+    // Remove processed polynomials
+    std::vector<std::size_t> indicesToRemove(processedIndices.begin(), processedIndices.end());
+    std::sort(indicesToRemove.begin(), indicesToRemove.end());
+    auto lastValid = zeroPolynomials.size() - 1;
+    for (auto rit = indicesToRemove.rbegin(); rit != indicesToRemove.rend(); ++rit) {
+        auto index = *rit;
+        zeroPolynomials[index] = std::move(zeroPolynomials[lastValid]);
+        zeroPolynomials.pop_back();
+        --lastValid;
```
aehyvari, line 375:
```suggestion
    for (auto rit = indicesToRemove.rbegin(); rit != indicesToRemove.rend(); ++rit) {
        zeroPolynomials[*rit] = std::move(zeroPolynomials.back());
        zeroPolynomials.pop_back();
```

aehyvari (commented):
I forgot to fix these in the merged PR...  But I guess you wanted to have them like that, right?

src/logics/ArithLogic.cc
blishko: Yes, I wanted to have the `.back()` version.
I'll include these.

blishko: Pushed new version.

blishko:
I finally got the results back from the cluster. For LRA, the results are the same as the version before refactoring.
So, I think this should be ready now.
@aehyvari , can you make a final review?

aehyvari (approved):
LGTM!

PR #554, opened 2022-09-21
Rewrite substitutions suggestions

aehyvari:


src/logics/ArithLogic.cc
```diff
@@ -368,12 +364,11 @@ Logic::SubstMap collectSingleEqualitySubstitutions(ArithLogic & logic, std::vect
     // Remove processed polynomials
     std::vector<std::size_t> indicesToRemove(processedIndices.begin(), processedIndices.end());
     std::sort(indicesToRemove.begin(), indicesToRemove.end());
-    auto lastValid = zeroPolynomials.size() - 1;
     for (auto rit = indicesToRemove.rbegin(); rit != indicesToRemove.rend(); ++rit) {
         auto index = *rit;
-        zeroPolynomials[index] = std::move(zeroPolynomials[lastValid]);
+        std::swap(zeroPolynomials.at(index), zeroPolynomials.back());
```
blishko, line 369: Same as the comment above.
Now that I see it here, I see the code is duplicated, so could be refactored into a single function.

blishko (commented):
Thanks for the suggestions. I like them, except the `swap` :)

src/logics/ArithLogic.cc
```diff
@@ -281,16 +281,15 @@ Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<pol
...
                 if (not substitutions.has(var)) {
-                    auto val = -((poly.begin() + 1)->coeff) / term.coeff;
+                    auto val = -((poly.begin() + 1)->coeff) / coeff;
                     substitutions.insert(var, logic.mkConst(logic.getSortRef(var), val));
                     new_keys.push(var);
                 }
                 processedIndices.push_back(i);
                 varToPolyIndices.at(var).pop_back();
-                continue;
+                continue; // Unnecessary?  Is it clearer this way?
```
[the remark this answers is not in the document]
aehyvari: Makes sense to me.

src/logics/ArithLogic.cc
```diff
@@ -310,20 +309,18 @@ Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<pol
             }
         }
         // remove used polynomials
-        auto lastValid = zeroPolynomials.size() - 1;
         for (auto rit = processedIndices.rbegin(); rit != processedIndices.rend(); ++rit) {
             auto index = *rit;
-            zeroPolynomials[index] = std::move(zeroPolynomials[lastValid]);
+            std::swap(zeroPolynomials[index], zeroPolynomials.back());
```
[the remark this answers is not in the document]
aehyvari: Out of curiosity I tried benchmarked this with quick-bench.  In my experiment the code with swap is 3% smaller but it seems to be slower than with move.  The difference was is to measure, but sometimes it seemed to be up to 20%.  Not sure how long this link will work, but here's my experiment.

https://quick-bench.com/q/GKUrUn5YBYzyKAHItpFbhMBLiFk

src/logics/ArithLogic.cc
aehyvari: This suggestion seems to be wrong anyway, using `at` on a vector.  I'm not sure what happened.

blishko: You can use `at` with `vector`, it's just going to have implicit bound checking, compared to `operator []`.

aehyvari: I didn't know that :)
Good, anyway, I didn't want to use it on my suggestion.

blishko (approved):
LGTM!

PR #555, opened 2022-09-21
Style fixes pointed out by more recent compilers with `--wall -Wextra -Werror`

aehyvari:
It seems that some clang versions (14?) complain in our code where we are using `&` instead of `&&` (or `and`) when compiled with `--Wall -Wextra -Werror`.  I believe that the code should be using `and` in this case, so this PR suggests the fix.

### DIFF
--- a/src/common/FastRational.cc
+++ b/src/common/FastRational.cc
@@ -110,7 +110,7 @@ std::string FastRational::get_str() const
 
 FastRational gcd(FastRational const & a, FastRational const & b)
 {
-    assert(a.isInteger() & b.isInteger());
+    assert(a.isInteger() and b.isInteger());
     if (a.wordPartValid() && b.wordPartValid()) {
         return FastRational(gcd(a.num, b.num));
     }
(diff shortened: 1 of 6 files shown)


aehyvari:
There were a few others that showed up when I updated the compiler to clang 15.0.0

blishko (approved):
LGTM!

blishko:
The problem is that these do not seem to be caught in our CI.
Could we update clang in the corresponding job?

aehyvari:
> The problem is that these do not seem to be caught in our CI.
> Could we update clang in the corresponding job?

Yes!  I'm not sure where to update it.  It'd make sense to add another build target for Apple Silicon, but I'm not ready yet (see #556 & #557).  Maybe we could update Debian's LLVM if the clang-15 is already available there.  I've also been thinking of a light-weight Linux build based on arch.  Any opinions?

blishko:
No opinions :)

aehyvari:
It looks like there's only clang-12 in `cimg/base:stable` that we use in circleci.  The tag `stable` is deprecated and we should instead use `current`, but if I update to `current` our `pysmt`-based model validation fails to install for some reason.

I suggest that I'll open an issue about `cimg/base:current`, and meanwhile proceed with this merge so we don't forget these changes and have the upcoming tag update stored somewhere.

blishko:
Sounds good to me!

PR #559, opened 2022-10-05
Model validation update (continuous integration)

aehyvari:
OpenSMT's model validation regression test suite runs OpenSMT with model production enabled and checks the resulting models for correctness.  Currently the model validation  uses `pySMT` for validating models, but `pySMT` seems to be hard to compile on modern operating systems.

This PR suggests to add another model validator based on the [`dolmen`](https://github.com/Gbury/dolmen.git) project.  The `pySMT-based` system will still be used in the `build-starexec` CI build, and the dolmen-based system will be run in the `build-recent`.

The environment scripts are updated accordingly, and the `run_travis_commands.sh` script will run the model validation for the first tool for which it determines that the environment has been set up, or what the user specifies for the environment variable: `MODEL_VALIDATION=Dolmen` or `MODEL_VALIDATION=pySMT`.


ci/run_travis_commands.sh
```diff
@@ -29,7 +29,16 @@ if [[ ${CMAKE_BUILD_TYPE} == Debug ]]; then
...
-    cd ../regression_models && ./bin/run-tests.sh ../build/opensmt;
+    cd ../regression_models
+    if [ -f ./data/ENV-MODERN ]; then
+        ./bin/run-tests-dolmen.sh ../build/opensmt;
+    fi
+    if [ -f ./data/ENV-SMTCOMP ]; then
+        ./bin/run-tests-smtcomp.sh ../build/opensmt;
+    fi
+    if [ ! -f ./data/ENV-SMTCOMP -a ! -f ./data/ENV-MODERN ]; then
+        echo "Warning: the model regression environment is not set.  Skipping tests"
```
blishko, line 40: Is this a valid situation? I am just wandering if this should not be an error which the script should propagate with `exit 1`...

blishko (commented):
LGTM!
I am just trying to clarify one thing.

blishko:
It looks like setting up `dolmen` took 8 minutes in the CI machine.
That's a bit unfortunate.
Maybe we should look at caching options here?

blishko (approved):
LGTM!

PR #561, opened 2022-10-07
Build image update

aehyvari:
This PR proposes the use of custom-built docker images in opensmt.  This should bring the build times down especially since the model validation environment doesn't need to be compiled every time.

In addition, the PR suggests to split the jobs into smaller jobs, one for each `(compiler, build type)` combination.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,10 +93,27 @@ jobs:
                 name: Release build llvm
                 command: ./ci/run_travis_commands.sh
 
-    # To set up the centos-7 environment:
-    #  - yum install centos-release-scl devtoolset-8 gmp-devel libedit-devel
-    #  - scl enable devtoolset-8 bash
-    build-starexec:
+    build-starexec-debug:
+        docker:
+            - image: verifyusi/verify-env:starexec
+              auth:
+                username: mydockerhub-user
+                password: $DOCKERHUB_PASSWORD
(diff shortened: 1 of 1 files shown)


aehyvari:
The docker images used by the circleci config are available at [this repository](https://github.com/usi-verification-and-security/verify-env)

.circleci/config.yml
```diff
@@ -31,122 +46,91 @@ jobs:
...
+            auth:
+                username: mydockerhub-user
+                password: $DOCKERHUB_PASSWORD
+            environment:
+                CMAKE_BUILD_TYPE: Release
+                PARALLEL: ON
+                FLAGS: -Wall -Wextra -Werror
+                OSMT_INSTALL: ~/osmt-install
+        steps:
+            - checkout
```
blishko, line 91: I think we are missing the `Setup clang` step which should be the same as in  `build-recent-clang-debug`, no?

blishko (commented):
Nice! I like that we can now split the big jobs into smaller ones!

I believe there is one thing missing in the `clang-release` job, which would probably now not use clang, but gcc.

.circleci/config.yml
aehyvari: That's right.  Should be fixed now!

blishko (approved):
Nice!

PR #564, opened 2022-10-22
Changelog: Update for version 2.4.2

blishko:
This PR adds changelog entries for the upcoming release 2.4.2

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,10 +2,18 @@
 
 Bug fixes:
  - TSolverHandler: Remove dependency on the global objects initialization order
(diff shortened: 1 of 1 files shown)


blishko:
@aehyvari, looking at the merged PRs since last release, there were changes from you in the CI setup and something related to partitioning. Do you want to add entries for that?

aehyvari:
I added some points, and this looks ok to me.   We should add the as-static-as-possible os x binary for AArch64 for as well now that we can produce one that works.

aehyvari (approved):
LGTM!

PR #567, opened 2022-10-25
Logic: Evaluate `distinct` on constant arguments

blishko:
This is required for our Model to function properly.
Previously, `Logic` would just create a term like "(distinct 0 1 2)", instead of directly simplifying that to "true".

Note: It seems that the check with `Logic::isConstant` works well now.
For the numeric and boolean constants it returns `true`.
In uninterpreted functions, it returns `false` for the zero-argument functions (treated as variables), but it returns `true` for the elements from UF models.
I am afraid this might be a bit fragile, but it seems like the easiest fix.


aehyvari (approved):
LGTM!

PR #571, opened 2022-11-04
UFLA: Fix unsoundness

blishko:
Running SMT-LIB QF_UFLRA benchmarks pointed out unsoundness in our latest release.
The bug was caused by a combination of two factors.

First, UFLATHandler was creating a schedule that put LA solver before UF solver.
Second, LA solver was never clearing its `has_conflict` flag, so it was signaling it has detected a conflict even though it was not true.
This resulted in a situation where UF solver detected a conflict, but LA solver was the one asked for explanation. And even though LA solved signaled it has explanation, it did not actually have any. This resulted in empty explanation and immediate report of UNSAT, which was not the correct answer.

This PR puts UF solver on the first place in the schedule, which I believe should always be the case.
It also clears the `has_explanation` flag properly (hopefully) in the LA solver.


aehyvari:
> Is there any compelling argument to keep the previous schedule? That is, check LASolver before UFSolver?

I think it's better to do first UF, then LA.

aehyvari (approved):
LGTM!

aehyvari:
We should do a new release ASAP I suppose, it's not nice to have the most recent release being unsound.

PR #573, opened 2022-11-06
ArithLogic: Fix substitutions in presence of UF

blishko:
Previously, in `collectSingleEqualitySubstitutions`, the variables were processed in order as stored in `unordered_map`. Additionally, UF terms are treated as variables in ArithLogic. This meant that from an equality f(x) = x (with x being Real), the code could derive a substitution "x -> f(x)". However, such substitution results in an infinite cycle in the function that computes transitive closure of substitutions.

We fix this by forcing an order in which variables are checked, starting from variables with higher PTRef, which mean they they were created later. This ensures for the above case, that f(x) will be considered before x, so the computed substitution will be "f(x) -> x".


blishko:
@aehyvari , this is a second fix in UFLA. My changes in how substitutions are computed in ArithLogic caused a regression in QF_UFLRA (not unsoundness anymore). For some benchmarks, a substitution was derived that caused infinite cycle later in computing transitive closure of substitutions.

This fixes this problem, but we also loose a few benchmarks in QF_LRA and QF_LIA with this. Not many though, so I think it is still worth it. Substitutions should be re-visited later, there probably is a better way to do this.

However, in the current situation I propose to go with this change and make a release to have a fixed version out.

aehyvari:
Why is this happening?  Don't we have the substitution loop breaker?  Maybe it's broken somehow..

blishko:
There are two things:
1. SubstitutionLoopBreaker runs at the end of `Logic::retrieveSubstitutions`, but arithmetic substitutions are added only after this parent method has run.
2. I am not sure SubstitutionLoopBreaker would be able to handle this anyway. There isn't really a substitution cycle. The problem is even a single substitution "x -> f(x)". `Logic::retrieveSubstitutions` has a special code that handles this in UF terms: https://github.com/usi-verification-and-security/opensmt/blob/8c58827a9a33469e586158be66df860aa5a2ad35/src/logics/Logic.cc#L1030
If a variable occurs on the right-hand side of the substitution, this substitution is ignored.

aehyvari:
But I thought that f(x) = x would be defined as a substitution loop?  The substitution DAG should have both edges f(x) -> x and x -> f(x).

blishko:
No, we always derive only one substitution from each equality.

Deriving both does not make much sense to me. And if substitution breaker would decide to keep `x -> f(x)`? It would again lead to the infinite loop in transitive closure.

blishko:
Here are the results evaluating the effect of this change.
Note, the labels are a little misleading for QF_LRA and QF_LIA, just because I ran the experiments on the older version later.

blishko:
QF_UFLRA:
non-incremental-QF_UFLRA-uflra-2022-11-04_vs_uflra-2022-11-05

blishko:
QF_LRA:
non-incremental-QF_LRA-uflra-2022-11-09_vs_uflra-2022-11-05

blishko:
QF_LIA:
non-incremental-QF_LIA-uflra-2022-11-09_vs_uflra-2022-11-05

blishko:
In all cases, y-axis is this PR, x-axis is the master before this PR.

blishko:
At some point, the whole substitution computation should be re-visited.

blishko:
QF_UFLIA:
non-incremental-QF_UFLIA-uflra-2022-11-10_vs_uflra-2022-11-10

aehyvari:
I think this is reasonable for now.  In the future we should understand why Substitution loop breaking isn't doing its job here at the moment.

src/logics/ArithLogic.cc
```diff
@@ -341,9 +342,15 @@ PTRef polyToPTRef(ArithLogic & logic, poly_t const & poly) {
...
 
 Logic::SubstMap collectSingleEqualitySubstitutions(ArithLogic & logic, std::vector<poly_t> & zeroPolynomials) {
-    Logic::SubstMap substitutions;
-    std::unordered_map<PTRef, std::vector<std::size_t>, PTRefHash> varToPolyIndices;
+    // MB: We enforce order to ensure that later-created terms are processed first.
+    //     This ensures that from an equality "f(x) = x" we get a substitution "f(x) -> x" and not the other way
+    //     around, which would cause infinite cycle in transitive closure
+    struct PTRefGreaterThan {
+        bool operator()(PTRef first, PTRef second) const { return first.x > second.x; }
+    };
```
aehyvari, line 350: Isn't there already a greater than operator for `PTRef`?  But in any case the locality of the definition here might actually be a benefit.  I'm fine with this.

aehyvari (approved):
LGTM!

src/logics/ArithLogic.cc
blishko: I am not aware of it.

As the next step, I actually want to clean up the comparison operators for `PTRef`, because in the class, there is `operator<`, which however, compares the ids with `>`, so it is quite confusing.

PR #574, opened 2022-11-13
improved checking for let overloads

aehyvari:
This PR makes a localised change to the code that forbids overloading
let variables.  It makes no sense for a user to overlad `let` variables:
```
(let ((l t) (l u)) l)
```
Currently we use a set to check that this does not happen.  This PR
suggests an implementation avoiding the set construction by a reuse of
the existing data structures for lets.


blishko (commented):
The change from  char* to strings should be an improvement, although I think there is a problem in the current version.

Regarding the let binder, I am not sure if the benefits outweigh the cost.
Storing additional unsigned long for every `let` comes with a small memory overhead, just to remove the single check for duplicates. Is it really worth it?

aehyvari:
> Regarding the let binder, I am not sure if the benefits outweigh the cost.
Storing additional unsigned long for every let comes with a small memory overhead, just to remove the single check for duplicates. Is it really worth it?

We could just not check this condition...  The standard doesn't say anything about it, so we could have whatever is the natural semantic, the first or the last one is the valid one.

blishko:
> We could just not check this condition... The standard doesn't say anything about it, so we could have whatever is the natural semantic, the first or the last one is the valid one

Well, that sounds like the perfect solution.
I went and tried the example below with z3, yices and opensmt.
opensmt reports error because of exactly this check we are discussing.
But z3 and yices both happily return `unsat`.
So I think we should follow their lead, drop this check altogether and just remember the last value for the let identifier.

```
(set-logic QF_LRA)
(declare-fun x () Real)
(declare-fun y () Real)
(assert (let ((z x) (z y)) (not (= z y))))
(check-sat)
```

aehyvari:
This should implement now what we discussed above.

blishko (approved):
LGTM!

I think that we will get some more string copies with the change from `char *` to `std::string`, in the current code, but it should not really be noticeable.

The code looks cleaner this way :)

aehyvari:
This compares master (x-axis) against this PR with `(check-sat)` disabled.  Timeout was 10 seconds.  No surprises here, I'll merge.
non-incremental-QF_LIA-ce21c361-2022-11-27_vs_7578cee5-2022-11-27

blishko:
OK, now I am wondering about the instances where we need more than 10 seconds of processing the input file :D

PR #575, opened 2022-11-21
Post-release version bump

blishko:
I have some larger changes in mind (that will most likely propagate to API), so I am directly incrementing the minor version number, instead of just the patch version number.


blishko:
@BritikovKI , I think it is time you start actively participating in the reviewing process.

BritikovKI:
Will do gladly :)

BritikovKI (approved):
Lgtmt
I started a review, but it's not this much :)

PR #576, opened 2022-11-23
PTRef: Use intuitive implementation of comparison operators 

blishko:
Previously, PTRef was exposing `operator<` which was, however, using
`operator>` on the underlying `uint32_t` member. 
This is extremely counterintuitive.

This PR normalizes comparison of PTRefs.
To preserve the existing behaviour, all code that previously relied on 
`operator<` on `PTRefs` now relies on `operator>` (possibly through `std::greater`).

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -1589,7 +1589,7 @@ bool        Logic::hasSortBool(SymRef sr) const { return sym_store[sr].rsort() =
 bool        Logic::isArraySelect(SymRef sr) const { return selects.has(sr); }
 bool        Logic::isArrayStore(SymRef sr) const { return stores.has(sr); }
 
-void Logic::termSort(vec<PTRef>& v) const { sort(v, LessThan_PTRef()); }
+void Logic::termSort(vec<PTRef>& v) const { sort(v, std::less<PTRef>{}); }
 
 void  Logic::purify     (PTRef r, PTRef& p, lbool& sgn) const {p = r; sgn = l_True; while (isNot(p)) { sgn = sgn^1; p = getPterm(p)[0]; }}
 
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -532,7 +532,7 @@ PTRef Logic::mkXor(vec<PTRef>&& args) {
     else if (args[0] == getTerm_false() || args[1] == getTerm_false())
         return (args[0] == getTerm_false() ? args[1] : args[0]);
 
-    sort(args);
+    sort(args, std::greater<PTRef>{});
     tr = mkFun(getSym_xor(), std::move(args));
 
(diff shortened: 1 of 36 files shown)


aehyvari:
I can't think of any place where it would make a (logical) difference, so I'm optimistic that we can just switch to the more intuitive implementation with at most minor refactoring.

aehyvari:
This compares runs skipping `(check-sat)` on `QF_LIA`, with 10 second timeout, comparing master (x-axis) and the PR.  No surprises here.  I'll try to merge!

non-incremental-QF_LIA-ce21c361-2022-11-27_vs_7578cee5-2022-11-27

src/logics/Logic.cc
```diff
@@ -532,7 +532,7 @@ PTRef Logic::mkXor(vec<PTRef>&& args) {
     else if (args[0] == getTerm_false() || args[1] == getTerm_false())
         return (args[0] == getTerm_false() ? args[1] : args[0]);
 
-    sort(args);
```
blishko, line 535: Do we need to sort the arguments?
Shouldn't this be handled in `mkFun`?

src/logics/SubstLoopBreaker.cc
```diff
@@ -81,7 +81,7 @@ TargetVarList::TargetVarList(vec<PTRef>&& _children)
 
 TVLRef TargetVarListAllocator::alloc(vec<PTRef>&& _children)
 {
-    sort(_children);
```
blishko, line 84: Why do these need to be sorted?
Does it matter how they are sorted (ascending/descending)?

blishko (commented):
There are two cases of sorting about which I am not sure.
And one place where I would keep std::greater, just in case.

blishko:
I got the results for QF_LRA for the previous version, where I tried to preserve original behaviour. There was not change there.
Now I have removed most of std::greater and will re-run the experiments once the cluster is available.

blishko:
I finally got back the results from the cluster for the current version of this PR, it looks good.
@aehyvari , I removed `std::greater` in most cases.
I am also uncertain around two sorting operations, why they are needed.
Can you have another look?

src/logics/Logic.cc
aehyvari: Right now it isn't handled there.  Whether it should be done there is another question.  Quickly thinking not doing this in `mkFun` could allow additional optimisations when we are merging two already sorted boolean operators.

src/logics/SubstLoopBreaker.cc
aehyvari: I doubt the order matters.  I vaguely remember it being some sort of a requirement for termination, but I'm not sure.

aehyvari (approved):
LGTM!

PR #577, opened 2022-11-24
ArithLogic: Small simplifications

blishko:
These are two, hopefully not controversial, simplifications of the code in ArithLogic operations.
See the commit messages for details on each.


aehyvari (approved):
LGTM!

PR #578, opened 2022-12-25
Pterms: Remove now redundant struct ValPair

blishko:
Since the introduction of `Model` class, we have not been not using `ValPair` anymore.
I believe it can be safely removed now.


aehyvari (approved):
LGTM!

PR #580, opened 2022-12-29
Clang-format: Apply to Tableau files

blishko:
This PR changes a rule to disallow function definitions on a single line, unless this is a inline definition in a class body.
The modified rules are then applied to `src/tsolvers/lasolver/Tableau.{h,cc}`.

### DIFF
--- a/src/tsolvers/lasolver/Tableau.cc
+++ b/src/tsolvers/lasolver/Tableau.cc
@@ -1,6 +1,8 @@
-//
-// Created by Martin Blicha on 31.03.18.
-//
+/*
+ *  Copyright (c) 2018-2022, Martin Blicha <martin.blicha@gmail.com>
+ *
+ *  SPDX-License-Identifier: MIT
+ */
 
 #include "Tableau.h"
 #include <iostream>
@@ -13,11 +15,10 @@
 
 using namespace opensmt;
 namespace {
-    template<class C, class E>
-    inline bool contains(const C & container, const E & elem){
-        return container.find(elem) != container.end();
-    }
+template <class C, class E> inline bool contains(const C & container, const E & elem) {
+    return container.find(elem) != container.end();
 }
+} // namespace
 
 void Tableau::nonbasicVar(LVRef v) {
     if (isProcessed(v)) { return; }
@@ -38,7 +39,6 @@ void Tableau::newRow(LVRef v, std::unique_ptr<Polynomial> poly) {
     addRow(v, std::move(poly));
     varTypes[getVarId(v)] = VarType::QUASIBASIC;
     normalizeRow(v);
-
 }
 
 std::size_t Tableau::getNumOfCols() const {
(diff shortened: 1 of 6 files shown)


src/tsolvers/lasolver/Tableau.cc
```diff
@@ -13,11 +15,10 @@
...
 using namespace opensmt;
 namespace {
-    template<class C, class E>
-    inline bool contains(const C & container, const E & elem){
-        return container.find(elem) != container.end();
-    }
+template <class C, class E> inline bool contains(const C & container, const E & elem) {
+    return container.find(elem) != container.end();
 }
+} // namespace
```
BritikovKI, line 21: What is the point of this comment btw?

src/tsolvers/lasolver/Tableau.cc
```diff
@@ -78,9 +78,7 @@ std::vector<LVRef> Tableau::getNonBasicVars() const {
     std::vector<LVRef> res;
     res.resize(varTypes.size());
     for (unsigned i = 0; i < varTypes.size(); ++i) {
-        if (varTypes[i] == VarType::NONBASIC) {
-            res.push_back(LVRef{i});
-        }
+        if (varTypes[i] == VarType::NONBASIC) { res.push_back(LVRef{i}); }
```
BritikovKI, line 81: Hmm, that's a recommendation from linter?

src/tsolvers/lasolver/Tableau.cc
blishko: Huh, I have not noticed it before. It has been probably added by clang-format, but I will double-check it.

src/tsolvers/lasolver/Tableau.cc
blishko: That's the effect of the rule
`AllowShortIfStatementsOnASingleLine: WithoutElse`

I am undecided what is the best setting here.
Should we disallow even these short statements on single line?

src/tsolvers/lasolver/Tableau.cc
blishko: Looks like this is the the style people at LLVM are converging to: https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

There have been some discussions, but now they seem to comment the closing brace of namespace X with
`// namespace X`
Since this is anonymous namespace, there is no name at the end.

In clang-format, this is the setting `FixNamespaceComments`

I generally agree that it is good to annotate the end of namespace.
Maybe we can rise the limit for this, which is specified by setting `ShortNamespaceLines`.

BritikovKI: Hmm, ok, that's a very interesting point

src/tsolvers/lasolver/Tableau.cc
BritikovKI: Yeah, tbh I'm not sure, I suppose both is fine, if we already have this rule then let it be, it saves some space

BritikovKI (approved):
Lgtmt

PR #581, opened 2023-01-03
Remove the ad-hoc 'anon' symbol

blishko:
I wanted to do a small refactoring of symbols in OpenSMT and the `anon` symbol was causing some trouble for that.
The Egraph solver can work without this artificial symbol, so it should safe to remove it.

Previously, it was used to create special Enodes that do not need to represent the children of the corresponding Pterm.
However, it seems to be that these special Enodes can use the symbol of the Pterm without losing any effect on the program.

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -36,7 +36,6 @@ const char* Logic::tk_val_bool_default = "true";
 
 const char* Logic::tk_true     = "true";
 const char* Logic::tk_false    = "false";
-const char* Logic::tk_anon     = ".anon";
 const char* Logic::tk_not      = "not";
 const char* Logic::tk_uf_not   = ".uf-not";
 const char* Logic::tk_equals   = "=";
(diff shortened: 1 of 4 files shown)


blishko:
non-incremental-QF_UF-anon-2023-01-03_vs_master-2023-01-03

These are results from QF_UF. No regression.

src/logics/Logic.cc
```diff
@@ -65,7 +65,6 @@ Logic::Logic(opensmt::Logic_t _logicType) :
     , term_FALSE(mkConst(getSort_bool(), tk_false))
     , sym_TRUE(getSymRef(term_TRUE))
     , sym_FALSE(getSymRef(term_FALSE))
-    , sym_ANON(sym_store.newSymb(tk_anon, {}))
```
BritikovKI, line 68: Overall seem good, you've forgotten to remove tk_anon on line 40 though

blishko: Good catch! Fixed now.

BritikovKI (approved):
Lgtmt

PR #582, opened 2023-01-08
Symbols: Use interfaces with explicit return sort

blishko:
Previously, the implementation detail that stores the return sort in the same array as arguments sort leaked out up to Logic's creation of new function declaration.
It seems more natural and cleaner to keep the return sort explicitly separated from arguments' sorts up until the last moment.

This also avoids some unnecessary memory allocations.


aehyvari:
This handling looks now nice and clean!

aehyvari (approved):
LGTM!

PR #584, opened 2023-01-13
Pterm: Return arguments by value

blishko:
Pterm's arguments should never be changed once the term is created. Hence, it does not make sense to return an argument by reference. Additionally, PTRefs are small enough to be returned by value instead of by const reference.


blishko:
@aehyvari , do you agree with this?

aehyvari:
Yes, especially since the change doesn't break existing code. Good!

aehyvari (approved):
LGTM!

PR #585, opened 2023-01-18
ArithLogic: Avoid VecMap usage in mkPlus

blishko:
In application that issues a large number of simple queries, such as Spacer algorithm, a surprisingly large portion of the runtime is spend in ArithLogic::mkPlus.

One reason is the usage of VecMap which is quite memory-intensive data structure.  With the proposed change we hope to reduce the amount of memory allocated and deallocated in each call to mkPlus.  Less work with memory should lead to better overall performance.

### DIFF
--- a/src/logics/ArithLogic.cc
+++ b/src/logics/ArithLogic.cc
@@ -13,6 +13,7 @@
 #include <map>
 #include <memory>
 #include <sstream>
+#include <variant>
 
 const std::string ArithLogic::e_nonlinear_term = "Logic does not support nonlinear terms";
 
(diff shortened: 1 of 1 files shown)


src/logics/ArithLogic.cc
```diff
@@ -585,34 +586,47 @@ PTRef ArithLogic::mkPlus(vec<PTRef> && args)
...
-        } else
-            s2t[v].push(c);
+        if (not varIndices.has(v)) {
+            varIndices.insert(v, simplified.size());
+            simplified.push_back({.var = v, .coeff = c});
+        } else {
+            auto index = varIndices[v];
+            auto & entry = simplified[index];
+            if (std::holds_alternative<PTRef>(entry.coeff)) {
+                entry.coeff = this->getNumConst(std::get<PTRef>(entry.coeff));
```
BritikovKI, line 610: Do we really need this if?
Maybe just put `.coeff = this->getNumConst(c)` in line 605, if we really need this func call?

BritikovKI: What is the reason to call getNumConst btw?

blishko: The point for this code is to avoid the conversion between `PTRef` and `FastRational` if possible. The conversion is only needed when there are multiple arguments of the sum with the same variable. Then we need to sum up all coefficients for the variable and for that we need coefficient terms converted to the rational numbers.

BritikovKI (changes requested):
Oh, actually I see the branch is in the need of rebase first

blishko:
Rebased.

BritikovKI (approved):
Approved :)

PR #587, opened 2023-01-26
Logic: Replace expensive method call with a lookup

blishko:
Previously, Logic::mkBinaryEq would make a call to PtStore::lookupSymbol to retrieve proper equality symbol for the arguments. However, Logic already caches equality symbol for each sort. Thus we can just look into this cache to retrieve the proper symbol.


aehyvari (approved):
LGTM!

PR #588, opened 2023-02-01
ArithLogic: Remove method isNumVarOrIte

blishko:
This method is a relict from the time when we tried to keep ITEs around and let the theory solver deal with them. That did not work and we now eliminate ITEs from the formulas before they are internalized. Thus LASolver should never encounter an ITE.


aehyvari (approved):
LGTM!

PR #589, opened 2023-02-02
Arrays: Add helper methods for querying array terms

blishko:
Methods isArrayStore and isArraySelect were missing the overloads taking PTRef as argument. This adds these overloads  to accompany the existing methods taking SymRef as argument.


aehyvari (approved):
LGTM!

PR #590, opened 2023-02-06
use fpu_control only on glibc

xgqt:
Signed-off-by: Maciej Barć <xgqt@gentoo.org>


blishko:
Hi @xgqt, thanks for pointing this out!
This is a research project so we don't have capacity to handle everything.
Please let us know if you find more current or future breaks.

blishko (approved):
LGTM!

PR #594, opened 2023-02-10
Arrays: Fix missing literals in read-over-weak-eq lemmas

### LINKED ISSUE #593: Incorrect computation of explanations in Array solver leads to unsoundness
OpenSMT gives incorrect result on the following example:

```
(set-logic QF_AX)
(set-info :status sat)
(declare-sort I 0)
(declare-sort E 0)
(declare-fun a () (Array I E))
(declare-fun b () (Array I E))
(declare-fun c () (Array I E))
(declare-fun j () I)
(declare-fun k () I)
(declare-fun i () I)
(declare-fun v () E)
(declare-fun x () Bool)
(assert (= b (store a i v)))
(assert (distinct k i))
(assert (distinct (select a j) (select b j)))
(assert (=> x (= j k)))
(assert (= (select c k) (select b k)))

(check-sat)
```

The reason seems to be incorrect computation of array lemmas.

For selects over weakly equivalent arrays, we compute conditions under which the selects must be equal.
This include that the indices on the weakly equivalent path must be different than the select index.
However, currently instead of the select index we use its current root in egraph.
This is not correct, because the lemma must be independent of any assumptions, while the fact that the index and its root are equal might hold only on the current trail, not at level 0.

blishko:
When computing the explanation for an array conflict (read-over-weak-eq lemma), we also need to add the conditions that explain why the select index is equivalent to its root.
The fact that they are equivalent might be true only on the current trail, so for the lemma to be logically valid, this explanation must be included.

Looks like we are already doing this for the extensionality lemmas, so we probably hit this problem before there.

Fixes #593.


aehyvari (approved):
LGTM!

I also remember having seen a similar problem before.  Great that this was found!

PR #595, opened 2023-02-10
Enable combination of arrays (and uninterpreted functions) with linear arithmetic

blishko:
This PR enables logic QF_ALIA, QF_ALRA, QF_AUFLIA and QF_AUFLRA (although only the integer versions exist in SMT-LIB benchmarks).

This is a prototype implementation. The combination requires only small changes.
The main differences are:

- generalization of the  `varLike` concept in LA
- generalization of the purification code for theory combination
- processing of lemmas generated from array solver with respect to theory combination

The last point is tricky and the current implementation is a bit hacky. Hopefully there will be a better solution in the future.


blishko:
Here are the comparisons with Z3.
ALIA:
non-incremental-QF_ALIA-alia-2023-02-10_vs_4 8 12-2023-02-07

AUFLIA:
non-incremental-QF_AUFLIA-alia-2023-02-10_vs_4 8 12-2023-02-08

regression/QF_ALIA/basic_unsat.smt2
```diff
@@ -0,0 +1,24 @@
+(set-logic QF_ALIA)
```
BritikovKI, line 1: Btw, where were the tests taken for regression?

src/logics/ArrayTheory.cc
```diff
@@ -5,43 +5,11 @@
  */
 
 #include "ArrayTheory.h"
+
```
BritikovKI, line 8: I think this space is not needed here

src/logics/ArithLogic.h
```diff
@@ -266,7 +266,7 @@ class ArithLogic: public Logic
 
     bool isNumVar(SymRef sr) const { return isVar(sr) and (yieldsSortInt(sr) or yieldsSortReal(sr)); }
     bool isNumVar(PTRef tr) const { return isNumVar(getPterm(tr).symb()); }
-    bool isNumVarLike(SymRef sr) const { return isNumVar(sr) or isIte(sr) or isIntDiv(sr) or isMod(sr) or (hasUFs() and isUF(sr)); }
+    bool isNumVarLike(SymRef sr) const { return yieldsSortNum(sr) and not isPlus(sr) and not isTimes(sr) and not isNumConst(sr); }
```
BritikovKI, line 269: Why there are so much differences between conditions?

src/tsolvers/axdiffsolver/ArraySolver.h
```diff
@@ -227,8 +227,8 @@ class ArraySolver : public TSolver {
 
     PTRef getEquality(ERef lhs, ERef rhs, Logic & logic) const { return logic.mkEq(egraph.ERefToTerm(lhs), egraph.ERefToTerm(rhs)); }
 
-    bool isFalsified(PTRef equality) const { return this->hasPolarity(equality) and this->getPolarity(equality) == l_False; }
-    bool isSatisfied(PTRef equality) const { return this->hasPolarity(equality) and this->getPolarity(equality) == l_True; }
+    bool isFalsified(PTRef equality) const { return logic.isFalse(equality) or (this->hasPolarity(equality) and this->getPolarity(equality) == l_False); }
```
BritikovKI, line 230: Why do we need this new requirement, if we have `this->getPolarity(equality) == l_False`?
Or is it possible to PTRef to be falsified without polarity?

src/logics/UFLATheory.cc
```diff
@@ -22,7 +27,7 @@ bool UFLATheory::simplify(const vec<PFRef>& formulas, PartitionManager &, int cu
 
 namespace {
     bool isArithmeticSymbol(ArithLogic const & logic, SymRef sym) {
-        return logic.isIF(sym) and not logic.isEquality(sym) and not logic.isBooleanOperator(sym);
+        return logic.isPlus(sym) or logic.isTimes(sym) or logic.isLeq(sym);
```
BritikovKI, line 30: I understand what it is doing here, but what was the logic of checking before? 
Everything is arithmetic symbol if it's IF and not equality or boolean operator?

src/logics/UFLATheory.cc
```diff
@@ -83,8 +89,10 @@ namespace {
         PTRef rewrite(PTRef ptref) override {
             auto const & term = logic.getPterm(ptref);
             if (isArithmeticSymbol(logic, term.symb())) {
-                return rewriteInternal(ptref, [this](PTRef child) { return isUninterpreted(logic, logic.getSymRef(child)); });
-            } else if (isUninterpreted(logic, term.symb())) {
+                return rewriteInternal(ptref, [this](PTRef child) {
+                    return isUninterpreted(logic, logic.getSymRef(child)) or (logic.hasArrays() and logic.isArraySelect(child));
+                });
+            } else if (isUninterpreted(logic, term.symb()) or logic.isArraySelect(term.symb()) or logic.isArrayStore(term.symb())) {
```
BritikovKI, line 95: How can we be sure if it's not AUFBV?

BritikovKI: This check won't pass in this case?

src/tsolvers/THandler.cc
```diff
@@ -87,18 +87,26 @@ std::vector<vec<Lit>> THandler::getNewSplits() {
...
-        assert(logic.isOr(clause));
-        for (int i = 0; i < logic.getPterm(clause).size(); i++) {
-            PTRef litTerm = logic.getPterm(clause)[i];
-            Lit l = tmap.getOrCreateLit(litTerm);
-            PTRef atomTerm = logic.isNot(litTerm) ? logic.getPterm(litTerm)[0] : litTerm;
-            assert(getLogic().isAtom(atomTerm)); // MB: Needs to be an atom, otherwise the declaration would not work.
-            declareAtom(atomTerm);
-            informNewSplit(atomTerm);
-            splitClauses.back().push(l);
+        if (logic.isAtom(clause)) {
```
BritikovKI, line 101: Do we need `if`, if we can do it inside of the loop below?
It'll take less space, or you just want to make a difference between atom and Or clause?

src/logics/ArrayHelpers.cc
```diff
@@ -0,0 +1,45 @@
...
+ *  SPDX-License-Identifier: MIT
+ *
+ */
+
+#include "ArrayHelpers.h"
+
+#include "TreeOps.h"
+
+vec<PTRef> collectStores(Logic const & logic, PTRef fla) {
+    class CollectStoresConfig : public DefaultVisitorConfig {
```
BritikovKI, line 13: As I understand we use it in LATheory and in Array theory
Why you've decided to add the processing of ALIA, ALRA, ... to the UFLA, and not to the array btw?

src/logics/ArrayTheory.cc
blishko: This follows the Google guidelines: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

It visually separates the header corresponding to this .cc file from other includes.

src/logics/UFLATheory.cc
blishko: Yes, this was a prototype for the theory combination of UF and LA, but it does not generalize.
Previously, it was saying that every interpreted function that is not equality or boolean connective is an arithmetic symbol. This is obviously not true when we add array-related functions.

src/logics/UFLATheory.cc
blishko: AUFBV is not supported by OpenSMT.

src/tsolvers/THandler.cc
blishko: What do you suggest?
The loop loops over the arguments of the term. If it's an atom, we do not want to loop over its arguments.

src/tsolvers/axdiffsolver/ArraySolver.h
blishko: It was not working for the `false` term. The theory solvers currently do not store information about `true` and `false` terms.

src/logics/ArithLogic.h
blishko: The previous version enumerated what should be included. The new version enumerates what should be excluded (included is anything that `yieldsSortNum` as the starting point).

This change means to include array select terms as `numVarLike`.

src/tsolvers/THandler.cc
BritikovKI: But what difference does it make? We will loop over its single argument, that's it

BritikovKI (approved):
Lgtmt

src/logics/ArrayHelpers.cc
blishko: The Array theory was created as the first implementation to handle QF_AX logic. However, since then I came to the conclusion that we cannot create separate theory for each logic, because of combinatorial explosion.
Instead we should try to unify what we have. My plan is to remove separate Array theory and instead put the functionality in `UFTheory`.

BritikovKI: Gotcha

PR #596, opened 2023-02-20
LASolver: Make debug print clearer

blishko:
If the current value of a variable is above upper bound, the message should say exactly that.
Similarly if the current value is below lower bound.


BritikovKI (approved):
Lgtmt

PR #600, opened 2023-02-22
Picky cdcl implementation

BritikovKI:


blishko (commented):
Instead of duplicating all the test files, we should just have a single file with picky configuration. Then we need to modify the test scripts to run all the files with picky configuration.

blishko (approved):
Nice!

README.md
```diff
@@ -88,6 +88,17 @@ $ opensmt test.smt2
 ```
 It is also possible to run OpenSMT without any arguments, in which case it reads the input from the standard input.
 
+### Other engines
+
+OpenSMT supports multiple SMT solving approaches. By the default it uses the approach described in the CoreSMTSolver. But it can also utilize 
+Lookahead or Picky CDCL. Those engines can be enabled by using following config:
```
blishko, line 94:
```suggestion
OpenSMT supports multiple SMT solving approaches. By the default it uses CDCL(T) with VSDIS decision heuristic. But it can also utilize (picky) Lookahead. Those engines can be enabled by using following config:
```

Also, add the option for setting the picky width.

PR #607, opened 2023-04-15
Interpret: Reduce size of call stack's activation record for parseTerm

blishko:
Last changes in parseTerm (a838804c) led to increased size of activation record and this resulted in more stack overflows on SMT-LIB benchmarks with huge nested lets.
Even though according to my intuition, the abovementioned changes should have decreased the memory footprint on the stack, it seems the opposite happened. My guess is that the changes allowed compiler to inline something that was not inlined previously, but this required more stack space to accomodate the inlined code.

The changes in this PR move more work (and data structure) to another method, which is then not inlined anymore (my guess). This not only reverts the size increase, but actually leads to further decrease of the activation record's size.

Proper solution would be to reimplement parsing in a non-recursive way, but this patch significantly helps in the meantime.

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -374,7 +374,16 @@ void Interpret::interp(ASTNode& n) {
     }
 }
 
-bool Interpret::addLetFrame(std::vector<opensmt::pair<PTRef,std::string>> const & bindings, LetRecords& letRecords) {
+bool Interpret::addLetFrame(ASTNode const & bindingsNode, LetRecords & letRecords) {
+    std::vector<opensmt::pair<PTRef, std::string>> bindings;
+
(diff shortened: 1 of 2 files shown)


blishko:
With this change, OpenSMT manages to succesfully parse (and solve) 12 more instances in QF_LIA than the latest released version.

non-incremental-QF_LIA-2 5 0-2023-04-13_vs_e1d45feb-2023-04-14

aehyvari:
Yeah, it must be related to how it passes the argument in stack.  The vector is bigger in stack than the `ASTNode`.  This was supposed to be a temporary step when writing the better parser, but because of delays in parser I'm happy to include this patch now.

aehyvari (approved):
LGTM!

PR #611, opened 2023-05-03
ArithLogic: Fix computation of substitutions

### LINKED ISSUE #610: Infinite cycle in transition closure of substitutions
On the following example, OpenSMT tries to infinitely expand the substitution "x -> f(x+1)".
```
(set-logic QF_UFLRA)
(declare-fun f (Real) Real)
(declare-fun x () Real)
(assert (= x (f x)))
(assert (= x (f (+ x 1.0))))
(assert (= (f x) (f (+ x 1.0))))
(check-sat)
```

Such substitutions should not occur, so this is actually a bug in the code that collects substitutions from equalities.

blishko:
ArithLogic: Fix computation of substitutions
    
Since we use transitive closure when computing substitutions, we should never start with a substitution "x -> t(x)" where we want to substitute variable x with a term containing this variable.
Such a substitution leads to an infinite rewriting in our code in transitive closure computation.
    
Since the terms in our polynomials are ordered by increasing PTRef of term's variable, we need to take the largest one, not smallest one when choosing which variable to use for the substitution.
Otherwise we might end up in the situation described above.

Fixes #610


blishko:
While this change is needed to successfully fix the mentioned issue, it causes some problems in particular categories of QF_LIA (20220307-SMPT/SharedMemory-*).
These categories encode large sums by using auxiliary variables for all intermediate sums. In our preprocessing we try to rewrite these variables to the full intermediate sums. This causes blowup in the number of terms we create and this causes bottleneck in computing substitutions.

I am thinking we should maybe try to only allow substitutions that make the terms simpler, not more complex.

blishko:
I am closing this PR, because the changes proposed here led to worse performance on the QF_LIA benchmarks. We instead try a different (quick) fix (#623).
Later, we should rewrite the substitution code. My idea would be to enable equalities in the LASolver and (mostly) let it deal with them properly.

PR #616, opened 2023-05-11
Interpret: Use cout to print get-value response

### LINKED ISSUE #615: get-value command doesn't flush output
This is a problem when calling OpenSMT from other tools as a system process.

Here is a Python code demonstrating the issue:

```
import subprocess

p = subprocess.Popen(['opensmt'], stdout=subprocess.PIPE, stdin=subprocess.PIPE)

p.stdin.write(b'(set-logic QF_LIA)\n')
p.stdin.write(b'(declare-const c Int)\n')
p.stdin.write(b'(check-sat)\n')
print p.stdout.readline()
#p.stdin.write(b'(get-model)\n')
p.stdin.write(b'(get-value (c))\n')
print p.stdout.readline()
```
The code prints `sat` and then it gets stuck in the second call to `p.stdout.readline()`. The expected behavior of the code is to terminate after printing `((c 0))` or similar.

OpenSMT version: 2.5.0
OS: Ubuntu 20.04

daniel-larraz:
This ensures stdout is flushed.

Fixes #615

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -663,13 +663,13 @@ void Interpret::getValue(const std::vector<ASTNode*>* terms)
         } else
             comment_formatted("Error parsing the term %s", (**(term.children->begin())).getValue());
     }
-    printf("(");
+    std::cout << "(";
     for (auto const & valPair : values) {
         auto name = logic.printTerm(valPair.first);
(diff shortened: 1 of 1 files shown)


blishko (approved):
LGTM!

PR #620, opened 2023-05-15
Interpret: Fix treatment of define funs

### LINKED ISSUE #613: Assertion returns an unknown sort
In a discussion with @blishko, he mentioned that OpenSMT struggles when a variable name, e.g., `G` below, is used with different sorts.

SMT instance
```
(set-logic QF_LIA)

(define-fun state ((A Bool) (B Bool) (C Bool) (D Bool) (E Int) (F Int) (G Int)) Bool (or (or (or (or (or (or (or (and (and (and (= A true) (= B true)) (not (= D true))) (>= E 1)) (and (and (and (= A true) (= D true)) (and (not (= B true)) (not (= C true)))) (>= E 1))) (and (and (= A true) (and (not (= B true)) (not (= D true)))) (>= E 1))) (and (and (and (= B true) (= D true)) (and (not (= A true)) (not (= C true)))) (>= E 1))) (and (= B true) (and (and (not (= A true)) (not (= C true))) (not (= D true))))) (and (and (= B true) (and (not (= A true)) (not (= D true)))) (>= E 1))) (and (and (= D true) (and (and (not (= A true)) (not (= B true))) (not (= C true)))) (>= E 1))) (and (and (and (not (= A true)) (not (= B true))) (not (= D true))) (>= E 1))))

(declare-fun A ()  Bool)
(declare-fun B ()  Bool)
(declare-fun C ()  Int)
(declare-fun D ()  Int)
(declare-fun E ()  Int)
(declare-fun F ()  Bool)
(declare-fun G ()  Bool)

(assert
    (and
      (and (= A true) (not G) (not F) (not B))
      (not (state B A G F C D E))
    )
)

(check-sat)
(exit)
```

Output
```
(error "Ambiguous symbol: `G'")

(error "assertion returns an unknown sort")

sat
```

I used OpenSMT 2.5.0. For this instance, both Z3 and CVC5 return `unsat`.

blishko:
Previously, we created variables corresponding to the formal arguments of defined functions (define-fun command). However, since in OpenSMT, we store variables with its name and sort, this led to clashes when a proper variable with the same name but different sort was declared later. However, the names of formal arguments of define-fun should not prevent a declaration of a variable with the same name.

The solution proposed here is to use unique internal names to store the formal arguments, while using our existing let mechanism to introduce the actual names used in the script to refer to these internal variables.

Fixes #613.


aehyvari (commented):
This should work.  I was wondering if it'd be possible to have the names of `define-fun`s live in a different `Logic` instance where they'd be translated to the search's `Logic`.  I haven't tried this, so it's possible that there are other complications.  But in principle I've always found (my) solution of declaring the names of `define-fun`s (only) as `PTRef`s in the main `Logic` a little hacky.

aehyvari:
True, it would still need the type of handling this PR suggests, but limited to that special `Logic` for templates.

aehyvari (approved):
LGTM!

src/logics/FunctionTools.h
```diff
@@ -31,7 +32,12 @@ class FunctionSignature {
 class TemplateFunction {
     FunctionSignature signature;
     PTRef tr_body;
+
+    inline static std::string_view constexpr template_arg_prefix = ".arg";
```
BritikovKI, line 36: Do we want all our arg names to start with dot?

src/api/Interpret.cc
```diff
@@ -899,21 +899,23 @@ bool Interpret::defineFun(const ASTNode& n)
...
         SRef sortRef = sortFromASTNode(sortNode);
         if (sortRef == SRef_Undef) {
             notify_formatted(true, "Undefined sort %s in function %s", sortSymbolFromASTNode(sortNode).name.c_str(), fname);
             return false;
         }
-        arg_sorts.push(sortRef);
-        PTRef pvar = logic->mkVar(arg_sorts.last(), varName.c_str());
-        arg_trs.push(pvar);
+        PTRef pvar = logic->mkVar(sortRef, TemplateFunction::nextFreeArgumentName().c_str());
+        std::string formalName = childNode.getValue();
```
BritikovKI, line 917: It doesn't seem to change much, but why do we use string instead of SRef now?
Was it just unoptimal before?

blishko:
I would prefer to do the formatting changes separately.

PR #621, opened 2023-05-16
ArithLogic: Fix normalization of equality

blishko:
Previously we were normalizing equality (almost) the same way as inequality. However, while for inequality it is possible that the right-hand-side can have both positive and negative leading term, for equality this meant that the same equality can be normalized to two different terms.
For example "x = 2" got normalized to "-2 = -x", while "2 = x" got normalized to "2 = x".

The proposed fix here is to ensure that the equalities always have positive leading term on the right-hand-side. Thus now even "x = 2" is normalized to "2 = x".


aehyvari:
Great!  This could actually have some positive effect on the run times!

PR #623, opened 2023-05-22
ArithLogic: Fix substitutions in the presence of uninterpreted functions

blishko:
This is a hot fix for a situation where ArithLogic picks a substitution like "x -> f(x)". In this case we ignore these equalities and do not add such substitutions.

This is not ideal, but necessary for new Certora benchmarks in QF_UFLIA. Several other things have been tried, but all attempts prove to be worse, especially on the 20220307-SMPT/SharedMemory-PT-000200 benchmarks from QF_LIA. I believe these benchmarks require a new approach. For those benchmarks, if substitutions are collected in a certain way, the transitive closure will blow up.
If we limit substitutions to not grow in number of variables, the Tableau blew up in size for these benchmarks.


blishko:
Fixes #610.

BritikovKI (approved):
Lgtmt

PR #625, opened 2023-05-27
Enable QF_UFIDL and QF_UFRDL logics

blishko:
For now, we are going to treat them as QF_UFLIA and QF_UFLRA. That infrastructure is (reasonably) well-tested. Maybe later we can try combining the specialized solver for difference logic with UFs.


BritikovKI (approved):
Lgtmt

BritikovKI:
Are there any benchmark results for this btw?

blishko:
I haven't run OpenSMT on QF_UFIDL benchmark set from SMT-LIB. I guess we will get the results directly from SMT-COMP :)
I believe there is no QF_UFRDL benchmark set.

PR #627, opened 2023-06-09
Fix uflia with mods rewrite

### LINKED ISSUE #626: Assertion failure / unsoundness on `QF_UFLIA` with `mod`
Assertion fails on the following `QF_UFLIA` instance.  The non-debug run is unsound.

```
(set-logic QF_UFLIA)
(declare-fun t () Int)
(declare-fun u (Int Int) Int)
(assert (and (> 0 t) (= t (mod (u 0 0) 2))))
(check-sat)
```

aehyvari:
Currently `QF_UFLIA` theory not call `DivModRewriter` on the formula.
As a result, the arithmetic solver sees, e.g., `mod` terms, and crashes
on an assertion failure.  In optimized code the error results in the
worst case in unsoundness.

This PR adds the necessary rewrites to `UFLATheory.cc`, as well as a
missing call to rewriter for `distinct`s.  The latter is untested.

There is also a new regression test for this.

Fixes #626 .


blishko (approved):
LGTM! :)

PR #631, opened 2023-06-28
CUFLogic: Add missing override for insertTerm

blishko:
This method is required by the Rewriter code, which now runs as part of preprocessing. Because of the missing override, the examples using CUFLogic were crashing.


BritikovKI (approved):
Lgtmt

PR #632, opened 2023-06-28
Arrays: Add example of API usage

blishko:
This PR adds an example how to create (and check) formula on the theory of arrays. In order to do that, we extend API of `Logic` to provide an easy way to get an array sort with the given domain and codomain.


BritikovKI (approved):
Lgtmt

PR #633, opened 2023-06-29
Remove outdated unused code

blishko:


blishko:
I guess what we have in `Opensmt.h` currently counts as our API. In general, all public methods in `Logic` and `MainSolver` are our API.
Which is not ideal.
It would be good to have a proper one, but that is quite a lot of work.

PR #634, opened 2023-06-29
Add support for QF_AUFLIRA logic

blishko:
This currently represents the most general logic OpenSMT can support. Even though we don't support mixing Int and Real terms, with this logic a user can create formulas using either Int or Real terms, in combination with arrays and uninterpreted functions.


BritikovKI (approved):
Lgtmt

PR #635, opened 2023-07-02
EGraph: Remove redundant mapping

blishko:
A map from `ERef` to `PTRef` is redundant since `Enode` itself remembers its corresponding `PTRef` a `Egraph` can access this information given `ERef`.


BritikovKI (approved):
Lgtmt

PR #640, opened 2023-08-08
Change to LASolver: Constraint reordering 

M4mbo:


blishko:
Results from StarExec on QF_LIA (20min timeout):
Before change: solved 7999 SAT + 4655 UNSAT = 12654
After change: solved 8003 SAT + 4666 UNSAT = 12669

Results from mantella (5 min timeout):
comparison

blishko:
Results are positive, but still all over the place. It would be great if we could figure out how to always find the good cuts.

blishko (approved):
LGTM!

PR #642, opened 2023-08-30
Change architecture around assertion stack, preprocessing and cnfization

blishko:
Maintaining the assertion stack should be the responsibility of the MainSolver, not the Theory.
Related to this change of responsibilities, we also change the preprocessing, so that it operates only on the current frame. (This should probably be changed so that simplifications from lower frames can be applied to higher frames as well, but we need to test if this would really be useful.)

Another change is a more separate Cnfizer, which now does not need to know about the PartitioningManager or the CoreSolver. Instead, it just reports new clauses to a callback that needs to be set up beforehand.

The larger goal here is to eventually remove separate Theory and Logic classes and have a single class responsible for managing terms (former Logic) and a single class for preprocessing (former Theory). These could be customized using a plugin architecture.

Depends on #676.


blishko:
I have anecdotal evidence that the main change, which means that we do not process the whole formula in the preprocessing for every check-sat call is beneficial in incremental SMT-LIB benchmarks and in BMC engine in Golem (which relies heavily on incremental solving).

src/cnfizers/Cnfizer.h
```diff
@@ -38,91 +38,75 @@ class SimpSMTSolver;
...
+
+
 //
 // Generic class for conversion into CNF
 //
 class Cnfizer
 {
 public:
-    SimpSMTSolver&      solver;
+    struct ClauseCallBack {
```
BritikovKI, line 49: What is the point of this structure?

src/cnfizers/Tseitin.cc
```diff
@@ -1,357 +1,160 @@
...
     // <=>
-    //
     // aux = ( -aux | a_0 ) & ... & ( -aux | a_{n-1} ) & ( aux & -a_0 & ... & -a_{n-1} )
-    //
     Lit v = this->getOrCreateLiteralFor(and_term);
-    vec<Lit> little_clause;
     vec<Lit> big_clause;
     int size = logic.getPterm(and_term).size();
     big_clause.capacity(size + 1);
-    little_clause.push(~v);
```
BritikovKI, line 103: Why remove little clause(or rather why it was there, because it looks like a data duplication)

src/cnfizers/Tseitin.cc
```diff
@@ -1,357 +1,160 @@
...
-        res &= addClause(little_clause);        // Adds a little clause to the solver
-
-        little_clause.pop();
+    vec<Lit> big_clause;
+    int size = logic.getPterm(or_term).size();
+    big_clause.push(~v);
+    for (int i = 0; i < size; ++i) {
+        PTRef arg = logic.getPterm(or_term)[i];
+        Lit argLit = this->getOrCreateLiteralFor(arg);
+        big_clause.push(argLit);
```
BritikovKI, line 80: Maybe add Tseitin to the clang-format, almost whole file is changed anyway(and Cnfizer, MainSolver, etc.)

src/logics/ArrayTheory.cc
```diff
@@ -8,18 +8,15 @@
 
 #include "ArrayHelpers.h"
 #include "DistinctRewriter.h"
-#include "TreeOps.h"
 
-
-bool ArrayTheory::simplify(vec<PFRef> const & formulas, PartitionManager &, int curr)
-{
+PTRef ArrayTheory::simplifyTogether(const vec<PTRef> & assertions, bool) {
```
BritikovKI, line 12: Why do we need bool param if it is never used?

src/logics/BVLogic.cc
```diff
@@ -9,7 +9,6 @@ Module: New Logic for BitVector
 #include "SStore.h"
 #include "PtStore.h"
 #include "BVLogic.h"
-#include "CUFLogic.h"
```
BritikovKI, line 12: Do we need BV logic at all? Or is it used in some specific cases?

src/tsolvers/bvsolver/BitBlaster.cc
```diff
@@ -2195,14 +2182,7 @@ BitBlaster::notifyEquality(PTRef tr)
...
-    sstat status = mainSolver.insertFormula(iff_tr, &msg);
-
-    if (status == s_True)
-        return l_True;
-    else if (status == s_False)
-        return l_False;
-    else
-        return l_Undef;
+    mainSolver.insertFormula(iff_tr);
+    return l_Undef;
```
BritikovKI, line 2186: Is it only undef? Never false or true?

src/smtsolvers/SimpSMTSolver.cc
```diff
@@ -184,7 +184,7 @@ lbool SimpSMTSolver::solve_(bool do_simp, bool turn_off_simp)
 //=================================================================================================
 // Added code
 
-bool SimpSMTSolver::addOriginalSMTClause(const vec<Lit> & smt_clause, opensmt::pair<CRef, CRef> & inOutCRefs)
+bool SimpSMTSolver::addOriginalSMTClause(vec<Lit> && smt_clause, opensmt::pair<CRef, CRef> & inOutCRefs)
```
BritikovKI, line 187: Why make it && and not const& like before?
Speed?

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -652,7 +650,7 @@ void CoreSMTSolver::analyze(CRef confl, vec<Lit>& out_learnt, int& out_btlevel)
         // Select next clause to look at:
         while (!seen[var(trail[index--])])
             ; // Do nothing
-        assert(index >= 0);
+        assert(index >= -1);
```
BritikovKI, line 653: Why index can be -1?

src/parallel/MainSplitter.h
```diff
@@ -20,21 +20,21 @@
 class MainSplitter : public MainSolver {
 
 private:
-    inline bool isSplitTypeScatter() const &       { return dynamic_cast<Splitter&>(ts.solver).isSplitTypeScatter(); }
+    inline bool isSplitTypeScatter() const &       { return dynamic_cast<Splitter&>(*smt_solver).isSplitTypeScatter(); }
```
BritikovKI, line 23: Is there some diff between smt_solver and ts.solver?

src/api/MainSolver.h
```diff
@@ -68,85 +68,122 @@ const sstat s_Error = toSstat( 2);
...
+            assert(i < frames.size());
+            return frames[i];
+        }
+        PushFrame & operator[](std::size_t i) {
+            assert(i < frames.size());
+            return frames[i];
+        }
 
-        void setSimplifiedUntil(std::size_t n) { simplified_until = n; }
+        void push() { frames.emplace_back(frameId++); }
```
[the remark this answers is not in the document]
blishko: Nothing changes for the public API. Golem can use OpenSMT after this change without any change in its code.

src/api/MainSolver.cc
```diff
@@ -205,73 +186,103 @@ PTRef MainSolver::rewriteMaxArity(PTRef root)
...
+        }
+        if (not res) { return s_False; }
+    }
+    return s_Undef;
+}
+
 sstat MainSolver::check()
 {
-    check_called ++;
+    ++check_called;
```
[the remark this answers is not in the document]
blishko: Style

src/api/MainSolver.cc
```diff
@@ -42,117 +42,96 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
-    {
-        int chars_written = asprintf(msg, "Top-level assertion sort must be %s, got %s",
-                 Logic::s_sort_bool, logic.printSort(logic.getSortRef(root)).c_str());
-        (void)chars_written;
-        return s_Error;
+    if (logic.getSortRef(fla) != logic.getSort_bool()) {
+        throw OsmtApiException("Top-level assertion sort must be Bool, got " + logic.printSort(logic.getSortRef(fla)));
     }
 
-    root = logic.conjoinExtras(root);
```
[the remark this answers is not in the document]
blishko: Originally, it was used for ITEs and probably for the CUF theory. But we don't need this anymore.

src/logics/ArrayTheory.cc
blishko: That parameter is used only in `UFTheory` now. It is related to handling `distinct` operations.
There should be a better mechanism for that, but I don't know yet how that should look like.

src/logics/BVLogic.cc
blishko: We don't, support for bit-vectors should be rewritten. But we might still re-use some code from BVLogic and BitBlaster, so we can keep them for now.

src/parallel/MainSplitter.h
blishko: No, previously the Tseitin class hold a reference to the CoreSMTSolver, but part of the proposed change is to make the cnfization process independent of that.

blishko:
Here are some statistics, running with 5 minute timeout:
On incremental QF_LRA, this PR can answer 946 `check-sat` queries compared to 879 answered by current version.
On incremental QF_LIA, this PR can answer 957341 queries compared to 848391 answered by current version.

Golem's KIND engine also shows improvement when running on LRA-TS benchmarks:
comparison

src/api/MainSolver.h
```diff
@@ -68,136 +51,162 @@ const sstat s_Error = toSstat( 2);
...
+                for (PTRef key : subs.getKeys()) {
+                    assert(not allSubst.has(key));
+                    allSubst.insert(key, subs[key]);
+                }
+            }
+            return allSubst;
+        }
+    };
+/** Actual MainSolver members **/
+protected:
```
BritikovKI, line 139: Fix tabulation(and below/above)

src/cnfizers/Cnfizer.cc
```diff
@@ -317,178 +137,92 @@ bool Cnfizer::isClause(PTRef e) {
...
-        CRef ref = iorefs.first;
-        if (ref != CRef_Undef) {
-            ipartitions_t parts = 0;
-            assert(currentPartition != -1);
-            opensmt::setbit(parts, static_cast<unsigned int>(currentPartition));
-            pmanager.addClauseClassMask(ref, parts);
-        }
-    }
-    return res;
+    (*clauseCallBack)(std::move(clause));
```
BritikovKI, line 173: That is a bit weird func, clauseCallBack is not static, I'm not sure how it works tbh
Where does it add a clause now?

src/cnfizers/Cnfizer.h
```diff
@@ -38,95 +38,79 @@ class SimpSMTSolver;
...
```
BritikovKI, line 49: What is the point of this structure? 
Only thing it does is an operator rewrite

src/api/MainSolver.h
```diff
@@ -68,136 +51,162 @@ const sstat s_Error = toSstat( 2);
 
 class MainSolver
 {
-  protected:
+protected: /** Helper classes to deal with assertion stack, preprocessing and substitutions **/
+    struct FrameId
```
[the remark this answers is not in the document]
BritikovKI: (And it will somewhat unite it withCnfizer)

src/logics/UFLATheory.cc
```diff
@@ -8,25 +8,17 @@
...
-        fla = rewriteDistincts(getLogic(), fla);
-        fla = rewriteDivMod<ArithLogic>(logic, fla);
-        PTRef purified = purify(fla);
-        if (logic.hasArrays()) {
-            purified = instantiateReadOverStore(logic, purified);
-        }
-        PTRef noArithmeticEqualities = splitArithmeticEqualities(purified);
-        this->getTSolverHandler().setInterfaceVars(getInterfaceVars(noArithmeticEqualities));
-        currentFrame.root = noArithmeticEqualities;
+PTRef UFLATheory::preprocessAfterSubstitutions(PTRef fla, PreprocessingContext const & context) {
```
BritikovKI, line 11: Is it possible to do before substitutions as well?

src/parallel/ScatterSplitter.h
```diff
@@ -56,7 +56,7 @@ class ScatterSplitter :  public SimpSMTSolver, public Splitter {
 protected:
     bool     scatterLevel();                                                  // Are we currently on a scatter level.
     opensmt::pair<SplitData,lbool> createSplitAndBlockAssumptions();          // Create a split formula and place it to the splits vector.
-    bool     excludeAssumptions(vec<Lit> const & neg_constrs);                // Add a clause to the database and propagate
+    bool     excludeAssumptions(vec<Lit> && neg_constrs);                     // Add a clause to the database and propagate
```
BritikovKI, line 59: Why not const anymore?

src/smtsolvers/CoreSMTSolver.h
```diff
@@ -651,35 +648,22 @@ inline bool     CoreSMTSolver::enqueue         (Lit p, CRef from)
     return value(p) != l_Undef ? value(p) != l_False : (uncheckedEnqueue(p, from), true);
 }
 
-inline bool     CoreSMTSolver::addOriginalClause(const vec<Lit> & ps)
+inline bool     CoreSMTSolver::addOriginalClause(vec<Lit> && ps)
 {
-    return addOriginalClause_(ps);
+    return addOriginalClause_(std::move(ps));
 }
-inline bool     CoreSMTSolver::addEmptyClause  ()
```
BritikovKI, line 658: addEmptyClause is no longer needed?
Mb also remove it in Solver.h and SimpSolver.h?

src/api/MainSolver.h
blishko: I believe indentation is correct, no?
This (and below) belongs to the `MainSolver` class.
Above you have some nested classes, so their members are indented more.

src/api/MainSolver.h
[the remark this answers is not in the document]
blishko: Sometimes, it is good to have strong types. It can make the interfaces cleaner and safer.
But it looks like we do not really need it, so maybe we can remove the struct here.

src/cnfizers/Cnfizer.h
blishko: It's a callback, a customization point to specify, what should happen with the clauses that Cnfizer produces. This means that `Cnfizer` is only concerned with creating the clauses and does not need to care about specifics of what to do with the clauses.

src/logics/UFLATheory.cc
blishko: It is, but this theory was not doing anything before substitutions before, so it is also not doing anything in this PR.

src/parallel/ScatterSplitter.h
blishko: To avoid extra copy. It just passes the vector to `addOriginalClause`.

src/smtsolvers/CoreSMTSolver.h
blishko: It was not used anywhere.
`Solver.h` and `SimpSolver.h` are MiniSat files that are not currently in use.
No reason to go and make changes in those.

src/smtsolvers/SimpSMTSolver.cc
```diff
@@ -180,7 +180,7 @@ lbool SimpSMTSolver::solve_(bool do_simp, bool turn_off_simp)
...
```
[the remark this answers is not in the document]
blishko: So we can move it to `CoreSMTSolver::addOriginalClause_`.

src/api/MainSolver.cc
```diff
@@ -1,28 +1,10 @@
-/*********************************************************************
```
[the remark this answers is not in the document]
blishko: I think there will be more formatting changes, so we should do it afterwards, not to obfuscate the actual changes in this PR.

src/cnfizers/Cnfizer.cc
blishko: `clauseCallBack` is a pointer to a function object, we are just calling it here to process the `clause`.

src/api/MainSolver.cc
```diff
@@ -37,124 +19,182 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+    initialize();
+}
+
+void MainSolver::initialize() {
+    frames.push();
+    frameTerms.push(logic.getTerm_true());
+    substitutions.push();
+    smt_solver->initialize();
+    opensmt::pair<CRef, CRef> iorefs{CRef_Undef, CRef_Undef};
+    smt_solver->addOriginalSMTClause({term_mapper->getOrCreateLit(logic.getTerm_true())}, iorefs);
```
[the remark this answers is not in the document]
blishko: This corresponds to the code that was in `Cnfizer::initializer()` before.
I am not sure if we could go without it, maybe we could, but I have not tried.
It is just asserting to the underlying SAT solver, that literals corresponding to `true` and `not false` should be assigned to true (because we are adding unit clauses here).

src/api/MainSolver.cc
```diff
@@ -37,124 +19,182 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
 #include "IDLTHandler.h"
+#include "Substitutor.h"
 #include <thread>
-#include <fcntl.h>
 
 namespace opensmt { bool stop; }
 
-void
-MainSolver::push()
+sstat MainSolver::simplifyFormulas() {
```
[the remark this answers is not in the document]
blishko: Fixed!

src/smtsolvers/CoreSMTSolver.h
BritikovKI: Ok, makes sense, will we need to use them though?
Or they should be refactored/removed totally(as a general knowledge)

src/parallel/ScatterSplitter.h
BritikovKI: Ok, makes sense, one can argue that const is good because it guarantees that excludeAssumptions won't change anything => it will be more safe...

But tbh I am far away from being c++ purist with strong opinions 😄

src/api/MainSolver.cc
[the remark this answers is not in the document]
BritikovKI: Ok

src/api/MainSolver.cc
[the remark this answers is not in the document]
BritikovKI: Ok, sounds fair, if it works let's not change it (for now at least)!

src/smtsolvers/CoreSMTSolver.h
blishko: We can discuss if we should remove MiniSAT files that we do not use. They are there for historical reason, but maybe we can remove them now.

src/parallel/ScatterSplitter.h
blishko: rvalue (`&&`) means the caller no longer care what happens to the vector, the callee can do whatever they want to do with it.

src/smtsolvers/SimpSMTSolver.cc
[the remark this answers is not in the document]
BritikovKI: Without copies you mean?
Ok on one hand, on the other we lose the guarantee that the vector won't be changed in the function...

src/api/MainSolver.cc
```diff
@@ -37,93 +19,131 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
-    for (std::size_t i = frames.getSimplifiedUntil(); i < frames.size() && status != s_False; i++) {
-        getTheory().simplify(frames.getFrameReferences(), pmanager, i);
-        frames.setSimplifiedUntil(i + 1);
-        const PushFrame & frame = pfstore[frames.getFrameReference(i)];
-
-        if (keepPartitionsSeparate) {
-            vec<PTRef> const & flas = frame.formulas;
-            if (flas.size() == 0 or std::all_of(flas.begin(), flas.end(), [&](PTRef fla) { return fla == logic.getTerm_true(); })) {
-                pmanager.assignTopLevelPartitionIndex(0, logic.getTerm_true());
-                status = giveToSolver(logic.getTerm_true(), frame.getId());
```
BritikovKI, line 122: Why did you exchange it for simple push on line 141?

src/smtsolvers/SimpSMTSolver.cc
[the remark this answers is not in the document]
blishko: Yes, without copies.
We do not need such the guarantees.
The change means that we are consuming the vector. Thus, the caller knows that when it passes us the vector, it cannot access it again, because we have consumed it.

src/api/MainSolver.h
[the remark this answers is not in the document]
BritikovKI: Ok, sounds good!

src/api/MainSolver.h
```diff
@@ -68,136 +51,162 @@ const sstat s_Error = toSstat( 2);
...
+    {
+        uint32_t id;
+        bool operator== (const FrameId other) const { return id == other.id; }
+        bool operator!= (const FrameId other) const { return id != other.id; }
+    };
+
+    struct PushFrame {
```
BritikovKI, line 62:
```suggestion
    struct PushFrame 
    {
```

src/api/MainSolver.h
BritikovKI: Yup, sry, my bad (was a bit confused with deetions and additions), but there is duplicate `public` lines 182, 205

src/api/MainSolver.cc
blishko: I am not sure what you mean. The line 141 in the new version does not correspond to this line.
It seems I have simple remove asserting literal `true` to the solver in case the frame is empty or consists of only `true` formulas. I am not sure why that was the case before. Either we can really just ignore such frames, or we are missing a test in our regression test suite.

src/cnfizers/Cnfizer.cc
BritikovKI: But it is never assigned anything. This is the only place `clauseCallBack` pointer is actually used and it is directing to the empty object (I searched by the `clauseCallBack` name)

src/api/MainSolver.h
blishko: You are right that we are inconsistent in the style.
I think the style that we use elsewhere is to have opening brace at the same line as class name.
We will catch these style problems when we add it to clang-format files.

src/cnfizers/Cnfizer.cc
blishko: The callback is set using `setClauseCallBack` method and we set the callback in `MainSolver::giveToSolver`.

src/api/MainSolver.h
BritikovKI: Ok!

src/api/MainSolver.cc
BritikovKI: Hmm, ok, that is indeed weird
Well, in theory it shouldn't break anything, if it will, we will find out then 😅

src/cnfizers/Cnfizer.cc
BritikovKI: Aaaaah, sry, I missed the fact that you have "recreated" the same struct inside a function

Now it seems solid!

PR #643, opened 2023-08-31
Remove support for CUF

blishko:
CUF theory is not used anymore and it represents a burden for maintenance.
As a preparation for a larger refactoring, we remove everything related to CUF logic.

It turned out that even our BV solver is using CUF theory, so this disables the limited support for bit-vectors we had in OpenSMT.
I don't see this as a problem since the whole support for BV in OpenSMT needs more work. Currently it only supports one-size bitvectors which does not correspond to SMT-LIB standard.
However, part of the code might still be useful, so we keep BV-related code around for now.


aehyvari (approved):
LGTM!

PR #644, opened 2023-09-01
Remove functionality to dump solver state

blishko:
I believe this functionality is no longer needed.
I did not find the corresponding read state functionality. Maybe we removed that already some time ago?


BritikovKI (approved):
Lgtmt

PR #645, opened 2023-09-01
MainSolver: Keep only one way to assert formula

blishko:


BritikovKI (approved):
Lgtmt

src/api/Interpret.cc
```diff
@@ -217,29 +217,20 @@ void Interpret::interp(ASTNode& n) {
...
-                        if (status == s_Error)
-                            notify_formatted(true, "Error");
-                        else if (status == s_Undef)
-                            notify_success();
-                        else if (status == s_False)
+                        if (logic->getSortRef(tr) != logic->getSort_bool()) {
+                            auto errorMessage = "Top-level assertion sort must be Bool, got " + logic->printSort(logic->getSortRef(tr));
+                            notify_formatted(true, errorMessage.c_str());
+                        } else {
+                            main_solver->insertFormula(tr);
```
aehyvari, line 231: Maybe
```suggestion
                        try {
                            main_solver->insertFormula(tr);
                            notify_success();
                        } catch (OsmtApiException const & e) {
                            notify_formatted(true, e.msg.c_str());
                        }

```
(modulo my c++ syntax & other errors...)

blishko: Fixed and added a test for this as well, just in case.

aehyvari (approved):
LGTM!

PR #646, opened 2023-10-08
Changes for JavaSMT integration

daniel-raffler:
Hello Martin,
as mentioned in https://github.com/sosy-lab/java-smt/pull/302#discussion_r1349601987, here are the changes I made to OpenSMT to make it work with JavaSMT. Simply merge the changes you can use, and we'll keep the rest in our API patch file.

Thanks in advance,
Daniel

Edit: I reverted [128d07a](https://github.com/usi-verification-and-security/opensmt/pull/646/commits/128d07aea81433c3803bb4e971e13befb35207fc) as it caused (unit) tests to fail, and [1ae6019](https://github.com/usi-verification-and-security/opensmt/pull/646/commits/1ae601976458e9f2740a39462e5787a16a966a05) as it was the reason for failed regression tests. Without those changes, however, some of the JavaSMT tests will fail. I'll see if I can rework the code, but this may take some time.

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -43,11 +43,6 @@ Interpret::~Interpret() {
     for (int i = 0; i < term_names.size(); ++i) {
         free(term_names[i]);
     }
-
-    // if the logic is given in the constructor and was not set via string-input, we cleanup the reference
-    if (!ownsLogic) {
-        logic.release();
-    }
 }
 
 PTRef
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -1021,11 +1021,13 @@ void Interpret::comment_formatted(const char* fmt_str, ...) const {
 
 
 void Interpret::notify_formatted(bool error, const char* fmt_str, ...) {
+    std::stringstream out;
+
     va_list ap;
     int d;
     char c1, *t;
     if (error)
-        std::cout << "(error \"";
+        out << "(error \"";
 
     va_start(ap, fmt_str);
     while (true) {
@@ -1034,25 +1036,31 @@ void Interpret::notify_formatted(bool error, const char* fmt_str, ...) {
             switch (*fmt_str++) {
             case 's':
                 t = va_arg(ap, char *);
-                std::cout << t;
+                out << t;
                 break;
             case 'd':
                 d = va_arg(ap, int);
-                std::cout << d;
+                out << d;
                 break;
             case '%':
-                std::cout << '%';
+                out << '%';
                 break;
             }
         }
         else if (c1 != '\0')
-            std::cout << c1;
+            out << c1;
         else break;
     }
     va_end(ap);
+
     if (error)
-        std::cout << "\")" << '\n';
-    std::cout << std::endl;
+        out << "\")" << '\n';
+    out << std::endl;
+
+    if (error)
+      throw new OsmtApiException(out.str());
+
+    std::cout << out.str();
 }
 
 void Interpret::notify_success() {
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -43,6 +43,11 @@ Interpret::~Interpret() {
     for (int i = 0; i < term_names.size(); ++i) {
         free(term_names[i]);
     }
+
+    // if the logic is given in the constructor and was not set via string-input, we cleanup the reference
+    if (!ownsLogic) {
+        logic.release();
+    }
 }
 
 PTRef
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -1026,13 +1026,11 @@ void Interpret::comment_formatted(const char* fmt_str, ...) const {
 
 
 void Interpret::notify_formatted(bool error, const char* fmt_str, ...) {
-    std::stringstream out;
-
     va_list ap;
     int d;
     char c1, *t;
     if (error)
-        out << "(error \"";
+        std::cout << "(error \"";
 
     va_start(ap, fmt_str);
     while (true) {
@@ -1041,31 +1039,25 @@ void Interpret::notify_formatted(bool error, const char* fmt_str, ...) {
             switch (*fmt_str++) {
             case 's':
                 t = va_arg(ap, char *);
-                out << t;
+                std::cout << t;
                 break;
             case 'd':
                 d = va_arg(ap, int);
-                out << d;
+                std::cout << d;
                 break;
             case '%':
-                out << '%';
+                std::cout << '%';
                 break;
             }
         }
         else if (c1 != '\0')
-            out << c1;
+            std::cout << c1;
         else break;
     }
     va_end(ap);
-
     if (error)
-        out << "\")" << '\n';
-    out << std::endl;
-
-    if (error)
-      throw new OsmtApiException(out.str());
-
-    std::cout << out.str();
+        std::cout << "\")" << '\n';
+    std::cout << std::endl;
 }
 
 void Interpret::notify_success() {
--- a/src/api/Interpret.h
+++ b/src/api/Interpret.h
@@ -135,6 +135,7 @@ class Interpret {
     std::unique_ptr<MainSolver> main_solver;
 
     bool            f_exit;
+    bool            ownsLogic;
 
     // Named terms for getting variable values
     MapWithKeys<const char*,PTRef,StringHash,Equal<const char*>> nameToTerm;
@@ -188,7 +189,19 @@ class Interpret {
     Interpret(SMTConfig & c)
         : config     (c)
         , f_exit     (false)
+        , ownsLogic  (true)
         { }
+
+    /** This constructor reuses the existing logic from caller context. */
+    Interpret(SMTConfig & c, Logic & l)
+        : config     (c)
+        , f_exit     (false)
+        , ownsLogic  (false)
+        {
+          logic.reset(&l);
+          main_solver = createMainSolver(l.getName().c_str());
+          main_solver->initialize();
+        }
 
     ~Interpret();
 
--- a/src/api/Interpret.h
+++ b/src/api/Interpret.h
@@ -135,7 +135,6 @@ class Interpret {
     std::unique_ptr<MainSolver> main_solver;
 
     bool            f_exit;
-    bool            ownsLogic;
 
     // Named terms for getting variable values
     MapWithKeys<const char*,PTRef,StringHash,Equal<const char*>> nameToTerm;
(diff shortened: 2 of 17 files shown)


blishko:
Hi Daniel,
I will check the proposed changes.
Some of them look reasonable, some we cannot apply. I will try to give concrete comments later.

blishko:
Hi @daniel-raffler,
I went over the commit and here are my comments:

The following commits can be merged immediately: c1367fc, 46efc1f, 5aa3220, 65a4a4d, d85f529, bae494f, db08146 
Regarding bae494f,  why does it mention race condition? I am not sure if JavaSMT is not trying to use OpenSMT in some unexpected way. I general, OpenSMT is not thread-safe, because `Logic` is not thread-safe.

The following commits can be merged later after some clarifications: 
3a1bd3c -> See the comment above above thread safety. Also, if needed, maybe the pool should be `thread_local`?
526d1ff -> Not sure we want to expose this. What is the actual use case in JavaSMT?
0d955b3 -> This currently does not cover all characters marked as quotable before. Needs checking which version is correct.
f86ef99 -> Why is JavaSMT sensitive to whitespaces in the formula?

The following commits will most likely not be merged:
9ff747e - I have to investigate again, but trying to divide by zero does not seem to me to fit `LANonLinearException`. 
e5eb49b - Why is JavaSMT making this assumption?
1ae6019, b8457e2  - There should be a better design for this. Which we need to implement on our side. it would be good if you can open a separate issue and describe the use case and expectations you have from this method.

daniel-raffler:
Hello Martin,
thanks for your review! Regarding the commits you mentioned:

> Regarding  https://github.com/usi-verification-and-security/opensmt/commit/bae494f2e605c27f63ab68b571251f608d0794e4, why does it mention race condition? I am not sure if JavaSMT is not trying to use OpenSMT in some unexpected way. I general, OpenSMT is not thread-safe, because Logic is not thread-safe.

> https://github.com/usi-verification-and-security/opensmt/commit/3a1bd3c4ab437fd72403a92abcdf616d9eea8d68 -> See the comment above above thread safety. Also, if needed, maybe the pool should be thread_local?

About these two commits: they are both linked to a failed JavaSMT test that can be found [here](https://github.com/sosy-lab/java-smt/blob/master/src/org/sosy_lab/java_smt/test/SolverConcurrencyTest.java#L428) along with the description:

>Test that starts multiple threads that solve several problems such that they tend to not finish
at the same time. Once a problem finishes, the formula is then send to another thread (with the
context) and the sending thread starts solving a new problem (either send to it, or by
generating a new one). The thread that just got the formula + context translates the formula
into its own context once its finished (and send its formula to another context) and starts
solving the formula it got send. This test is supposed to show the feasability of translating
formulas with contexts in (concurrent) use.

As far a I can tell the test should work for OpenSMT as Logic instances are never used by more than one thread at a time. Instead the formulas are just copied from one context to the next with synchronization between the threads. However, it still fails as mpgPool is global and shared by all Logic objects, even if they are unrelated and don't share any formulas. In my opinion the pool should either be "per Logic instance", or it needs to be synchronized. I'm not sure if a thread local pool would be enough, especially since Logic instances may move from one thread to another.

> https://github.com/usi-verification-and-security/opensmt/commit/526d1fff98c19dc5998960ef7cefafb6686ff50a -> Not sure we want to expose this. What is the actual use case in JavaSMT?

JavaSMT offers some type safety and needs to be able to extract the sort of a term. For terms with array sort this means that we need access to the sorts of the indices and the elements of the array to recursively translate the (OpenSMT) sort into a JavaSMT FormulaType. You can find the code [here](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/solvers/opensmt/OpenSmtFormulaCreator.java#L173). This is in fact the only time the _getSortDefinition(SRef)_ method is used and it would also have been possible to add two methods, _SRef getArraySortIndex(SRef)_ and _SRef getArraySortElements(SRef)_, instead.

> https://github.com/usi-verification-and-security/opensmt/commit/0d955b394b1a6f0823cde3e53bf71060c42c7ca4 -> This currently does not cover all characters marked as quotable before. Needs checking which version is correct.

Are you refering to the line with "(, ), <,>" ? As far as I know we don't have to change the quotable characters.

@kfriedberger: Maybe you could comment on this?

> https://github.com/usi-verification-and-security/opensmt/commit/9ff747e9247238fceaf1b662a102a97cb2fb48c8 - I have to investigate again, but trying to divide by zero does not seem to me to fit LANonLinearException.

Sorry, I messed up the commit message for this one. And there probably should have been two commits.

The change in line 810 makes sure that LANonLinearException is thrown for x%a if **a** is anything other than a constant. This is more in line with the rest of the code and makes it easier for us to catch the exception. The second change in line 832 is unrelated and makes terms of the form 1/a illegal. This is probably linked to a failed test case in JavaSMT, but I'll have to go back and check..

> https://github.com/usi-verification-and-security/opensmt/commit/1ae601976458e9f2740a39462e5787a16a966a05, https://github.com/usi-verification-and-security/opensmt/commit/b8457e26ec499b012620fe7574e485dcf0177d75 - There should be a better design for this. Which we need to implement on our side. it would be good if you can open a separate issue and describe the use case and expectations you have from this method.

Ok, I've opened a new issue [here](https://github.com/usi-verification-and-security/opensmt/issues/648).

(All other commits you mentioned were added to fix failed tests in JavaSMT. I'll go back and look up the exact details and will then come back to you.)

Kind regards,
Daniel

daniel-raffler:
I've now been able to re-run the test suite for the remaining commits:

> https://github.com/usi-verification-and-security/opensmt/commit/f86ef99a73f7c0b5efa79bf547b30228d2d9e44b - Why is JavaSMT sensitive to whitespaces in the formula?

Removing this patch breaks 6 tests in [SolverFormulaIOTest](https://github.com/sosy-lab/java-smt/blob/master/src/org/sosy_lab/java_smt/test/SolverFormulaIOTest.java). The issue is with a helper method ([here](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/SolverFormulaIOTest.java#L409)) that is supposed to make sure that the final assert is not followed by any other commands. Unfortunately the method breaks for OpenSMT as it doesn't expect any newlines within the (assert ..) statement.

This is something we could easily fix and then we won't need this patch.

> https://github.com/usi-verification-and-security/opensmt/commit/9ff747e9247238fceaf1b662a102a97cb2fb48c8 - I have to investigate again, but trying to divide by zero does not seem to me to fit LANonLinearException.

Removing this patch breaks [testDivision](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/NonLinearArithmeticTest.java#L243) and [testDivisionUnsatisfiable](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/NonLinearArithmeticTest.java#L258) in NonLinearArithmeticTest. For both of these test OpenSMT seems to crash in Logic.check(). The issue is related to the change in line 823 and I don't think the other part of the patch is needed.

Does OpenSMT actually support terms of the form 1/a?

> https://github.com/usi-verification-and-security/opensmt/commit/e5eb49b1b06890c9fc08919ecea91deea1a988ea - Why is JavaSMT making this assumption?

There is a test [testGetBooleansEvaluation](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/ModelEvaluationTest.java#L135) that specifically tests for this. As Karlheinz pointed out in https://github.com/sosy-lab/java-smt/pull/302#discussion_r1349488825 this could, however, easily be disabled for OpenSMT and then we won't need this patch.

Should I look up any more details on the other commits?

blishko:
Hi Daniel,
Thanks for the detailed explanations!

> As far a I can tell the test should work for OpenSMT as Logic instances are never used by more than one thread at a time. Instead the formulas are just copied from one context to the next with synchronization between the threads. However, it still fails as mpgPool is global and shared by all Logic objects, even if they are unrelated and don't share any formulas. In my opinion the pool should either be "per Logic instance", or it needs to be synchronized. I'm not sure if a thread local pool would be enough, especially since Logic instances may move from one thread to another.

Interesting! I believe you are right. We have not been considering threads in OpenSMT yet, so this again will need larger changes in the design I think. I suggest to keep 3a1bd3c  in your OpenSMT patch and revert in this PR.

>> https://github.com/usi-verification-and-security/opensmt/commit/9ff747e9247238fceaf1b662a102a97cb2fb48c8 - I have to investigate again, but trying to divide by zero does not seem to me to fit LANonLinearException.

>Removing this patch breaks [testDivision](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/NonLinearArithmeticTest.java#L243) and [testDivisionUnsatisfiable](https://github.com/sosy-lab/java-smt/blob/45cb3e09ad7818c2fa45bce38860cc4d4fb69887/src/org/sosy_lab/java_smt/test/NonLinearArithmeticTest.java#L258) in NonLinearArithmeticTest. For both of these test OpenSMT seems to crash in Logic.check(). The issue is related to the change in line 823 and I don't think the other part of the patch is needed.
Does OpenSMT actually support terms of the form 1/a?

I believe you are correct, we can include this change after the fix you have added in [e42cb33](https://github.com/usi-verification-and-security/opensmt/pull/646/commits/e42cb3300242f3f23e6e7ad938701d41ebeb94b1).

Regading the quotable characters, looking into the SMT-LIB standard, it seems some of the characters are wrong. I think I need to fix this separately from this PR. I created in issue for that: https://github.com/usi-verification-and-security/opensmt/issues/649

I suggest to keep 526d1ff in your patch and revert here. We need to decide what is the best way to expose this information.

For f86ef99, I don't have a strong opinion, but I don't see why JavaSMT should require that assertions are printed on one line. Seems like too strong a requirement. So I am inclining towards rejecting this change in this PR.

I think that's all the commits. Thus we can add the changes dealing with division to my original list of acceptable changes. The rest should not be included for now in this PR. We will work on some of the issues you have raised here, with the priority on fixing the quotable characters.

daniel-raffler:
Hello Martin,
Thanks again for your feedback! 
I think I've now reverted all the commits you mentioned.  If you could accept the pull request as is this would already simplify our API patch a lot!

blishko (approved):
LGTM!

blishko:
Thanks Daniel for your work!

Issue #649, opened 2023-10-24
Fix detection of characters that require quoting according to SMT-LIB


blishko:
Looks like characters that we consider as requiring quotes do not match the list in SMT-LIB specification (paragraph Symbols on page 23)

blishko:
Fixed by #651

PR #651, opened 2023-10-29
Logic: Make test for quotes more transparent

blishko:
Instead of using a hard-to-read table, let's follow the standard to the letter. The standard defines which characters are permitted in the simple symbol (which does not require quotes). If we encounter a character not in the specified set, we need to add quotes.

The reason for this change has been moivated by compilation problem in JavaSMT and the fact that even though the table was created with length 256, only the first 250 entries were specified.


blishko:
Fixes #649.

blishko:
See also the discussion in JavaSMT: https://github.com/sosy-lab/java-smt/pull/302#discussion_r1349656525

aehyvari (approved):
LGTM!

PR #652, opened 2023-11-05
Logic: Make API for getting array sort more robust

blishko:
Creating array sorts should be allowed only for logics that support arrays.


BritikovKI (approved):
Lgtmt

PR #656, opened 2023-12-25
Support `ALL` as a valid argument to `set-logic`

### LINKED ISSUE #654: Support for `(set-logic ALL)`
Looks like opensmt doesn't support ALL for logic selection:

```
$ cat a.smt2
(set-logic ALL)
$ opensmt a.smt2
(error "unknown logic ALL")
```

As specified in Section 4.2.1 of https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf, `ALL` is a valid parameter:

>The argument l can also be the predefined symbol ALL. With this argument, the solver sets the logic to the most general logic it supports.(35) Note that while the reaction to (set-logic ALL) is the same for every compliant solver, the chosen logic is solver- specific.

This is used by upstream tools to simplify logic specifications; it'd be nice if opensmt supported it.

blishko:
We define `ALL` to be an alias of `QF_AUFLIRA`.
Note, however, that while `opensmt` can work with either reals or integers, the solver
does not yet support the combination of both.

**WARNING:** We do not check if user is trying to use a combination of reals and integers.
Doing so will result in undefined behaviour in the solver, most likely a crash at some point.

Resolves #654


aehyvari (approved):
LGTM!

PR #657, opened 2024-01-26
Getters of numerator and denominator for `FastRational`

Tomaqa:
I noticed that it may be the first time using `std::optional` in the project but I hope it won't cause any trouble..

- `force_ensure_mpq_valid` is `const`, uses `const_cast`, but may modify non-mutable data members of `FastRational`, which I am afraid results in [undefined behavior](https://en.cppreference.com/w/cpp/language/const_cast)
- I also changed `FastRational` to `opensmt::Number` in `ArithLogic.h` to maintain consistency - the header includes only the abstraction `Number.h` and is using it in the declaration of several functions but not all. I also changed that in most cases in `ArithLogic.cc`, but not in all. I observed that many pieces of code in OpenSMT assume that `FastRational` is being used in the place of `Number`, so changing it to `mpq_class` would break a lot of code. Therefore, I just focused on most parts of `ArithLogic`, at least to avoid confusion for guys who are not that familiar with the internals (for example me :smiling_face_with_tear:).

### DIFF
--- a/src/common/FastRational.h
+++ b/src/common/FastRational.h
@@ -13,6 +13,7 @@ Copyright (c) 2008, 2009 Centre national de la recherche scientifique (CNRS)
 #include "Vec.h"
 #include <stack>
 #include <vector>
+#include <optional>
 
 typedef int32_t  word;
 typedef uint32_t uword;
@@ -165,6 +166,7 @@ private:
     }
     void force_ensure_mpq_valid() const
     {
+        // TK: I am afraid that this is UB
         const_cast<FastRational *>(this)->ensure_mpq_valid();
     }
     void ensure_mpq_memory_allocated()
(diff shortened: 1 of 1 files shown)


blishko:
Can we split this into two PRs? One for adding the getters, and a separate one for the name change in `ArithLogic`?

And you are right about the UB, that's unfortunate. We should come up with a better design then...

aehyvari:
Great work @Tomaqa, this is awesome!  I didn't know we had UB with this design but it's really good you found it!

src/common/FastRational.h
```diff
@@ -245,6 +247,16 @@ class FastRational
         }
     }
 
+    std::optional<std::pair<word, uword>> tryGetNumDen() const {
+        if (!wordPartValid()) return {};
+        return std::make_pair(num, den);
+    }
+
+    mpq_ptr getMpq() const {
```
blishko, line 255: I am wondering if this is the right API?
Should we instead return a value, i.e., mpq_class?
So that the user does not do anything stupid like calling `mpq_clear` on the returned `mpq_ptr` and with that corrupting our storage.
It would also allow us to *not* use the problematic `force_ensure_mpq_valid` method.

Tomaqa: I agree, I somehow thought that `mpq_ptr` is sth. like `unique_ptr` but it is just a pointer without taking the ownership..
It seems to me that `mpq_class` should be a good choice, but I am not sure how to construct it. I tried to look up a constructor from `mpq_ptr` but I failed - although I found that it is used at `FastRational.cc:84`. But I do not know what exactly it does. Still it seems that it requires `mpqPartValid()`, so I suppose that `force_ensure_mpq_valid` is still necessary.

Tomaqa:
> And you are right about the UB, that's unfortunate. We should come up with a better design then...

#662

blishko (approved):
LGTM!

PR #659, opened 2024-01-26
Got rid of useless (actually possibly harmful) default destructors

Tomaqa:
I experienced in my project that some classes in OpenSMT have implicitly deleted move semantics. I found out that because my class forbids copying and requires moving, but class `Opensmt` was not able to.

I observed that this is caused by a user-defined destructor, referring to [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three): "the presence of a user-defined (or = default or = delete declared) destructor, ... prevents implicit definition of the move constructor and the move assignment operator". This also means that wherever move semantics are desired (e.g. using `std::move`), the object is actually still copied because move constructor is not available. Therefore, performance can be possibly improved: "failing to provide move constructor and move assignment is usually not an error, but a missed optimization opportunity".

I found a lot of user-defined destructors that are empty, i.e. that do the same thing as would the implicitly defined destructor, but which would not prevent implicit def. of move constructor. I got rid of such destructors, but only in cases where it did not break semantics of the code. For example, destructors should be defined as `virtual` in virtual base classes.

In cases where I did not delete the destructor, I did not fix the possible problem by providing the missing move constructor and `operator=`, because I was lazy, and also since I do not know all the classes and I am not sure if move semantics should be allowed in all cases.

As a last comment, there are a lot of classes with user-defined destructor but with no explicit definitions of copy semantics. This is almost always a bad practice, as discussed in the [rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three). If these classes are not intended to be copied (which I did not investigate), the copy semantics should be explicitly deleted. Otherwise, copying will usually result in a mess.

### DIFF
--- a/src/api/Opensmt.cc
+++ b/src/api/Opensmt.cc
@@ -62,8 +62,3 @@ Opensmt::Opensmt(opensmt_logic logic_, const char* name, std::unique_ptr<SMTConf
     mainSolver = std::unique_ptr<MainSolver>(new MainSolver(*logic, *this->config, name));
     mainSolver->initialize();
 }
-
-Opensmt::~Opensmt()
-{
-
-}
--- a/src/api/Opensmt.h
+++ b/src/api/Opensmt.h
@@ -45,7 +45,6 @@ public:
      * @param config Configuration for the OpenSMT instance
      */
     Opensmt(opensmt_logic _logic, const char* name, std::unique_ptr<SMTConfig> config);
-    ~Opensmt();
 
     SMTConfig& getConfig() { return *config; }
     Logic& getLogic() { return *logic; }
--- a/src/cnfizers/Tseitin.h
+++ b/src/cnfizers/Tseitin.h
@@ -48,9 +48,6 @@ public:
         , alreadyCnfized(logic_.getTerm_true())
       {}
 
-    ~Tseitin( ) { }
-
-
 private:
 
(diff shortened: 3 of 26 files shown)


blishko (commented):
I think this is good!

Thanks for going through the code!

I noticed that you are also changing files that are have not been in use in OpenSMT for a long time.
I think we should remove them in a separate PR and avoid changing them here.
These are all the files you have touched in the folder `simplifiers` in this PR.
Note that some files in that folder are still in use, you can see which files need to be kept in `simplifiers/CMakeLists.txt`.

Tomaqa:
I reverted all from `src/simplifiers`

Tomaqa:
.. and rebased the branch on the current master

blishko:
Can you also not touch `symmetry/bliss/graph.*`? Those should also be removed completely. Sorry!

aehyvari:
Great cleanup also here!  I second @blishko , the bilss code was an old experiment that we wanted to do to remove symmetries in the solver for uninterpreted functions.  The project has been on hold for a long time after a good start.  If we want to pick this project up again (there are some evidence that it helps), we shouldn't include the code this way to the repository.

Tomaqa:
Sure, I removed it.

blishko (approved):
LGTM!

PR #661, opened 2024-01-29
Provided move semantics for `Opensmt`

Tomaqa:
Allows to move the main solver in applications that require it (e.g. if they use `std::unique_ptr<Opensmt>` and do stuff with it). This required to provide move semantics in several nested/used classes, using the rule of five.


blishko (approved):
LGTM!

PR #663, opened 2024-01-30
FastRational: Eliminate potential UB

### LINKED ISSUE #662: UB in `FastRational::force_ensure_mpq_valid`
`const_cast` in `FastRational::force_ensure_mpq_valid` results in UB.
Instead of `const_cast`, either the function must be made non-const, which would break some existing code.
Or, it should be sufficient to make `mpq` `mutable`. Since the variable is not public _and_ it should probably serve as just a fallback option, this solution should not cause much harm. Non-public const functions just should take care whether they really should modify `mpq` or not.

blishko:
Casting away const and then modifying the object is UB if the original object was const.
From [cppreference](https://en.cppreference.com/w/cpp/language/const_cast):
"Modifying a const object through a non-const access path results in undefined behavior."

We fix the problem by making `mpq` and `state` mutable, as we can change the inner representaton in certain circumstances without changing the rational value represented by the object.

Fixes #662.

### DIFF
--- a/src/common/FastRational.cc
+++ b/src/common/FastRational.cc
@@ -115,8 +115,8 @@ FastRational gcd(FastRational const & a, FastRational const & b)
         return FastRational(gcd(a.num, b.num));
     }
     else {
-        a.force_ensure_mpq_valid();
-        b.force_ensure_mpq_valid();
+        a.ensure_mpq_valid();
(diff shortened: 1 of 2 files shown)


aehyvari (approved):
LGTM!

Tomaqa (approved):
LGTM :)

PR #670, opened 2024-02-07
664 explicitly divide debug and release builds

### LINKED ISSUE #664: Explicitly divide `debug` and `release` builds
Currently, the debug/release mode must be specified in CMake and then `build` contains only this single mode. I consider this as inconvenient, for example in cases when I want to quickly switch from one mode to another.

I suggest resolving this the same way I dealt with it in my previous project: to explicitly divide builds `release` and `debug` in order to make them independent. Some IDEs use such architecture. Concrete solutions may differ, but in the first place, I would like to discuss the idea in general.

I am not experienced in CMake, just in GNU Make, so I am not sure if it is possible to easily arrive at something like this there. But I hope so.

Tomaqa:
Root Makefile that serves only for initial builds.

`make` (or `make default` or `make release`) builds `build` and `make debug` builds `build-test`.
`make all` builds both.
Once the build directory is created, the rule is "up-to-date" and not run anymore. This is not a mistake but my intention.

The build directories can also be removed using `make clean` (or `make clean-release`) and `make clean-debug` (or `make clean-all`). Then it is possible to run e.g. `make` again.

The target build directories can be modified via cmdline, for example: `make RELEASE_BUILD_DIR=build-release`.
But then the option should also be used for `make clean` otherwise it will try to remove the default build directory (it won't fail but won't do anything).

More importantly, CMake options may be modified using `CMAKE_FLAGS` option, for example:
`make CMAKE_FLAGS=-DENABLE_LINE_EDITING:BOOL=ON`
`make debug CMAKE_FLAGS='-DENABLE_LINE_EDITING:BOOL=ON -DUSE_READLINE:BOOL=ON'`

The Makefile is not documented yet, README is not updated.

Resolves #664.

I think that the `install` rules should be handled too, but here I am not sure how to handle the target install directory for the library (`include` is the same for both) - should there also be a separate directory for each build type (e.g. `/usr/local/lib/release/libopensmt.a` and `/usr/local/lib/debug/libopensmt.a`, or `/usr/local/lib/libopensmt.a` and `/usr/local/lib/debug/libopensmt.a`), or should it be just one and always overwrite the previous file? In order to change the build type. I think that on the system level, at least on Linux distros, only one type at a time is usually supported.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,16 @@ BUILD_DIR_BASE = build
 RELEASE_BUILD_DIR = $(BUILD_DIR_BASE)
 DEBUG_BUILD_DIR = $(BUILD_DIR_BASE)-debug
 
-CMAKE_FLAGS =
+INSTALL_DIR = /usr/local
+
+CMAKE_FLAGS += -DCMAKE_INSTALL_PREFIX=$(INSTALL_DIR)
 RELEASE_CMAKE_FLAGS = $(CMAKE_FLAGS)
 DEBUG_CMAKE_FLAGS = $(CMAKE_FLAGS) -DCMAKE_BUILD_TYPE=Debug
 
-.PHONY: default all release debug _dir-release _dir-debug _build-release _build-debug clean clean-all clean-release clean-debug
+################################################################
+
+.PHONY: default all release debug
+.PHONY: _dir-release _dir-debug _build-release _build-debug
 
 default: release
 
@@ -32,6 +37,10 @@ _build-debug: | $(DEBUG_BUILD_DIR)
 $(DEBUG_BUILD_DIR):
 	cmake -B $(DEBUG_BUILD_DIR) $(DEBUG_CMAKE_FLAGS)
 
+################################
+
+.PHONY: clean clean-all clean-release clean-debug
+
 clean: clean-release
 
 clean-all: clean-release clean-debug
@@ -41,3 +50,15 @@ clean-release:
 
 clean-debug:
 	rm -fr $(DEBUG_BUILD_DIR)
+
+################################
+
+.PHONY: install install-release install-debug
+
+install: install-release
+
+install-release: release
+	$(MAKE) -C $(RELEASE_BUILD_DIR) install
+
+install-debug: debug
+	$(MAKE) -C $(DEBUG_BUILD_DIR) install
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,34 @@
 BUILD_DIR_BASE = build
+## These can be customized via command line
 RELEASE_BUILD_DIR = $(BUILD_DIR_BASE)
 DEBUG_BUILD_DIR = $(BUILD_DIR_BASE)-debug
 
 DEFAULT_INSTALL_DIR := /usr/local
 USER_INSTALL_DIR := $(INSTALL_DIR)
+## This can be customized via command line,
+## it takes effect within both `cmake --build` and `cmake --install`
 INSTALL_DIR = $(DEFAULT_INSTALL_DIR)
 
+## These can be customized via command line
+## Arguments for `cmake -B <build_dir>`
 CMAKE_FLAGS += -DCMAKE_INSTALL_PREFIX=$(INSTALL_DIR)
-RELEASE_CMAKE_FLAGS = $(CMAKE_FLAGS)
-DEBUG_CMAKE_FLAGS = $(CMAKE_FLAGS) -DCMAKE_BUILD_TYPE=Debug
+RELEASE_CMAKE_FLAGS += $(CMAKE_FLAGS)
+DEBUG_CMAKE_FLAGS += $(CMAKE_FLAGS) -DCMAKE_BUILD_TYPE=Debug
 
+## These can be customized via command line
+## Arguments for `cmake --build <build_dir>`
 CMAKE_BUILD_FLAGS +=
-RELEASE_CMAKE_BUILD_FLAGS = $(CMAKE_BUILD_FLAGS)
-DEBUG_CMAKE_BUILD_FLAGS = $(CMAKE_BUILD_FLAGS)
+RELEASE_CMAKE_BUILD_FLAGS += $(CMAKE_BUILD_FLAGS)
+DEBUG_CMAKE_BUILD_FLAGS += $(CMAKE_BUILD_FLAGS)
 
+## These can be customized via command line
+## Arguments for `cmake --install <build_dir>`
 CMAKE_INSTALL_FLAGS +=
 ifneq ($(USER_INSTALL_DIR),)
   CMAKE_INSTALL_FLAGS += --prefix=$(USER_INSTALL_DIR)
 endif
-RELEASE_CMAKE_INSTALL_FLAGS = $(CMAKE_INSTALL_FLAGS)
-DEBUG_CMAKE_INSTALL_FLAGS = $(CMAKE_INSTALL_FLAGS)
+RELEASE_CMAKE_INSTALL_FLAGS += $(CMAKE_INSTALL_FLAGS)
+DEBUG_CMAKE_INSTALL_FLAGS += $(CMAKE_INSTALL_FLAGS)
 
 ################################################################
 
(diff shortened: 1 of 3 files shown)


Tomaqa:
I am wondering if this is connected to issue #664 automatically or if it is necessary to do so manually ..?

blishko:
> I am wondering if this is connected to issue #664 automatically or if it is necessary to do so manually ..?

If you add "Resolves #664" to the description, the issue will be automatically closed when the PR is accepted.
(You can also use [different phrases](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)).

blishko:
> `make` (or `make default` or `make release`) builds `build` and `make debug` builds `build-test`.

This should say that `make debug` builds `build-debug`.

blishko (commented):
Looks good!

We should also update the README to document this.
Since this only works if the directory does not exist, if I want to rebuild the project after some change, I still have to do it explicitly with `cmake --build` or in the build directory invoking the build system such as `make` or `ninja`, right?
So it is only a shortcut when I want to build the project for the first time?

Tomaqa:
> Since this only works if the directory does not exist, if I want to rebuild the project after some change, I still have to do it explicitly with `cmake --build` or in the build directory invoking the build system such as `make` or `ninja`, right? So it is only a shortcut when I want to build the project for the first time?

Yes. It can run either always or only once, but nothing in between - unless all source code prerequisites were supplied, which would duplicate the purpose of makefiles in `build*`.
Nonetheless, you can always run `make *` with `-B` which skips all prerequisite checks and runs all corresponding rules. This can be used for a "rebuild".
... However, I noticed now that doing so also results in recompiling all related sources, contrary to what e.g. `cmake --build` does.
This is not too convenient.

Therefore, I may change it s.t. the `Makefile` invokes `cmake -B *` only once (i.e. every time the build directory does not exist yet) but runs `cmake --build *` every time.
Do you agree? Or do you have better idea?

blishko:
> Therefore, I may change it s.t. the Makefile invokes cmake -B * only once (i.e. every time the build directory does not exist yet) but runs cmake --build * every time.
Do you agree? Or do you have better idea?

Yes, I think that would be good.

blishko (dismissed):
LGTM!

I tested it briefly and it seems to work nicely!

Tomaqa:
Cool.

I still have this pending question. I think it should also be included in this root makefile. What do you think?

> I think that the `install` rules should be handled too, but here I am not sure how to handle the target install directory for the library (`include` is the same for both) - should there also be a separate directory for each build type (e.g. `/usr/local/lib/release/libopensmt.a` and `/usr/local/lib/debug/libopensmt.a`, or `/usr/local/lib/libopensmt.a` and `/usr/local/lib/debug/libopensmt.a`), or should it be just one and always overwrite the previous file? In order to change the build type. I think that on the system level, at least on Linux distros, only one type at a time is usually supported.

blishko:
I think there should only be one target install directory for both.
I guess here you have to explicitly specified which configuration you want to install.

Tomaqa:
I believe it is done. The installation dir can be customized either within `make` or be overridden within `make install` via `INSTALL_DIR` variable in both cases.
Plain `make install` works as well but only if it does not require sudo, otherwise everything is compiled with root privileges which is not neat. I would not even document this option and would just write to use `make && make install` (or `make && sudo make install`).
Another problem is that it is not possible to combine root and non-root installations (regardless of the build): `sudo make install` and `make install INSTALL_DIR=local_dir` fails because the previous call to CMake still produces some root files. In this case, `make clean` is necessary.
This can still be avoided by using two separate build dirs, one for root installations and the other one for local ones.

Anyway, the basic setups work perfectly fine. For example:
`make && sudo make install`
`make && make install INSTALL_DIR=local_dir`
`make INSTALL_DIR=local_dir && make install`
`make install INSTALL_DIR=local_dir`
`make INSTALL_DIR=local_dir && make install && make install INSTALL_DIR=install_dir2`

If this is fine, I will update the `README` and it should be completed.

Tomaqa:
.. and if you want to install the debug build, `install-debug` must be used instead of `install` (or `install-release`).

Makefile
```diff
@@ -0,0 +1,77 @@
...
+DEBUG_CMAKE_FLAGS = $(CMAKE_FLAGS) -DCMAKE_BUILD_TYPE=Debug
+
+CMAKE_BUILD_FLAGS +=
+RELEASE_CMAKE_BUILD_FLAGS = $(CMAKE_BUILD_FLAGS)
+DEBUG_CMAKE_BUILD_FLAGS = $(CMAKE_BUILD_FLAGS)
+
+CMAKE_INSTALL_FLAGS +=
+ifneq ($(USER_INSTALL_DIR),)
+  CMAKE_INSTALL_FLAGS += --prefix=$(USER_INSTALL_DIR)
+endif
```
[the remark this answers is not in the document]
Tomaqa: Resolved in commit e3e1304

blishko (dismissed):

Tomaqa:
Updated README.

The last thing is that the executable is for some reason placed at `build/opensmt` and not at `build/bin/opensmt` - inconsistently with the installation. Also it is weird that `build/bin` is actually present but it is empty.
However, I have no experience with CMake and I do not know how to achieve that. But I suppose it is easy, when one knows what to do.

Tomaqa:
Maybe connected to the above, there's this weird `bin/opensmt.C`. I suppose it is some relict and that `src/bin/opensmt.cc` is currently being used.

README.md
```diff
@@ -70,13 +101,54 @@ $ ctest
...
+```
+make debug INSTALL_DIR=local_dir-debug && make install-debug
+```
+
+Build (and configure) with the default values but install into `local_dir`:
+```
+make && make install INSTALL_DIR=local_dir
+```
+
+Build and configure to by default install into `local_dir` but also install into `local_dir2`:
```
blishko, line 148: I am not sure how to parse this sentence.

README.md
```diff
@@ -114,6 +186,4 @@ When using OpenSMT as a library, the option needs to be set in `SMTConfig` **bef
 Interpolation is supported for SMT-LIB logics `QF_UF`, `QF_LRA`, and `QF_LIA` in both single-query and incremental mode. An example of how SMT-LIB2 file can be extended to instruct OpenSMT to compute interpolants can be found [here](regression_itp/itp_bug_small.smt2).
 
 ## Contact
-If you have questions please mail them to me at antti.hyvarinen@gmail.com, or at [github](https://github.com/usi-verification-and-security/opensmt)
-
-
+If you have questions, bug reports, or feature requests, please refer to our [GitHub](https://github.com/usi-verification-and-security/opensmt/issues) issue tracker or send us a mail at tomaqa@gmail.com or antti.hyvarinen@gmail.com.
```
blishko, line 189: Since we are making changes here, can you add my email as well?
`martin.blicha@gmail.com`

blishko (commented):
Looks good!

We can consider if build and install instructions deserve a separate README now :)

blishko:
> The last thing is that the executable is for some reason placed at build/opensmt and not at build/bin/opensmt - inconsistently with the installation (e.g. /usr/local/bin/opensmt). Also it is weird that build/bin is actually present but it is empty.

This was by design. Or at least for me it was more comfortable.
But we can move it to `bin` folder.
And yes, there is some old file, I think the cleanup of `bin` directory in `src` should be a separate PR.

README.md
Tomaqa: Another try:
Build and configure to install into `local_dir` and install there, but also perform another installation into `local_dir2`

README.md
Tomaqa: Sure. Should I also include you into the copyright at the top?

Tomaqa:
> And yes, there is some old file, I think the cleanup of `bin` directory in `src` should be a separate PR.
#671

README.md
Tomaqa: And I suppose that `AUTHORS` should be updated too ..

blishko: Yes that would be good as well.
I have never actually looked into the AUTHORS file :D
Maybe we can leave that to separate PR as well, so this one does not lose the focus on the Makefile.

Tomaqa: Done.

Tomaqa:
Should be ready to merge.

blishko (approved):

PR #673, opened 2024-02-09
updated authors (also in the license)

### LINKED ISSUE #672: update `AUTHORS`

Tomaqa:
resolves #672

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,18 +5,20 @@
   Led by Prof. Natasha Sharygina <natasha.sharygina@usi.ch>
 --------------------------------------------------------------
 
-Project leader and main code designer
+Project leaders and main code designers
 -------------------------------------
-Dr. Antti Hyvarinen           <antti.hyvaerinen@usi.ch>
+Dr. Tomas Kolarik           <tomas.kolarik@usi.ch>  
+Dr. Martin Blicha           <martin.blicha@usi.ch>
 
-Previous project leader and main code designer
+Previous project leaders and main code designers
 -------------------------------------
+Dr. Antti Hyvarinen          <antti.hyvaerinen@usi.ch>  
 Dr. Roberto Bruttomesso      <roberto.bruttomesso@usi.ch>
 
 Previous members
 ----------------
-Dr. Aliaksei Tsitovich       <aliaksei.tsitovich@usi.ch>
-Dr. Simone Fulvio Rollini    <simone.fulvio.rollini@usi.ch>
+Dr. Aliaksei Tsitovich       <aliaksei.tsitovich@usi.ch>  
+Dr. Simone Fulvio Rollini    <simone.fulvio.rollini@usi.ch>  
 Edgar Pek                    <edgar.pek@usi.ch>
 
 Acknowledgments
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,19 +7,31 @@
 
 Project leaders and main code designers
 -------------------------------------
-Dr. Tomas Kolarik           <tomas.kolarik@usi.ch>  
+Dr. Tomas Kolarik           <tomas.kolarik@usi.ch>
 Dr. Martin Blicha           <martin.blicha@usi.ch>
 
 Previous project leaders and main code designers
 -------------------------------------
-Dr. Antti Hyvarinen          <antti.hyvaerinen@usi.ch>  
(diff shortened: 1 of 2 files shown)


Tomaqa:
It is though a bit confusing that there is `LICENSE` (MIT) and `COPYING` (GNU/GPL). We declare in `README` that we use MIT license, but under some circumstances that it is GPL. Is this the reason why both are included?
In any case, it is confusing ...

AUTHORS
```diff
@@ -5,18 +5,20 @@
...
 -------------------------------------
+Dr. Antti Hyvarinen          <antti.hyvaerinen@usi.ch>  
 Dr. Roberto Bruttomesso      <roberto.bruttomesso@usi.ch>
 
 Previous members
 ----------------
-Dr. Aliaksei Tsitovich       <aliaksei.tsitovich@usi.ch>
-Dr. Simone Fulvio Rollini    <simone.fulvio.rollini@usi.ch>
+Dr. Aliaksei Tsitovich       <aliaksei.tsitovich@usi.ch>  
+Dr. Simone Fulvio Rollini    <simone.fulvio.rollini@usi.ch>
```
blishko, line 21: Is this adding extra space at the end of the line?
Probably should not do that, no?

blishko (commented):
We could also add several previous members here...

blishko:
> It is though a bit confusing that there is `LICENSE` (MIT) and `COPYING` (GNU/GPL). We declare in `README` that we use MIT license, but under some circumstances that it is GPL. Is this the reason why both are included? In any case, it is confusing ...

I vote to just remove `COPYING` file.
Regarding the mention of GPL, the problem is the `readline` library, which is GPL licensed, and apparently, when you link with GPL library, you have to be under GPL license as well.
Here I also vote to stop using `readline` altogether. I don't think anyone ever used that feature (except Antti :))
And for that functionality, we have support `editline` alternative.

AUTHORS
Tomaqa: It was my intention, this syntax tells a Markdown renderer to add linebreak after such lines (otherwise it does not). Although it is just a text file it can still be nicely rendered this way ... I admit that the syntax is weird in general, and that Markdown is just not necessary here.

Tomaqa:
I just discovered the functionality and I think it is nice. But I am not sure when `editline` is used, is it with `make CMAKE_FLAGS=-DENABLE_LINE_EDITING:BOOL=ON` ? On the first look, in seems to be the same as `make CMAKE_FLAGS='-DENABLE_LINE_EDITING:BOOL=ON -DUSE_READLINE:BOOL=ON'`.

Tomaqa:
> We could also add several previous members here...

I do not know which to add and which to not ...

aytey:
> I just discovered the functionality and I think it is nice. But I am not sure when `editline` is used, is it with `make CMAKE_FLAGS=-DENABLE_LINE_EDITING:BOOL=ON` ? On the first look, in seems to be functionally same as `make CMAKE_FLAGS='-DENABLE_LINE_EDITING:BOOL=ON -DUSE_READLINE:BOOL=ON'`.

https://github.com/usi-verification-and-security/opensmt/pull/120 changed OpenSMT such that `libedit` was the default, so you have to opt-in to `readline`, which avoids GPL issues.

Doing `cmake -DENABLE_LINE_EDITING:BOOL=ON` will build with `libedit`, while `cmake -DENABLE_LINE_EDITING:BOOL=ON -DUSE_READLINE:BOOL=ON` will build with `readline`.

AUTHORS
blishko: I don't think this was intended for Markdown.
But let's not spend too much time on this.
I would prefer without the spaces, but I am fine with this version as well.

blishko (approved):

blishko:
> I just discovered the functionality and I think it is nice. But I am not sure when `editline` is used, is it with `make CMAKE_FLAGS=-DENABLE_LINE_EDITING:BOOL=ON` ? On the first look, in seems to be functionally same as `make CMAKE_FLAGS='-DENABLE_LINE_EDITING:BOOL=ON -DUSE_READLINE:BOOL=ON'`.

I think the basic functionality should be the same. That's why I think we can drop `readline` completely.

Tomaqa:
Moved `readline` stuff to #675

blishko (approved):
LGTM!

PR #674, opened 2024-02-10
Remove old files that have not been used for a very long time

blishko:
These files have not been used in OpenSMT for several years now, let's remove them.


src/PySMT/opensmt_python.i
Tomaqa, line 1: Isn't this still useful for Python bindings? If someone wants to use OpenSMT in a Python app.

src/simplifiers/BVBooleanize.cc
Tomaqa, line 1: OK this just contains commented code, but can this be used for eager translation of bitvector variables to a boolean encoding?

src/PySMT/opensmt_python.i
blishko: I don't think it works anymore.

Tomaqa (commented):
I also see that within some `CMakeLists` contain obsolete `lrasolver` and `liasolver` within `tsolvers` which I guess are now superseded by `lasolver` ? I also do not see `dlsolver` and `emptysolver` being included.

src/simplifiers/BVBooleanize.cc
blishko: I think it is easier to write the translation anew rather than trying to adapt this file to the current codebase.

src/PySMT/opensmt_python.i
Tomaqa: Then I guess `get-python-wrappers.sh` should be removed as well.

Tomaqa: .. and `sample/` ?

Tomaqa:
I suggest removing `NEWS` as well.

src/PySMT/opensmt_python.i
blishko: Removed.

blishko:
I removed all of the mentioned files, and also `KNOWN_BUGS`, which was outdated as well.

Tomaqa (approved):
LGTM, but I am not sure if I am fully qualified to approve this..

PR #676, opened 2024-02-24
Rewriting: Do not collect rewritten arithmetic inequalities

blishko:
The information about which inequalities originated from arithmetic equalities were recorded during preprocessing. However, this makes it difficult to separate preprocessing passes and make them more independent, i.e., so we can run them on single formula as well as on group of formulas.
If this information is needed, it should be obtain in a different way. At the moment, this information is only used by the lookahead solver for partitioning, where the strategy is to not split on such inequalities. However, I do not see benefit in such strategy.
Since the need to record this information is blocking further refactoring that has the potential to unlock performance improvements in incremental solving, I propose to just not record this information anymore.


src/smtsolvers/LookaheadSMTSolver.h
```diff
@@ -72,7 +72,7 @@ class LookaheadSMTSolver : public SimpSMTSolver {
     laresult expandTree(LANode & n, std::unique_ptr<LANode> c1,
                         std::unique_ptr<LANode> c2); // Do lookahead.  On success write the new children to c1 and c2
     std::unique_ptr<LookaheadScore> score;
-    bool okToPartition(Var v) const { return theory_handler.getTheory().okToPartition(theory_handler.varToTerm(v)); };
+    bool okToPartition(Var) const { return true; }
```
BritikovKI, line 75: It is always Ok to partition?
What's the point in this function then?
Mb makes sense to remove it completely?

blishko: I think it may make sense to revisit this in the future (if someone will pick up parallel solving again).
I can imagine that it might make sense to not pick certain atoms for partitioning.
Currently, this was done only for the inequalities resulting from splitting arithmetic equalities.
I am not convinced these should be disallowed. Moreover, I would like to implement handling equalities and disequalities in LA engine directly, so there would be no more splitting equalities.

If I remove this function, I would have to update the partitioning algorithm as well, but I wanted to avoid that. I want to keep the changes to minimum.

BritikovKI: Ok, sounds legitimate

BritikovKI (approved):
Lgtmt

PR #677, opened 2024-03-05
MainSolver: Apply clang format

blishko:


src/api/MainSolver.cc
```diff
@@ -95,34 +82,29 @@ bool MainSolver::pop()
...
-        if (not isLastFrameUnsat()) {
-            getSMTSolver().restoreOK();
-        }
+        if (not isLastFrameUnsat()) { getSMTSolver().restoreOK(); }
         return true;
     }
     return false;
 }
 
-PartitionManager & MainSolver::getPartitionManager() { return pmanager; }
```
Tomaqa, line 106: This is what clang-format did by itself? :o

src/api/MainSolver.h
```diff
@@ -35,19 +34,21 @@ class sstat {
...
+const sstat s_True = toSstat(1);
 const sstat s_False = toSstat(-1);
-const sstat s_Undef = toSstat( 0);
-const sstat s_Error = toSstat( 2);
-
+const sstat s_Undef = toSstat(0);
+const sstat s_Error = toSstat(2);
 
 class MainSolver {
 protected: /** Helper classes to deal with assertion stack, preprocessing and substitutions **/
```
Tomaqa, line 54: I would recommend to follow the order public, protected, private in general.
In most cases, when we look into header files, we are interested in the public API.
But this way, we have to scroll down and search for public parts.

In most cases, it is not a problem when a member that has not been defined yet is accessed.
It is not the case of member types and classes, but still public API should not depend on protected/private types.

src/api/MainSolver.h
```diff
@@ -35,19 +34,21 @@ class sstat {
...
+        else
+            assert(false);
     }
-    char getValue()            const { return value; }
+    char getValue() const { return value; }
     friend sstat toSstat(int v);
 };
 
-inline sstat toSstat(int v) {return sstat(v); }
+inline sstat toSstat(int v) {
```
Tomaqa, line 44: I do not understand why in this case it is not one-line?

src/api/MainSolver.cc
```diff
@@ -492,12 +445,10 @@ MainSolver::SubstitutionResult MainSolver::computeSubstitutions(PTRef fla) {
...
         // l_True : exists and is valid
         // l_False : exists but has been disabled to break symmetries
-        MapWithKeys<PTRef,lbool,PTRefHash> new_units;
+        MapWithKeys<PTRef, lbool, PTRefHash> new_units;
         vec<PtAsgn> current_units_vec;
         bool rval = logic.getNewFacts(simp_formula, new_units);
-        if (not rval) {
-            return SubstitutionResult{{}, logic.getTerm_false()};
-        }
+        if (not rval) { return SubstitutionResult{{}, logic.getTerm_false()}; }
```
Tomaqa, line 451: I think the former format was better

src/api/MainSolver.cc
```diff
@@ -303,27 +274,21 @@ sstat MainSolver::giveToSolver(PTRef root, FrameId push_id) {
...
-        printFramesAsQuery();
+    if (config.dump_query()) printFramesAsQuery();
 
     if (rval == s_Undef) {
         try {
             rval = solve();
-        } catch (std::overflow_error const& error) {
-            rval = s_Error;
-        }
+        } catch (std::overflow_error const & error) { rval = s_Error; }
```
Tomaqa, line 291: I think the former format was better, I also think this would be better:
```
try {
...
}
catch (...) {
...
}
```

src/api/MainSolver.cc
```diff
@@ -95,34 +82,29 @@ bool MainSolver::pop()
...
     if (getConfig().produce_inter()) {
-        // MB: Important for HiFrog! partition index is the index of the formula in an virtual array of inserted formulas,
-        //     thus we need the old value of count. TODO: Find a good interface for this so it cannot be broken this easily
+        // MB: Important for HiFrog! partition index is the index of the formula in an virtual array of inserted
+        // formulas, thus we need the old value of count.
+        // TODO: Find a good interface for this so it cannot be broken this easily
         unsigned int partition_index = insertedFormulasCount++;
         pmanager.assignTopLevelPartitionIndex(partition_index, fla);
         assert(pmanager.getPartitionIndex(fla) != -1);
+    } else {
```
Tomaqa, line 105: I would prefer
```
if (...) {
...
}
else {
...
}
```
which is more line-oriented

src/api/MainSolver.cc
```diff
@@ -95,34 +82,29 @@ bool MainSolver::pop()
...
     return false;
 }
 
-PartitionManager & MainSolver::getPartitionManager() { return pmanager; }
-
-sstat MainSolver::getStatus() const { return status; }
-
-void MainSolver::insertFormula(PTRef fla)
-{
+void MainSolver::insertFormula(PTRef fla) {
```
Tomaqa, line 91: I think that format
```
Type foo(...)
{
...
}
```
is better, because when the function should catch an exception, then it looks like
```
Type foo(...)
try {
...
}
catch ...
```
In the former format, the `try` keyword may be difficult to spot, e.g. `Type foo(maaany params...) try {`.

src/api/MainSolver.cc
```diff
@@ -95,34 +82,29 @@ bool MainSolver::pop()
         frames.pop();
         substitutions.pop();
         firstNotSimplifiedFrame = std::min(firstNotSimplifiedFrame, frames.frameCount());
-        if (not isLastFrameUnsat()) {
-            getSMTSolver().restoreOK();
-        }
+        if (not isLastFrameUnsat()) { getSMTSolver().restoreOK(); }
```
Tomaqa, line 85: The former was better ...

Tomaqa (commented):
I think that all my comments should be easily achieved by modifying clang-format config file.

src/api/MainSolver.cc
blishko: There was a bit of manual intervention here :)
Because it is out-of-line, clang-format did not want to have on one line. Then I realized this is an artifact from my previous changes and that it can safely go back to the header.

src/api/MainSolver.cc
blishko: This is the style we eventually agreed on. For one-line if-blocks, we put them on the same line.

Tomaqa: Since it is a block, it does not make too much sense to me.
In the case of one-liners, only one command should be used, so I would keep it this way but remove the block `{}`.
In the rest cases, I would use the multiline variant.

src/api/MainSolver.cc
Tomaqa: I would remove `{}`

Tomaqa (approved):
LGTM!

PR #679, opened 2024-03-05
Get rid of readline and GPL license

### LINKED ISSUE #675: Get rid of `readline` and GPL license
.. unless there is a serious reason why `readline` should be preferred over `libedit` (or how is is called).
I also think that `CMAKE_FLAGS=-DENABLE_LINE_EDITING:BOOL=ON` should be enabled by default, that is, to link with a line editing tool by default.
I do not see any drawback to doing so.

Tomaqa:
Resolves #675.

However, GNU license is still used in SMTLIB2 parser (`src/parsers/smt2new`) and also in `src/proof`.


blishko (approved):
LGTM!

PR #680, opened 2024-03-05
move the built exec. into bin directory

### LINKED ISSUE #671: consolidate `bin` directories and executables
- get rid of `./bin/` where confusingly a source code is present
- either place the executable `./build/opensmt` into `./build/bin` or get rid of currently empty `./build/bin`

Tomaqa:
Resolves #671 

In case someone prefers having the executable also in `build` instead of `build/bin`, a symlink can be added. But I have not figured out how to do it easily in CMake..


Tomaqa:
It seems that it would require too much effort, so I am abandoning this attempt.

Hence it would be good to remove `build/bin` directory (resp. not creating it), but I still have not figured out how to.

PR #681, opened 2024-03-05
Remove outdated mentions of GPL

blishko:
Periplo was originally a sibling project to OpenSMT and was eventually incorporated into it, see (https://verify.inf.usi.ch/periplo)
As OpenSMT has been using MIT license for quite some time, we update also proof-related files that originated in Periplo to MIT license.

Moreover, we remove several unused files, mostly related to parsers, that also mention GPL license.


Tomaqa (approved):
LGTM! Have not found any further relevant mentions of GNU|GPL.

PR #682, opened 2024-03-05
Minor version bump

blishko:
The accumulated changes since last release demand a bump in minor version instead of only the patch version for the next release.


Tomaqa:
The PR is not into master :]

Tomaqa:
So you do not need a review, but I agree. Not sure though if merging to the branch will again require review of that branch.

blishko:
With that, I express the dependency, so that this should be merged later than the other PR. When the other is merged, this is automatically rebased to target master.
And it does require a review :)

Tomaqa (approved):
OK it changed to master automatically :] here you go

PR #685, opened 2024-03-14
Consolidate tracking of partitions for proofs and interpolation

Tomaqa:
Replaces almost all `produce_inter` conditions by `produce_proof`, and applies rule `produce_inter` => `produce_proof`.
It reverts a commit from 2020 that, on the contrary, replaced `produce_proof`s with `produce_inter`.

Note that there are still some places where `produce_inter` remains but I am not sure if it is right or should be replaced too.

### DIFF
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -236,7 +236,7 @@ lbool MainSolver::getTermValue(PTRef tr) const {
 std::unique_ptr<InterpolationContext> MainSolver::getInterpolationContext() {
     if (status != s_False) { throw OsmtApiException("Interpolation context cannot be created if solver is not in UNSAT state"); }
     return std::make_unique<InterpolationContext>(
-            config, *theory, *term_mapper, getSMTSolver().getProof(), pmanager
+            config, *theory, *term_mapper, getSMTSolver().getResolutionProof(), pmanager
     );
 }
 
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -131,7 +131,7 @@ void MainSolver::insertFormula(PTRef fla)
 sstat MainSolver::simplifyFormulas() {
     status = s_Undef;
     for (std::size_t i = firstNotSimplifiedFrame; i < frames.frameCount() && status != s_False; i++) {
-        PreprocessingContext context {.frameCount = i, .perPartition = getConfig().produce_inter() };
+        PreprocessingContext context {.frameCount = i, .perPartition = getConfig().produceProof() };
         firstNotSimplifiedFrame = i + 1;
         if (context.perPartition) {
             vec<PTRef> frameFormulas;
@@ -280,7 +280,7 @@ sstat MainSolver::giveToSolver(PTRef root, FrameId push_id) {
     ClauseCallBack callBack;
     ts.setClauseCallBack(&callBack);
     ts.Cnfizer::cnfize(root, push_id);
-    bool keepPartitionsSeparate = getConfig().produce_inter();
+    bool keepPartitionsSeparate = getConfig().produceProof();
     Lit frameLit = push_id == 0 ? Lit{} : term_mapper->getOrCreateLit(frameTerms[push_id]);
     int partitionIndex = keepPartitionsSeparate ? pmanager.getPartitionIndex(root) : -1;
     for (auto & clause : callBack.clauses) {
@@ -484,7 +484,7 @@ PTRef MainSolver::substitutionPass(PTRef fla, PreprocessingContext const& contex
 
 MainSolver::SubstitutionResult MainSolver::computeSubstitutions(PTRef fla) {
     SubstitutionResult result;
-    assert(getConfig().do_substitutions() and not getConfig().produce_inter());
+    assert(getConfig().do_substitutions() and not getConfig().produceProof());
     PTRef root = fla;
     Logic::SubstMap allsubsts;
     while (true) {
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -82,7 +82,7 @@ void MainSolver::push()
 bool MainSolver::pop()
 {
     if (frames.frameCount() > 1) {
-        if (config.produce_inter() > 0) {
+        if (trackPartitions()) {
             auto const & partitionsToInvalidate = frames.last().formulas;
             ipartitions_t mask = 0;
             for (PTRef partition : partitionsToInvalidate) {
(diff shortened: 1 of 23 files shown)


Tomaqa:
Also, does `PRODUCE_PROOF` macro still mean anything, or should it be removed?

Tomaqa:
I noticed that `PRODUCE_PROOF` is mentioned in a previous PR https://github.com/usi-verification-and-security/opensmt/pull/118, from where I assume that it should indeed be completely removed.

src/api/MainSolver.cc
```diff
@@ -280,7 +283,7 @@ sstat MainSolver::giveToSolver(PTRef root, FrameId push_id) {
     ClauseCallBack callBack;
     ts.setClauseCallBack(&callBack);
     ts.Cnfizer::cnfize(root, push_id);
-    bool keepPartitionsSeparate = getConfig().produce_inter();
+    bool keepPartitionsSeparate = getConfig().produce_proof();
```
blishko, line 286:
```suggestion
    bool trackPartitions = getConfig().produce_proof();
```

I think it would be better to have a method `bool trackPartitions() const`, same as in `Theory`. (maybe one can just use the other?)

It can either check `SMTConfig::produce_proof()` or `SMTConfig::produce_interpolants()` (and later also `SMTConfig::produce_unsat_core()`).
I think the first option is slightly better, because then the proof is a bit closer to the original formula than in the second case.

src/api/MainSolver.cc
```diff
@@ -115,6 +117,7 @@ void MainSolver::insertFormula(PTRef fla)
     // TODO: Move this to preprocessing of the formulas
     fla = IteHandler(logic, getPartitionManager().getNofPartitions()).rewrite(fla);
 
+    //? TK: is this really related only to interpolants, or to proofs in general?
     if (getConfig().produce_inter()) {
```
blishko, line 121: I think this should call the method `MainSolver::trackPartitions` that I suggest to create below.

src/api/MainSolver.cc
```diff
@@ -82,7 +82,9 @@ void MainSolver::push()
 bool MainSolver::pop()
 {
     if (frames.frameCount() > 1) {
-        if (config.produce_inter() > 0) {
+        //? TK: is this really related only to interpolants, or to proofs in general?
+        //? e.g. in Theory, there is: `bool keepPartitions() const { return config.produce_proof(); }`
+        if (config.produce_inter()) {
```
blishko, line 87: This should use the new method `MainSolver::trackPartitions` that I suggest to create below.

blishko (commented):
Here are my suggestions to clear the semantics a bit.

src/logics/Theory.h
```diff
@@ -49,7 +49,7 @@ class Theory
 
     Theory(SMTConfig &c) : config(c) { }
 
-    inline bool keepPartitions() const { return config.produce_inter(); }
+    inline bool keepPartitions() const { return config.produce_proof(); }
```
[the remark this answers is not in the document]
blishko: Actually, it seems after the recent refactoring, this method is no longer used and should be removed.
The decision to track partitions should be made in the `MainSolver`.

src/smtsolvers/CoreSMTSolver.h
```diff
@@ -486,7 +486,7 @@ class CoreSMTSolver
     //
     // Proof production
     //
-    bool logsProofForInterpolation() const { return static_cast<bool>(proof); }
+    bool logsProof() const { return static_cast<bool>(proof); }
```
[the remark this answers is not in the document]
Tomaqa: Shouldn't be the whole object `proof` and class `Proof` also called `resolutionProof` and `ResolutionProof`, resp.?

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -725,7 +725,7 @@ void CoreSMTSolver::analyze(CRef confl, vec<Lit>& out_learnt, int& out_btlevel)
         pathC--;
         // Add resolution step only if this is not the last literal from current level
         // The last literal is not resolved! It is a part of the learnt clause
-        if (logsProofForInterpolation && pathC > 0)
+        if ( logProof && pathC > 0 )
```
blishko, line 728:
```suggestion
        if (logProof && pathC > 0)
```

src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -828,7 +828,7 @@ void CoreSMTSolver::analyze(CRef confl, vec<Lit>& out_learnt, int& out_btlevel)
 bool CoreSMTSolver::litRedundant(Lit p, uint32_t abstract_levels)
 {
     // MB: TODO: figure out if this is compatible with proof tracking
-    if (logsProofForInterpolation() || config.sat_minimize_conflicts <= 0 )
+    if ( logsResolutionProof() || config.sat_minimize_conflicts <= 0 )
```
blishko, line 831:
```suggestion
    if (logsResolutionProof() || config.sat_minimize_conflicts <= 0)
```

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -807,6 +807,7 @@ void LASolver::printStatistics(std::ostream & out) {
     laSolverStats.printStatistics(out);
 }
 
+//? TK: I do not understand this, is it related directly to inter. or to resolution proofs?
```
blishko, line 810: Not directly connected to resolution proofs.
This condition should stay as is, I believe.

src/smtsolvers/CoreSMTSolver.h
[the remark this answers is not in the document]
blishko: Yes, it might make sense to rename, as you suggest.

src/smtsolvers/CoreSMTSolver.h
```diff
@@ -484,14 +485,13 @@ class CoreSMTSolver
...
 
     //
-    // Proof production
+    // ResolutionProof production
     //
-    bool logsProofForInterpolation() const { return static_cast<bool>(proof); }
-    void finalizeProof(CRef finalConflict);
-    std::unique_ptr<Proof> proof;                 // (Pointer to) Proof store
+    void finalizeResolutionProof(CRef finalConflict);
+    std::unique_ptr<ResolutionProof> resolution_proof;                 // (Pointer to) ResolutionProof store
```
blishko, line 491: I think we should use `resolutionProof` to be consistent with the style.

blishko:
Looks good!
I think we should use consistent naming style and it would be good to squash some of the commits (at lease the current last one with the one that introduced the comment).

src/smtsolvers/CoreSMTSolver.h
Tomaqa: I think that all member variables use snake case, only functions use camelCase, right?

Tomaqa: Now I am confused, e.g. in `MainSolver.h` both cases of naming member variables appear, even after applying the Clang formatting in #677

blishko: All the new code has been using camelCase for everything for some years now.

blishko (approved):
LGTM!

PR #686, opened 2024-03-14
Implementation of SMT-LIB2 `get-proof`

Tomaqa:


Tomaqa:
I screwed up the branches, this should depend on #685 in the first place ...

blishko:
We still want to merge #685 first and this one only after that, right?

src/api/MainSolver.cc
```diff
@@ -223,6 +223,13 @@ std::unique_ptr<Model> MainSolver::getModel() {
     return modelBuilder.build();
 }
 
+const ResolutionProof& MainSolver::getResolutionProof() const {
+    //? check smt_solver->logsResolutionProof() ?
```
blishko, line 227: It would be good to resolve this comment.
There should be a way to check if we really have a proof.
Currently, `CoreSMTSolver::getResolutionProof` only uses `assert`, which is too weak.
I think checking `smt_solver->logsResolutionProof` should be sufficient. It can be conjoined to the `status != s_False` check.

blishko (commented):
A few style fixes.

src/smtsolvers/ResolutionProof.cc
```diff
@@ -158,7 +158,7 @@ void ResolutionProof::printSMT2(std::ostream & out, CoreSMTSolver & s, THandler
       continue;
     }
     assert( clause_to_proof_der.find( cr ) != clause_to_proof_der.end( ) );
-    ResolutionProofDer * d = &clause_to_proof_der.at(cr);
+    const ResolutionProofDer * d = &clause_to_proof_der.at(cr);
```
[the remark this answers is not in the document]
Tomaqa: OK! Is clang-format capable of enforcing this?

src/smtsolvers/ResolutionProof.cc
```diff
@@ -201,8 +201,8 @@ void ResolutionProof::printSMT2(std::ostream & out, CoreSMTSolver & s, THandler
       out << "(let (cls_" << cr;
       nof_lets ++;
 
-      std::vector< CRef > & chain_cla = d->chain_cla;
-      std::vector< Var > & chain_var = d->chain_var;
+      auto& chain_cla = d->chain_cla;
+      auto& chain_var = d->chain_var;
```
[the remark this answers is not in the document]
Tomaqa: OK, after some research, I agree. Again: is clang-format capable of enforcing this?

src/api/MainSolver.cc
Tomaqa: The reason why I did not implement it this way right away is that `getModel` also does not do such checks. To make it I consistent, I think an update where similar checks are added should follow. Do you agree?
And maybe it would be even better if both of these changes (i.e. of `getModel` and also `getResProof`) would be done in a single PR, that is, not here. And maybe also in the case of `getInterpolationContext`.

src/smtsolvers/ResolutionProof.cc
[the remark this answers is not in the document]
blishko: The space can be enforces by clang-format.

src/smtsolvers/ResolutionProof.cc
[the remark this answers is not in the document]
blishko: With east vs west const, I don't think clang-format can enforce this rule. But I may be wrong.

src/api/MainSolver.cc
blishko: `getModel` and `getInterpolationContext` are different, because they return unique pointers. They create these objects on demand. The situation with proof is different because we are returning a reference. So we need to make sure that we have the object.

src/api/MainSolver.h
```diff
@@ -236,6 +238,9 @@ class MainSolver {
     // Returns model of the last query (must be in satisfiable state)
     std::unique_ptr<Model> getModel();
 
+    // Returns proof of the last query (must be in unsatisfiable state)
+    ResolutionProof const & getResolutionProof() const;
```
blishko, line 242: I am actually starting to question if this should be exposed.

What if  `MainSolver` would only have a method to print the proof, but would never actually expose `ResolutionProof` to the outside world?

If some program includes OpenSMT as a library it will see this method on `MainSolver` object, but it will not be able to do anything with the reference, because `ResolutionProof` will be incomplete type at that point, no?

src/api/MainSolver.cc
blishko: Ah, but I see that we should check that OpenSMT has been configured to produce models and proofs, respectively.

src/smtsolvers/ResolutionProof.cc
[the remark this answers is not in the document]
Tomaqa: Maybe it works in clang14: https://stackoverflow.com/a/71695823/9669624
But it seems to be for qualifiers in general which I am not sure if fits our needs.

src/api/MainSolver.h
Tomaqa: A lot of classes in OpenSMT are designed as separate classes, not as nested classes. Hence it is problematic to "hide" them from the users. I honestly do not like too much the design where it is necessary to forward-declare classes and to define them later, I think that the header should be always available. In the case of inappropriate access to a "hidden" class, I think it is much cleaner and clearer to let the compiler complain that the class is protected/private than to have undefined references, where it is very unclear why it happened.

Anyway, I do not understand the class `ResolutionProof` that much so I cannot say if there are use cases where the user indeed wants the class itself. If not, I agree to remove it. However, this would not remove the problem, `ResolutionProof` is still available from `CoreSMTSolver` which is accessible from `MainSolver`.

src/smtsolvers/ResolutionProof.cc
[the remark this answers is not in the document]
blishko: Yes, I believe that is what we want.

src/api/MainSolver.h
blishko: `CoreSMTSolver` should also be hidden from the users.
My mental model is that `MainSolver` should work as the API of OpenSMT, exposing all the functionality without exposing the implementation. I see `CoreSMTSolver` and `ResolutionProof` as the implementation details.
Maybe there is a case for exposing some version of a proof to the user, but at the moment I don't believe our `ResolutionProof` is the abstraction that should be exposed.

blishko (approved):
LGTM!

Issue #688, opened 2024-03-25
Fix `MainSolver` API


Tomaqa:
The public API should not expose private classes and functions.
Also add appropriate checks for the existing API calls.

Tomaqa:
The issue is not fully resolved yet. For example, the SMT solver and `THandler` should not be necessary to be included in the public API. Not speaking of the initialization of `MainSolver` where for example it first constructs `thandler` and then constructs `logic(thandler->getLogic())`

PR #689, opened 2024-03-25
Fix some access levels in `MainSolver` and add some checks to public …

Tomaqa:
…member functions.

Related to #688 (does not fully resolve)

### DIFF
--- a/test/unit/test_UFInterpolation.cc
+++ b/test/unit/test_UFInterpolation.cc
@@ -7,10 +7,6 @@
 #include <MainSolver.h>
 #include "VerificationUtils.h"
 
-bool verifyInterpolant(PTRef itp, PartitionManager & pManager, ipartitions_t const & Amask, Logic & logic) {
-    return VerificationUtils(logic).verifyInterpolantInternal(pManager.getPartition(Amask, PartitionManager::part::A), pManager.getPartition(Amask, PartitionManager::part::B), itp);
-}
-
 class UFInterpolationTest : public ::testing::Test {
 protected:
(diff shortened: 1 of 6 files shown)


Tomaqa:
Making `getSMTSolver` protected and also `THandler` is not possible at the moment, some classes such as `BitBlaster` are using it. I am afraid it would require more thorough refactoring.

test/unit/test_UFInterpolation.cc
```diff
@@ -11,6 +11,12 @@ bool verifyInterpolant(PTRef itp, PartitionManager & pManager, ipartitions_t con
     return VerificationUtils(logic).verifyInterpolantInternal(pManager.getPartition(Amask, PartitionManager::part::A), pManager.getPartition(Amask, PartitionManager::part::B), itp);
 }
 
+struct TestMainSolver : MainSolver {
```
blishko, line 14: I don't particularly like the approach of creating subclasses solely for the purpose being able to access some protected member.

I think a better solution here would be to do what `LRAInterpolationTest` is doing, i.e., explicitly track the partitions in each test.
That way the public API is sufficient.

blishko (commented):
The changes in MainSolver are good, but I suggest to deal with `UFInterpolationTest` differently and make it consistent with how `LRAInterpolationTest` works.

blishko (approved):
LGTM!

blishko:
Please, next time, make sure that the title of git commit is limited to 72 characters, so that GitHub does not split it when creating a PR.
Having the sentence split between title and description like in this PR is weird.

PR #695, opened 2024-03-30
Improve Handling of Empty Name in Logic::mkConst

### LINKED ISSUE #694: Assertion failure at src/logics/Logic.cc:653
Hi, 
For this following instance, opensmt https://github.com/usi-verification-and-security/opensmt/commit/84f1c2b03d4864fe8e59b0aaf86baa8c843c4c5f

```
$ cat test.smt2 
(set-logic ALL)
(declare-fun x () Int)
(assert (= ""))
(check-sat)
$ opensmt test.smt2 
; this binary is compiled in debug mode (slow)
opensmt: /home/opensmt/src/logics/Logic.cc:653: virtual PTRef Logic::mkConst(const char*): Assertion `strlen(name) > 0' failed.
Aborted (core dumped)
```

I have prepared a fix that addresses this issue. I would appreciate it if you could review the PR at your convenience.

merlinsun:
Fixes #694.

When opensmt processes assertions involving empty name, it triggers an assertion failure in `Logic::mkConst`.

To resolve this issue, I propose a modification to the `Logic::mkConst` to check for empty name explicitly. If an empty name is encountered, the function will now print an error message and return `PTRef_Undef`.

### DIFF
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -648,12 +648,6 @@ PTRef Logic::mkNot(PTRef arg) {
 
 PTRef Logic::mkConst(const char* name)
 {
-    //assert(0);
-    // return PTRef_Undef;
-    if (strlen(name) == 0) {
-        std::cerr << "Error in Logic::mkConst: empty name" << std::endl;
-        return PTRef_Undef;
-    }
     return resolveTerm(name, {});
 }
 
--- a/src/logics/Logic.cc
+++ b/src/logics/Logic.cc
@@ -649,8 +649,11 @@ PTRef Logic::mkNot(PTRef arg) {
 PTRef Logic::mkConst(const char* name)
 {
     //assert(0);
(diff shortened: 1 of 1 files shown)


src/logics/Logic.cc
```diff
@@ -649,8 +649,11 @@ PTRef Logic::mkNot(PTRef arg) {
 PTRef Logic::mkConst(const char* name)
 {
     //assert(0);
-    //return PTRef_Undef;
-    assert(strlen(name) > 0);
+    // return PTRef_Undef;
+    if (strlen(name) == 0) {
```
blishko, line 653: I don't think this is the right solution.

Looking at the code, I believe we can just remove this assert (and the commented-out code above).
`resolveTerm` will report an error in this situation.

In any case, OpenSMT does not support strings, so this input is not a valid input for OpenSMT anyway.

blishko (approved):
LGMT!

PR #696, opened 2024-03-31
ArithLogic: Fix handling of minus operator with more than two arguments

### LINKED ISSUE #690: Assertion failure at opensmt/src/logics/ArithLogic.cc:563
Hi,
For this following instance, opensmt https://github.com/usi-verification-and-security/opensmt/commit/84f1c2b03d4864fe8e59b0aaf86baa8c843c4c5f 
```
$ opensmt small.smt2    
; this binary is compiled in debug mode (slow)
opensmt: /home/opensmt/src/logics/ArithLogic.cc:563: PTRef ArithLogic::mkMinus(vec<PTRef>&&): Assertion `args.size() == 2' failed.
Aborted (core dumped)
$ cat small.smt2 
(set-logic ALL)
(assert (= (- 0 0 0)))
```

blishko:
According to the SMT-LIB definition of theory of Ints, operator minus is left associative. This means that it is allowed to create a function application of minus to more than two arguments. This is then a syntactic sugar that applies the operator recursively first to all the arguments except the last one and then apply the operator to the result and the last argument.

As @aehyvari pointed out, we can treat expression `(- t1 t2 t3 ... tn)` as `(+ t1 (- t2) (- t3) ... (- tn))`.

Fixes #690.


src/logics/ArithLogic.cc
```diff
@@ -560,8 +560,11 @@ PTRef ArithLogic::mkMinus(vec<PTRef> && args)
     if (args.size() == 1) {
         return mkNeg(args[0]);
     }
-    assert(args.size() == 2);
-    if (args.size() > 2) { throw OsmtApiException("Too many terms provided to LALogic::mkNumMinus"); }
+    if (args.size() > 2) { // Minus is left associative according to SMT-LIB
+        PTRef last = args.last();
+        args.pop();
+        return mkMinus(mkMinus(std::move(args)), last);
```
aehyvari, line 566: We could do something like
```
for (int i = 1; i < args.size(); i++)
  args[i] = mkNeg(args[i]);
```
and pass `args` directly to `mkPlus`.  It doesn't look like it'd take much more space and we can get rid of special handling?  I guess there's a more elegant way than the `for` loop with an index, but anyway.

blishko: Ah, yes, you are right! For some reason I thought that we cannot flatten such expressions, but, yes, we can!

aehyvari (approved):
LGTM!

PR #700, opened 2024-04-24
CI: Update OSX resource

blishko:
CircleCI is shutting down its Intel-based OSX images and moving to M1-based images.
See: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,7 @@ jobs:
     build-macos:
         macos:
             xcode: 14.2.0
-        resource_class: macos.x86.medium.gen2
+        resource_class: macos.m1.medium.gen1
         steps:
             - run:
                 name: Packages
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ jobs:
             - checkout
             - run:
                 name: Store path
-                command: echo 'export PATH=/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH' >> $BASH_ENV
+                command: echo 'export PATH=/opt/homebrew/opt/flex/bin:/opt/homebrew/opt/bison/bin:$PATH' >> $BASH_ENV
```
Tomaqa, line 150: What is the reason for this?

blishko: This is the path where homebrew places binaries on this new image.
I don't know why it changed, but it did.

Tomaqa (approved):
OK

PR #703, opened 2024-05-01
Ignore files related to ccls language server

Tomaqa:
I use for example in VS Codium. It is an alternative to for example clang server.
It is possible that someone do not want to use GCC... We can either ignore this, or add the conf. file and at the same time add it to .gitignore so anyone can freely update the file on local machine.


blishko (approved):
LGMT!

PR #707, opened 2024-05-03
Add support for unsat core computation for theory combination

blishko:
This PR enables computation of unsat cores for theory combination by supporting computation of resolution proof for theory combination.
This builds on top of changes in #706, which enables per-partition preprocessing.
Besides preprocessing changes, the only missing feature seemed to be handling unit clauses at level 0 computed from splits (theory combination lemmas for interface variables), which is added in this PR (see the corresponding commit message for details).

### DIFF
--- a/src/smtsolvers/CoreSMTSolver.cc
+++ b/src/smtsolvers/CoreSMTSolver.cc
@@ -934,6 +934,27 @@ void CoreSMTSolver::finalizeResolutionProof(CRef finalConflict) {
     resolutionProof->endChain(CRef_Undef);
 }
 
+
+CRef CoreSMTSolver::logUnitClauseDerivationAtLevelZero(CRef cref) {
+    assert(logsResolutionProof() and decisionLevel() == 0);
+    resolutionProof->beginChain(cref);
+    Clause const & c = ca[cref];
+    assert(c.size() >= 2);
+    assert(value(c[0]) == l_Undef);
+    for (unsigned k = 1; k < c.size(); ++k) {
+        Lit l = c[k];
+        assert(value(l) == l_False);
+        Var v = var(l);
+        assert(level(v) == 0);
+        assert(reason(v) != CRef_Fake);
+        assert(reason(v) != CRef_Undef);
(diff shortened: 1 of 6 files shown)


src/smtsolvers/CoreSMTSolver.cc
```diff
@@ -934,6 +934,26 @@ void CoreSMTSolver::finalizeResolutionProof(CRef finalConflict) {
...
+    resolutionProof->beginChain(cref);
+    Clause const & c = ca[cref];
+    assert(c.size() >= 2);
+    assert(value(c[0]) == l_Undef);
+    for (unsigned k = 1; k < c.size(); ++k) {
+        Lit l = c[k];
+        assert(value(l) == l_False);
+        Var v = var(l);
+        assert(level(v) == 0);
+        assert(reason(v) != CRef_Fake and reason(v) != CRef_Undef);
```
Tomaqa, line 949: In case of assertion failure, I believe it is better to split the checks to know exactly what happened.

test/unit/test_UnsatCore.cc
```diff
@@ -82,3 +83,133 @@ TEST_F(UFUnsatCoreTest, Bool_ReuseProofChain) {
...
+        nb1 = logic.mkNot(b1);
+        nb2 = logic.mkNot(b2);
+        nb3 = logic.mkNot(b3);
+        nb4 = logic.mkNot(b4);
+        nb5 = logic.mkNot(b5);
+    }
+    ArithLogic logic;
+    SMTConfig config;
+    SRef intIntArraySort;
+    PTRef v1, v2;
```
Tomaqa, line 186: `arr1, arr2` ?

src/smtsolvers/CoreSMTSolver.cc
blishko: Changed!

test/unit/test_UnsatCore.cc
blishko: Changed!

Tomaqa (approved):
LGTM!

PR #711, opened 2024-05-15
Preprocessing: Fix incrementality in Preprocessor

### LINKED ISSUE #710: Fix bugs in `QF_ALIA`
There are wrong answers to several incremental queries in benchmark `incremental/QF_ALIA/UltimateBuchiAutomizer/LexIndexValue-Pointer_true-termination.c.smt2` (sat instead of unsat).

blishko:
Previous implementation of handling incrementality in Preprocessor was not sound. The problem was the difference between eager push/pop and lazy application of preprocessing in MainSolver.
The lazy approach means that formulas are sent to the preprocessor only after a call to check-sat, i.e., possibly after several push commands. This meant that Preprocessor actually stored information incorrectly, because it considered all formulas processed during a check-sat command to be part of the last frame, when actually they could come from lower frames.

The proposed solution is to make tracking frames in Preprocessor lazy as well. This ensures that the frame information is stored correctly and formulas are not popped in Preprocessor when they are actually still valid in the MainSolver.

Fixes #710.


src/api/MainSolver.cc
```diff
@@ -559,12 +560,32 @@ void MainSolver::Preprocessor::initialize() {
     substitutions.push();
 }
 
+void MainSolver::Preprocessor::prepareForProcessingFrame(std::size_t frameIndex) {
+    assert(frameIndex < virtualFrameCount);
+    while (actualFrameCount <= frameIndex) {
+        pushInternal();
+    }
```
Tomaqa, line 567: `assert(actualFrameCount == virtualFrameCount);` ?

src/api/MainSolver.h
```diff
@@ -172,8 +172,14 @@ class MainSolver {
         [[nodiscard]] opensmt::span<const PTRef> getPreprocessedFormulas() const;
         [[nodiscard]] Logic::SubstMap getCurrentSubstitutions() const { return substitutions.current(); }
         void setSubstitutions(std::size_t level, Logic::SubstMap && subs) { substitutions.set(level, std::move(subs)); }
+        void prepareForProcessingFrame(std::size_t frameIndex);
 
     private:
+        std::size_t virtualFrameCount {1};
+        std::size_t actualFrameCount {1};
```
Tomaqa, line 179: `internalFrameCount` ?

src/api/MainSolver.h
```diff
@@ -172,8 +172,14 @@ class MainSolver {
         [[nodiscard]] opensmt::span<const PTRef> getPreprocessedFormulas() const;
         [[nodiscard]] Logic::SubstMap getCurrentSubstitutions() const { return substitutions.current(); }
         void setSubstitutions(std::size_t level, Logic::SubstMap && subs) { substitutions.set(level, std::move(subs)); }
+        void prepareForProcessingFrame(std::size_t frameIndex);
 
     private:
+        std::size_t virtualFrameCount {1};
```
Tomaqa, line 178: `solverFrameCount` ?

src/api/MainSolver.h
```diff
@@ -172,8 +172,14 @@ class MainSolver {
...
+        std::size_t actualFrameCount {1};
+        void pushInternal();
+        void popInternal();
```
Tomaqa, line 181: ordering within a class: 1. subclasses and types, 2. member functions, 3. member variables ?

src/api/MainSolver.cc
```diff
@@ -559,12 +560,32 @@ void MainSolver::Preprocessor::initialize() {
...
+}
 void MainSolver::Preprocessor::push() {
+    ++virtualFrameCount;
+}
+
+void MainSolver::Preprocessor::pop() {
```
Tomaqa, line 573: `assert(virtualFrameCount >= actualFrameCount);` ?

src/api/MainSolver.cc
```diff
@@ -559,12 +560,32 @@ void MainSolver::Preprocessor::initialize() {
     substitutions.push();
 }
 
+void MainSolver::Preprocessor::prepareForProcessingFrame(std::size_t frameIndex) {
+    assert(frameIndex < virtualFrameCount);
+    while (actualFrameCount <= frameIndex) {
+        pushInternal();
+    }
+}
 void MainSolver::Preprocessor::push() {
```
Tomaqa, line 569: `assert(virtualFrameCount >= actualFrameCount);` ?

src/api/MainSolver.cc
```diff
@@ -559,12 +560,32 @@ void MainSolver::Preprocessor::initialize() {
...
+    ++virtualFrameCount;
+}
+
+void MainSolver::Preprocessor::pop() {
+    --virtualFrameCount;
+    if (virtualFrameCount < actualFrameCount) {
+        assert(virtualFrameCount + 1 == actualFrameCount);
+        popInternal();
+    }
```
Tomaqa, line 578: what about:
```
if (virtualFrameCount >= actualFrameCount) return;
popInternal();
assert(virtualFrameCount == actualFrameCount);
```
I think that the final state is more important than an intermediate.

src/api/MainSolver.cc
blishko: That does not hold, `virtualFrameCount` can still be larger.

src/api/MainSolver.cc
blishko: Added!

src/api/MainSolver.cc
blishko: Added!

src/api/MainSolver.h
blishko: Renamed

src/api/MainSolver.h
blishko: Renamed

src/api/MainSolver.h
blishko: Fixed, if I understood it correctly.

src/api/MainSolver.cc
Tomaqa: Right ...

Tomaqa (approved):
LGTM!

PR #712, opened 2024-05-15
FastRationals: Fix normalization of rational numbers

blishko:


Tomaqa (approved):
LGTM! (as a workaround :) )

PR #715, opened 2024-05-16
ArraySolver: Tiny performance optimization

blishko:
This is just a tiny optimization to not repeat the same work inside a loop when we can do it once before the loop.

### DIFF
--- a/src/tsolvers/arraysolver/ArraySolver.cc
+++ b/src/tsolvers/arraysolver/ArraySolver.cc
@@ -450,10 +450,10 @@ std::vector<ArraySolver::LemmaConditions> ArraySolver::collectLemmaConditions(Lo
         // TODO: Figure out better way how to compute all candidates for lemmas
         for (auto first : selects) {
             ERef firstRoot = getRoot(first);
+            NodeRef arrayFirst = getNodeRef(getRoot(getArrayFromSelect(first)));
             for (auto secondIt = selects.begin(); *secondIt != first; ++secondIt) {
                 ERef second = *secondIt;
                 if (firstRoot == getRoot(second)) { continue; } // The selects are already the same, no lemma needed
-                NodeRef arrayFirst = getNodeRef(getRoot(getArrayFromSelect(first)));
                 NodeRef arraySecond = getNodeRef(getRoot(getArrayFromSelect(second)));
                 if (arrayFirst == arraySecond or getIndexedRepresentative(arrayFirst, index) == getIndexedRepresentative(arraySecond, index)) {
                     std::unordered_set<PTRef, PTRefHash> undecidedEqualities;


src/tsolvers/arraysolver/ArraySolver.cc
```diff
@@ -450,10 +450,10 @@ std::vector<ArraySolver::LemmaConditions> ArraySolver::collectLemmaConditions(Lo
         // TODO: Figure out better way how to compute all candidates for lemmas
         for (auto first : selects) {
             ERef firstRoot = getRoot(first);
+            NodeRef arrayFirst = getNodeRef(getRoot(getArrayFromSelect(first)));
             for (auto secondIt = selects.begin(); *secondIt != first; ++secondIt) {
                 ERef second = *secondIt;
                 if (firstRoot == getRoot(second)) { continue; } // The selects are already the same, no lemma needed
-                NodeRef arrayFirst = getNodeRef(getRoot(getArrayFromSelect(first)));
                 NodeRef arraySecond = getNodeRef(getRoot(getArrayFromSelect(second)));
                 if (arrayFirst == arraySecond or getIndexedRepresentative(arrayFirst, index) == getIndexedRepresentative(arraySecond, index)) {
```
Tomaqa, line 458: `getIndexedRepresentative(arrayFirst, index)` can also be computed right after `arrayFirst`

src/tsolvers/arraysolver/ArraySolver.cc
```diff
@@ -450,10 +450,10 @@ std::vector<ArraySolver::LemmaConditions> ArraySolver::collectLemmaConditions(Lo
...
                     std::unordered_set<PTRef, PTRefHash> undecidedEqualities;
```
Tomaqa, line 459: What happens if this hash is made to reuse its storage? I.e. move it outside the loops and just call `undecidedEqualities.clear()` here?

blishko: We cannot do that. The map is currently being moved into `LemmaConditions`.

src/tsolvers/arraysolver/ArraySolver.cc
blishko: Yes, but in this case, it is possible that in the original code, it would not be called at all :)

I mean, I don't want to spend time on this now. I think the whole method (at also its usage) needs investigation about how to improve it.

src/tsolvers/arraysolver/ArraySolver.cc
Tomaqa: Yes, but after moving and clearing it is again in a valid state. It may potentially reuse previously dynamically allocated storage, it depends on implementation.

Tomaqa (approved):
LGTM (for now)

PR #717, opened 2024-05-17
Arrays: Fix unsoundness in handling nested arrays

### LINKED ISSUE #710: Fix bugs in `QF_ALIA`

blishko:
As part of the initialization of the array solver, we need to collect all array, store and select terms.
Previously, we incorrectly assume that array terms and select terms are mutually exclusive. This is not true in case of nested arrays when a select term can be of an array sort.
For this reason we previously missed some select terms and consequently did not add corresponding lemmas to the solver.

Fixes #710.


src/tsolvers/arraysolver/ArraySolver.cc
```diff
@@ -144,7 +144,8 @@ void ArraySolver::declareAtom(PTRef tr) {
                 if (logic.isArrayStore(term)) {
                     storeTerms.insert(eref);
```
Tomaqa, line 145: I assume that `logic.isArrayStore(term) && logic.isArraySelect(term)` is impossible. Hence we can use `return` here?

src/tsolvers/arraysolver/ArraySolver.cc
```diff
@@ -144,7 +144,8 @@ void ArraySolver::declareAtom(PTRef tr) {
...
                 }
-            } else if (logic.isArraySelect(term)) {
+            }
```
Tomaqa, line 147: What about adding a comment with e.g. `// Select does not necessarily yield array sort` ?

src/tsolvers/arraysolver/ArraySolver.cc
Tomaqa: Maybe also an assert such as `logic.isArrayStore(term) => logic.isArraySort(logic.getSortRef(term))` at the beginning could be useful.

blishko: Added an assertion at the beginning and the return in the nested `if`.

Tomaqa:
In `incremental/QF_ALIA/UltimateBuchiAutomizer`, it resolves `CookSeeZuleger-2013TACAS-Fig7a-alloca_true-termination.c.i.smt2` and `java_Sequence-alloca_true-termination.c.i.smt2`,  but not `CookSeeZuleger-2013TACAS-Fig7b-alloca_true-termination.c.i.smt2` where the very last query fails.

Tomaqa:
Still does not entirely fix #710.

Tomaqa (approved):
OK!

PR #718, opened 2024-05-21
Fix unsoundness in incremental ALIA

blishko:
Incremental solving including arrays and arithmetic contained a bug that could cause the solver to respond `sat` when the formula is actually `unsat`.
The theory combination relies on creating interface lemmas in order for the individual solvers to agree if a pair of interface variables have the same value or not.
Such interface lemmas are created on demand, lazily. However, the array solver may create new equalities (and introduce new variables to the SAT solver) when it formulates its new lemmas. In case of incremental solving, these outlive their frame in assertion stack after pop. They are still present in the core. If, afterwards, we assert the corresponding disequality (which gets split into inequalities), we miss the introduction of the interface lemmas, because the variables of the equality are not considered interface variables (they appear only in UF part of the formula in first frame, and, after pop, they appear only in the arithmetic part of the formula).

The proposed fix is to introduce interface lemmas in the case of equalities introduced by the array solver more eagerly. This, however, comes with a performance cost because we eagerly add (possibly unnecessary) interface lemmas.


Tomaqa:
It seems to fix all remaining SMT-LIB benchmarks with arrays (`QF_A*`) under the timeout of 20 minutes. I also have noticed no significant performance drop.

src/tsolvers/UFLATHandler.cc
```diff
@@ -81,34 +74,28 @@ vec<PTRef> UFLATHandler::getSplitClauses() {
...
         return res;
     }
-    if (equalitiesToPropagate.size() == 0) {
-        return res;
-    }
     for (PTRef eq : equalitiesToPropagate) {
-        if (knownEqualities.find(eq) == knownEqualities.end()) {
+        if (equalitiesWithAddedInterfaceClauses.find(eq) == equalitiesWithAddedInterfaceClauses.end()) {
             addInterfaceClausesForEquality(logic, eq, res);
+            equalitiesWithAddedInterfaceClauses.insert(eq);
```
Tomaqa, line 98:
```suggestion
        if (auto [_, inserted] = equalitiesWithAddedInterfaceClauses.insert(eq); inserted) {
            addInterfaceClausesForEquality(logic, eq, res);
```

src/tsolvers/UFLATHandler.cc
```diff
@@ -81,34 +74,28 @@ vec<PTRef> UFLATHandler::getSplitClauses() {
...
         res.copyTo(arrayLemmas);
-        std::unordered_set<PTRef, PTRefHash> equalitiesWithAddedClauses;
         // HACK: If array solver needs to decide equality on interface vars, we need to add corresponding lemmas already here
         for (PTRef lemma : arrayLemmas) {
             if (logic.isOr(lemma)) {
                 for (PTRef lit : logic.getPterm(lemma)) {
                     PTRef atom = logic.isNot(lit) ? logic.getPterm(lit)[0] : lit;
-                    if (logic.isNumEq(atom) and knownEqualities.find(atom) == knownEqualities.end() and
-                        equalitiesWithAddedClauses.find(atom) == equalitiesWithAddedClauses.end()) {
+                    if (logic.isNumEq(atom) and equalitiesWithAddedInterfaceClauses.find(atom) == equalitiesWithAddedInterfaceClauses.end()) {
```
Tomaqa, line 82:
```suggestion
                    if (logic.isNumEq(atom) and equalitiesWithAddedInterfaceClauses.count(atom) == 0) {
```

src/tsolvers/UFLATHandler.cc
```diff
@@ -81,34 +74,28 @@ vec<PTRef> UFLATHandler::getSplitClauses() {
...
                         PTRef lhs = logic.getPterm(atom)[0];
                         PTRef rhs = logic.getPterm(atom)[1];
-                        if (std::find(interfaceVars.begin(), interfaceVars.end(), lhs) != interfaceVars.end() and
-                            std::find(interfaceVars.begin(), interfaceVars.end(), rhs) != interfaceVars.end()
-                        ) {
+                        if ((logic.isVar(lhs) or logic.isNumConst(lhs)) and (logic.isVar(rhs) or logic.isNumConst(rhs))) {
```
Tomaqa, line 85:
```suggestion
                        if (isInterfaceVar(lhs) and isInterfaceVar(rhs)) {
```
using lambda
```
    auto isInterfaceVar = [this](PTRef arg){
        return logic.isVar(arg) or logic.isNumConst(arg);
    };
```
(or use another name)

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1637,12 +1637,7 @@ vec<PTRef> Egraph::collectEqualitiesFor(vec<PTRef> const & vars, std::unordered_
         for (int i = 0; i < equivalentVars.size(); ++i) {
             for (int j = i + 1; j < equivalentVars.size(); ++j) {
                 PTRef eq = logic.mkEq(ERefToTerm(equivalentVars[i]), ERefToTerm(equivalentVars[j]));
-                if (not enode_store.has(eq)) {
-                    if (knownEqualities.find(eq) != knownEqualities.end()) {
-                        throw OsmtInternalException("Internal error in computing interface equalities in Egraph");
-                    }
-                    equalities.push(eq);
-                }
+                if (knownEqualities.find(eq) == knownEqualities.end()) { equalities.push(eq); }
```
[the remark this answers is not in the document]
blishko: `contains` would be best. Until then, I would prefer `find`. It expresses the intent better (and `count` uses `find` internally at least in one standard library implementation: https://stackoverflow.com/questions/14159682/unordered-map-which-one-is-faster-find-or-count).

src/tsolvers/UFLATHandler.cc
blishko: See my comment above.

src/tsolvers/egraph/EgraphSolver.cc
```diff
@@ -1637,12 +1637,7 @@ vec<PTRef> Egraph::collectEqualitiesFor(vec<PTRef> const & vars, std::unordered_
         for (int i = 0; i < equivalentVars.size(); ++i) {
```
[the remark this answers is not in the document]
blishko: Changed!

src/tsolvers/UFLATHandler.cc
blishko: Changed!

src/tsolvers/UFLATHandler.cc
blishko: Changed!

src/tsolvers/egraph/EgraphSolver.cc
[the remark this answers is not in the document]
Tomaqa: I see. OK then.

Tomaqa:
Resolves #710. I have discovered no new bugs even with 1h timeout.

Issue #619, opened 2023-05-12
Support for get-unsat-core


daniel-larraz:
It would be great to have support for `get-unsat-core`.

Tomaqa:
Hello @daniel-larraz , can you test in your project whether the current solution works for you?

daniel-larraz:
Hi @Tomaqa , I tested the latest version in the master branch and found a minor issue. Given this [input](https://github.com/usi-verification-and-security/opensmt/files/15487612/two_counters.lus.ic3qe.1.txt), OpenSMT returns an unsat core that includes a name that is not in context when the `get-unsat-core` command is executed. Specifically, OpenSMT returns `( c0 c14 c15 )`, but I was expecting `( c17 c14 c15 )` since the named term `c0` was asserted in a level of the assertion stack that was subsequently popped.

Thank you for your work!

blishko:
Hi @daniel-larraz, thanks for the quick testing!
We have not tested unsat cores in incremental setting.
I have an idea where this issue could be coming from. We will investigate!

blishko:
Hi @daniel-larraz, the issue should be fixed now.
Can you test on HEAD, or do you need an official release?

daniel-larraz:
I can test on HEAD. I'll keep you posted.

Tomaqa:
Thank you @daniel-larraz for the quick test.
We still keep the issue open, though, because the current implementation of the unsat cores is quite naive.

blishko:
Actually, let's close this one.
We can document the effort on improving this functionality elsewhere.

PR #727, opened 2024-06-06
Interpret: Respect scope of term names

### LINKED ISSUE #726: OpenSMT does not remove term names from context after pop
`OpenSMT` reports error on the following example, which is deemed valid by other solvers, such as `z3` and `cvc5`.

```
(set-logic QF_LIA)
(declare-fun o () Int)
(declare-fun i () Bool)
(push 1)
(assert (! (< o 0) :named c0))
(pop 1)
(assert (! i :named c0))
(check-sat)
```

The problem is, that `OpenSMT` does not remove term names from its context after `pop`, and when reusing the name `c0`, it reports an error, because the it remembers the first use of `c0` in its context.

blishko:
Previously, we just collected information about user-specified term names globally, but according to SMT-LIB, term names are also limited to their stack frame (unless option :global-declarations is set to true).

Here we replace the data structure that stores information about user-specified term names with one that respect the scope of the name definitions.

Fixes #726.
Should also resolve the problem reported in #619 with incorrect unsat core reported in incremental setting.


regression/incremental/unsat_core.smt2.expected.out
```diff
@@ -0,0 +1,3 @@
+sat
+unsat
+( c17 c15 )
```
Tomaqa, line 3: I would suggest to also add `unsat_core_global` variant with additional `(set-option :global-declarations true)`.
Currently, this returns `( c0 c15 )`, which is certainly wrong. It should still return `( c17 c15 )` right?

Tomaqa (commented):
There still seems to be a bug when `:global-declarations` is set to true.

blishko:
Renamed the files and added a test and an explanation for the unsat core in incremental setting with global declarations.

src/api/Interpret.h
```diff
@@ -64,6 +64,55 @@ class DefinedFunctions {
...
 };
 
+class TermNames {
+public:
+    bool has(std::string const & name) const { return nameToTerm.find(name) != nameToTerm.end(); }
+    bool has(PTRef term) const { return termToNames.find(term) != termToNames.end(); }
+
+    void insert(std::string const & name, PTRef term, bool scoped = false) {
+        assert(not has(name));
+        nameToTerm[name] = term;
```
Tomaqa, line 74:
```suggestion
        nameToTerm.emplace(name, term);
```
is slightly more efficient

src/api/Interpret.h
```diff
@@ -64,6 +64,55 @@ class DefinedFunctions {
...
+        termToNames[term].push_back(name);
```
Tomaqa, line 75:
```suggestion
        termToNames.insert({term, {name}});
```
is slightly more efficient, although it does not look very nice

src/api/Interpret.cc
```diff
@@ -631,16 +623,13 @@ bool Interpret::getAssignment() {
...
-        PTRef tr = nameToTerm[name];
-        lbool val = main_solver->getTermValue(tr);
-        ss << '(' << name << ' ' << (val == l_True ? "true" : (val == l_False ? "false" : "unknown"))
-            << ')' << (i < term_names.size() - 1 ? " " : "");
-    }
+    termNames.forEachNamedTerm([&](auto const & name, PTRef term) {
+        lbool val = main_solver->getTermValue(term);
+        ss << '(' << name << ' ' << (val == l_True ? "true" : (val == l_False ? "false" : "unknown")) << ')' << " ";
+    });
+    ss.seekp(-1, ss.cur);
```
Tomaqa, line 630:
```suggestion
    ss.seekp(-1, std::ios::cur);
```

src/api/Interpret.h
```diff
@@ -64,6 +64,55 @@ class DefinedFunctions {
...
+        if (scoped) { scopedNames.push(name); }
+    }
+
+    void pushScope() { scopedNames.pushScope(); }
+
+    void popScope() {
+        scopedNames.popScope([&](std::string const & name) {
+            auto it = nameToTerm.find(name);
+            if (it != nameToTerm.end()) {
```
Tomaqa, line 84: `if (it == nameToTerm.end()) return;` avoids adding another scope and indentation

src/common/ScopedVector.h
```diff
@@ -31,6 +31,9 @@ class ScopedVector {
 
     [[nodiscard]] T * data() { return elements.data(); }
     [[nodiscard]] T const * data() const { return elements.data(); }
+
+    [[nodiscard]] auto begin() const { return elements.begin(); }
+    [[nodiscard]] auto end() const { return elements.end(); }
```
Tomaqa, line 36: Add also non-const iterators.

src/api/Interpret.cc
```diff
@@ -631,16 +623,13 @@ bool Interpret::getAssignment() {
     }
     std::stringstream ss;
```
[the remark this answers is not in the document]
blishko: Fixed!

src/api/Interpret.cc
blishko: Fixed!

src/api/Interpret.h
blishko: Updated!

src/api/Interpret.h
blishko: This is not equivalent. Here we need the first version, because the key `term` may already exist in the map `termToNames`.

src/api/Interpret.h
blishko: Fixed!

src/common/ScopedVector.h
blishko: Added!

src/api/Interpret.h
Tomaqa: You are right!

Tomaqa (approved):
LGTM!

PR #728, opened 2024-07-05
CMake: Bump required version of CMake

blishko:
The latest version of CMake at the time of writing is 3.30 and it warns that

"Compatibility with CMake < 3.5 will be removed from a future version of
 CMake."

We need to bump the required CMake version at least to 3.5. However, we have at a few places some configuration that requires CMake >= 3.12. I think it reasonable to bump the required version to that version and simplify our scripts a bit.


blishko:
> LGTM! I would also move the C++ version to 20 - here or in a new PR?

Let's do that in a new PR!

PR #732, opened 2024-07-05
used the abstraction `Number` instead of explicitly using `FastRational`

Tomaqa:
This is necessary if `FastRational` is eventually replaced by another class with the same interface.


src/tsolvers/stpsolver/STPGraphManager_implementations.hpp
```diff
@@ -72,7 +72,7 @@ typename STPGraphManager<T>::DFSResult STPGraphManager<T>::dfsSearch(VertexRef i
     size_t total = 0;
     std::stack<VertexRef> open;
     visited[init.x] = true;
-    length[init.x] = Converter<T>::getValue(0);
+    length[init.x] = Converter<T>::getValue(0L);
```
blishko, line 75: Why do we need this to be `long`?

blishko (commented):
Seems good, just two small questions.

src/tsolvers/bvsolver/BitBlaster.cc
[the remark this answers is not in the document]
Tomaqa: Sure. I formerly did some real changes there but then reverted them, but the WS stayed.

src/tsolvers/stpsolver/STPGraphManager_implementations.hpp
Tomaqa: `Converter<T>::getValue` is overloaded for types `ptrdiff_t` and `opensmt::Number`. On my machine, `ptrdiff_t` corresponds to `long`, but it may also be `int` elsewhere.
`0` is `int`. If `Number` is `double`, then calling `getValue` is ambiguous because both cases require a similar arithmetic conversion from `int`. Using `0L` ensures to use `ptrdiff_t` variant - assuming that it is `long` and `int`.
A more robust solution would be to use `ptrdiff_t(0)` instead of `0L`.
Another possibility is to add an additional constructor `static T getValue(integral auto val) { return getValue(static_cast<ptrdiff_t>(val)); }` - assuming that all integral types are meant to use `ptrdiff_t`. Here I am not sure because I do not fully understand the intended difference between the two variants.

Tomaqa: After some search, it seems that `Converter<T>::getValue` is only used once with `Number` (in `src/tsolvers/stpsolver/STPSolver_implementations.hpp`) and several times with `ptrdiff_t`, where, however, in all cases the argument is `0`. Hence I am a bit confused why specifically `ptrdiff_t` is used, but I guess that it can be completely replaced by `std::signed_integral auto`.

Tomaqa: `SafeInt` requires `ptrdiff_t`. Hence I would leave the current constructor and add the generic one.

blishko (approved):
LGTM!

blishko:
Looks like we need to update the compiler at starexec CI job first.

Tomaqa:
I hoped that the previous PR will fix it x)

Tomaqa:
Formatter CI already passes, starexec remains -> #734

blishko (approved):
LGTM!

PR #735, opened 2024-07-06
added more rules to clang-format

Tomaqa:
- `Standard: c++20`, `AlignOperands: AlignAfterOperator` and `NamespaceIndentation: All` do not affect the current files
- `QualifierAlignment: Right` has not been enforced yet but we seem to follow this guideline
- `AlwaysBreakTemplateDeclarations: Yes` changed a few places, but I believe that this setting is much easier to read, especially with the possibility of using C++20 concepts

Reference: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,12 +1,17 @@
 ---
 Language: Cpp
-BasedOnStyle:  LLVM
+Standard: c++20
+BasedOnStyle: LLVM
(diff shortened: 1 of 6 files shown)


blishko (approved):
LGTM!

PR #736, opened 2024-07-06
added 'scripts' directory and removed old scripts

Tomaqa:
I suppose that the directory will become useful for other scripts as well. A candidate to move it there is `ci/build_maximally_static.sh` which is not necessarily CI-related, but it is used within Travis commands so I did not want to touch it.


blishko (approved):
LGTM

PR #737, opened 2024-07-08
CircleCI: switched from Starexec-CentOS to Fedora

Tomaqa:


.circleci/config.yml
```diff
@@ -95,42 +95,44 @@ jobs:
...
-                PARALLEL: ON
+                PARALLEL: OFF
         steps:
             - checkout
             - run:
-                name: Release build gcc under devtoolset-8
-                command: |
-                    cat ./ci/build_maximally_static.sh |scl enable devtoolset-8 bash
+                name: Release build gcc in Fedora
+                command: ./ci/run_travis_commands.sh
```
blishko, line 135: Can we keep using `build_maximally_static.sh` script? It's useful to know we can build fully static executable.

Tomaqa: Then it should be necessary to install also `static` packages in a docker image. What about making it a separate job that would just build and not run tests?

Tomaqa:
I guess that the correct CI jobs will appear after the merge into master ..

.circleci/config.yml
blishko: I still don't know what you mean by `static` packages.
It was a job that was just building and not running tests, no?

Tomaqa: I formerly removed static packages from the Fedora image which were there to allow `build_maximally_static`. I included them again in the image in order to allow this test of static build - as it was before.

Tomaqa: https://github.com/usi-verification-and-security/verify-env/pull/4/files

Tomaqa:
I forgot to add the new `build-fedora-release-static` job into workflows - hence the last force-push.

blishko (approved):
LGTM!

PR #738, opened 2024-07-09
CMake: Remove workaround and fix order of GMP libs

blishko:
Apparently, order of static libraries when linking into executable can matter, and for GMP the C++ version gmpxx must come before the C version gmp.
On some systems (like Fedora, ArchLinux), this was not the cases, most likely because when creating the wrapper target GMP::GMP, we specified the libraries in the wrong order.


Tomaqa (approved):
LGTM!

Tomaqa (approved):
LGTM

PR #739, opened 2024-07-16
Rewriters: Hide internal details of rewriters

blishko:
DivModRewriter is exposed to the outside world, but it was broken, because it was including "OsmtInternalException.h" which is not suppose to be exposed and which we do not install. Thus, trying to include "DivModRewriter.h" in another application results in a compilation error.

We fix this problem by hiding the existing rewriters and exposing to the outside world only API methods to use this functionality.


src/rewriters/Rewritings.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef OPENSMT_REWRITINGS_H
+#define OPENSMT_REWRITINGS_H
+
+#include "ArithLogic.h"
+#include "PTRef.h"
+
+#include <optional>
+
+namespace opensmt::rewritings {
```
Tomaqa, line 16:
```suggestion
namespace opensmt {
```
since almost all the functions begin with `rewrite`
It is also better to use e.g. `opensmt::rewriteDistincts` than
```
using namespace opensmt::rewritings;
rewriteDistincts(...);
```

src/rewriters/Rewritings.h
```diff
@@ -0,0 +1,28 @@
...
+
+    PTRef rewriteDistincts(Logic & logic, PTRef fla);
+
+    PTRef rewriteDistinctsKeepTopLevel(Logic & logic, PTRef fla);
+
+    PTRef rewriteDivMod(ArithLogic & logic, PTRef fla);
+
+    std::optional<PTRef> tryGetOriginalDivModTerm(ArithLogic & logic, PTRef term);
+
+} // namespace opensmt::rewritings
```
Tomaqa, line 26:
```suggestion
} // namespace opensmt
```

src/rewriters/Rewritings.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) 2024 Martin Blicha <martin.blicha@gmail.com>
+ *
+ *  SPDX-License-Identifier: MIT
+ *
+ */
+
+#ifndef OPENSMT_REWRITINGS_H
+#define OPENSMT_REWRITINGS_H
```
Tomaqa, line 9: I understand the idea of distinguishing the files with classes and free functions, but I am afraid that it is a bit confusing.
I mean: Since `Rewriter.h` and `Rewritings.h` sound almost the same, it may be confusing which file the developer should choose and it may not be clear what the difference between them is.

Maybe adding `#include "Rewriter.h"` here could help to see that this is an extension and not a completely different thing?
Or to avoid unnecessary dependency, at least provide a short comment?

test/unit/test_LIALogicMkTerms.cc
```diff
@@ -231,8 +214,7 @@ TEST_F(LIALogicMkTermsTest, test_ReverseAuxRewrite) {
     PTRef nested = logic.mkEq(logic.getTerm_IntZero(), logic.mkMod(ite, c));
 
     for (PTRef tr : {term, eq, nested}) {
-
-        PTRef termWithAux = DivModRewriter(logic).rewrite(IteHandler(logic).rewrite(tr));
+        PTRef termWithAux = opensmt::rewritings::rewriteDivMod(logic, IteHandler(logic).rewrite(tr));
```
Tomaqa, line 217: Include `IteHandler(logic).rewrite(tr)` in `Rewritings.h` as well?
-> `rewriteIte(logic, tr)`

test/unit/test_Rewriting.cc
```diff
@@ -1,18 +1,22 @@
...
-#include <ArithLogic.h>
-#include <BoolRewriting.h>
-#include <Logic.h>
-#include <ArithmeticEqualityRewriter.h>
-#include <DivModRewriter.h>
-#include <MainSolver.h>
-#include <DistinctRewriter.h>
+
+#include "ArithLogic.h"
+#include "BoolRewriting.h"
```
Tomaqa, line 11: I am not sure what the difference between `src/rewriters` and `src/simplifiers` is.
Also, to maintain consistency, why this is called `*Rewriting.h` and the new file `Rewritings.h`?

test/unit/test_Rewriting.cc
```diff
@@ -1,18 +1,22 @@
...
+#include "Logic.h"
+#include "ArithmeticEqualityRewriter.h"
+#include "MainSolver.h"
+#include "Rewritings.h"
 
 #include <algorithm>
 
+using namespace opensmt::rewritings;
```
Tomaqa, line 19: I guess that in a test file this is fine.

test/unit/test_Rewriting.cc
```diff
@@ -1,18 +1,22 @@
...
-#include <Logic.h>
-#include <ArithmeticEqualityRewriter.h>
-#include <DivModRewriter.h>
-#include <MainSolver.h>
-#include <DistinctRewriter.h>
+
+#include "ArithLogic.h"
+#include "BoolRewriting.h"
+#include "Logic.h"
+#include "ArithmeticEqualityRewriter.h"
```
Tomaqa, line 13: Why this is not included in `Rewritings.h`?

src/rewriters/Rewritings.cc
```diff
@@ -0,0 +1,43 @@
...
+    PTRef rewriteDivMod(ArithLogic & logic, PTRef term) {
+        return DivModRewriter(logic).rewrite(term);
+    }
+
+    std::optional<PTRef> tryGetOriginalDivModTerm(ArithLogic & logic, PTRef tr) {
+        if (not logic.isVar(tr)) return std::nullopt; // Only variables can match
+        auto symName = std::string_view(logic.getSymName(tr));
+        if (symName.compare(0, DivModConfig::divPrefix.size(), DivModConfig::divPrefix) == 0) {
+            return DivModConfig::getDivTermFor(logic, tr);
+        } else if (symName.compare(0, DivModConfig::modPrefix.size(), DivModConfig::modPrefix) == 0) {
```
Tomaqa, line 38:
```suggestion
        }
        if (symName.compare(0, DivModConfig::modPrefix.size(), DivModConfig::modPrefix) == 0) {
```

src/logics/ArrayTheory.cc
```diff
@@ -7,7 +7,9 @@
 #include "ArrayTheory.h"
 
 #include "ArrayHelpers.h"
-#include "DistinctRewriter.h"
+#include "Rewritings.h"
+
+using namespace opensmt::rewritings;
```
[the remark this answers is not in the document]
blishko: My understanding is that it is a bad idea in header files, in .cpp files, it should not do any harm.
This is related to a general question, if we are gonna have sub-namespaces, or everything will be in a single `opensmt` namespace.

src/rewriters/Rewritings.h
blishko: See my comment about single namespace vs. organizing the codebase into sub-namespaces.

test/unit/test_LIALogicMkTerms.cc
blishko: ITE handling is a bit special, I think, at the moment.

src/logics/ArrayTheory.cc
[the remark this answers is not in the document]
Tomaqa: Update: in header files it is a really bad idea, in cpp files IMO a bit bad. In general, it can still cause name clashes, even though only locally and not globally. It can become a problem if the source file becomes large. But for small files it is probably perfectly fine.

I vote for using sub-namespaces. However, in this case, I am not sure, because the fully scoped names fall back to `opensmt::rewritings::rewrite_*` which is IMO already a bit too verbose.

src/rewriters/Rewritings.h
Tomaqa: I personally prefer to use sub-namespaces a lot, even to such an extent that the meaning of the function/class depends on the outer namespaces. For example, I could do
```
namespace opensmt::rewrite {
    PTRef distincts(Logic & logic, PTRef fla);
    PTRef distinctsKeepTopLevel(Logic & logic, PTRef fla);
    ...
}
```
which avoids redundancy in the code.
However, here in OpenSMT it is not used like this and such a style would currently be inconsistent with the rest.

Hence I rather recommend to use
```
namespace opensmt {
    PTRef rewriteDistincts(Logic & logic, PTRef fla);
    PTRef rewriteDistinctsKeepTopLevel(Logic & logic, PTRef fla);
    ...
}
```
which avoids redundant code at least a bit.

test/unit/test_LIALogicMkTerms.cc
Tomaqa: OK. What about `Substitutor`, `Rewriter`, `ArithmeticEqualityRewriter`?

test/unit/test_Rewriting.cc
Tomaqa: I meant why wrapper functions on `ArithmeticEqualityRewriter` are not included

blishko: I guess because I did not need that for what I wanted to achieve :)

But I guess we should write wrappers for these as well.

.clang-files
```diff
@@ -6,9 +6,10 @@
 ./src/proof/PGCheck.cc
 ./src/rewriters/ArithmeticEqualityRewriter.h
 ./src/rewriters/DistinctRewriter.h
-./src/rewriters/DistinctRewriter.cc
 ./src/rewriters/DivModRewriter.h
 ./src/rewriters/Rewriter.h
+./src/rewriters/Rewritings.cc
+./src/rewriters/Rewritings.h
```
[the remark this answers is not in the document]
blishko: That would not be in the alphabetical order, though :)

test/unit/test_LIALogicMkTerms.cc
Tomaqa: #740

test/unit/test_Rewriting.cc
Tomaqa: #740

Tomaqa (approved):
LGTM

PR #741, opened 2024-08-08
Remove unused files from minisat component

blishko:
This should be all the files from minisat that are not used in OpenSMT.


Tomaqa (approved):
LGTM!

PR #744, opened 2024-08-16
CircleCI: used latest Ubuntu for clang-format

Tomaqa:
There is a bug in older versions of clang-format. Using the latest Ubuntu image fixes the problem.


blishko (approved):
LGTM!

PR #747, opened 2024-08-19
CMake: Do not install redundant file + small cleanup

blishko:
I am pretty sure we do not need to install UFInterpolator.h, it should not be exposed to consumers.

While removing it, I reorganized the corresponding CMakeLists.txt.


Tomaqa (approved):
LGTM!

PR #748, opened 2024-08-21
Extended test_LRALogicMkTerms:ConstantSimplification of 1/3

Tomaqa:
Added additional test that checks the case of creating "1/3" instead of just "1/2".
This can catch errors if the implementation of rationals is changed since 1/3 does not have finite representation if the numerator and denominator are not separated (such as with floats).


blishko (approved):
LGTM!

PR #754, opened 2024-08-29
Dedicated class for unsat cores

Tomaqa:
This allows to do more stuff with unsat cores in the future, or even build a class hierarchy.


src/api/Interpret.cc
```diff
@@ -1284,8 +1284,9 @@ void Interpret::getProof()
 
 void Interpret::getUnsatCore() {
     auto unsatCore = main_solver->getUnsatCore();
+    auto & unsatCoreTerms = unsatCore->getTerms();
```
blishko, line 1287:
```suggestion
    auto const & unsatCoreTerms = unsatCore->getTerms();
```

src/unsatcores/UnsatCore.h
```diff
@@ -0,0 +1,46 @@
...
+    inline vec<PTRef> const & getTerms() const;
+
+protected:
+    void computeTerms() const;
+
+    vec<CRef> clauses;
+
+    mutable std::optional<vec<PTRef>> termsCache{};
+
+    PartitionManager const * partitionManagerPtr;
```
blishko, line 27: This belongs to the builder in my opinion.

Otherwise, we are still leaking internal details to the outside world.

blishko (changes requested):

src/api/Interpret.cc
Tomaqa: Getterms is already const so auto is const as well. It depends whether getterms could ever be made as non-const as well? But for better readability, I can add the const...

Tomaqa:
I removed const because I was afraid that for example minimal unsat cores may require some non-const operations such as checking unsatisfiability. It can be done with a standalone copy of the solver, but I am not sure if it would be efficient. Do you think keeping it const is important?

blishko (approved):
LGTM!

PR #757, opened 2024-09-03
CMake: Update minimal version and how googletest is fetched

blishko:
We update googletest to v1.15.2. As part of that we use `FetchContent_MakeAvailable` to fetch it, because our previous way is now deprecated in CMake 3.30.
In order to use the new command, we need to bump minimal CMake version to `3.14`


Tomaqa (approved):
LGTM!

PR #758, opened 2024-09-04
CMake: consistently used target_sources for object libs

Tomaqa:
PUBLIC usually corresponds to the installed headers, but sometimes also contains source files.
According to CMake doc, using PUBLIC/PRIVATE is probably not crucial for headers. We may use them later for automatic installation of public headers which seems to be supported by newer versions of CMake.


blishko (approved):
LGTM!

PR #759, opened 2024-09-05
Remove CUF logic from frontend

blishko:
The CUF logic code has already been removed from the internals some time ago (see #643).


Tomaqa (commented):
`bindCUFToBV` in `BitBlaster.h` seems to be the only remained place that uses CUF. Shouldn't we remove it as well?

blishko:
I found it but I did not want to touch BitBlaster in this PR.

I have a consecutive PR in mind to disable BV-related files, where I would explain more.

Tomaqa (approved):
In that case... LGTM!

PR #760, opened 2024-09-06
Disable compilation of BV-related files since we do not support BV solving at the moment anyway

blishko:
We can revive these files if #601 is successfully finished.
However, currently we do not support any way for the user to use BV solving, so it does not make sense to keep compiling these files.


Tomaqa (approved):
LGTM!

Issue #666, opened 2024-01-31
Reimplement `SMTConfig` and handling of options/parameters


Tomaqa:
For example, clearly distinguish SMT-LIB options and internal parameters.
Includes #481.

Tomaqa:
Add support for 64-bit numerals in case the option value does not fit into 32 bits.

PR #763, opened 2024-09-08
Make `TermNames` accessible within the solver

Tomaqa:
Further commits will follow ...


Tomaqa:
According to [SMT-LIB](https://smt-lib.org/papers/smt-lib-reference-v2.6-r2021-05-12.pdf):

> There is a new solver option `:global-declarations` that makes all definitions and declarations global and not removable by pop operations. Global declarations can be removed only by the new command `reset`.

> `(reset)` resets the solver completely to the state it had after it was started and before it started
reading commands.

Furthermore, it seems that we do not support this command.

blishko:
Yes, we do not support `reset` (and we should), but that is a different issue.

The point is, that in my example, I turn the `global` option off again before any declarations are made, so all declarations should be scoped.

Tomaqa:
But according to the standard, I consider the example invalid because it changes global declarations without using `reset` which is not allowed. Hence I guess that the correct behavior is to exit with an error or just report an error and ignore the attempt to change the option.
Assuming that we want to comply with the standard.

blishko:
The standard only says that the declarations you have made when the option was on are permanent (and can only be removed with `reset`). You can switch the option on and off freely.

Tomaqa:
I am not sure if the option is meant to affect previous declarations retrospectively.
It says:

> If the solver supports this option, setting it to true causes all declarations and definitions to be global (permanent) as opposed to being added to the current assertion level.

If it is set to true after some scoped declarations already happened, I do not know if they are supposed to be treated as global as well or not.

Tomaqa:
Z3 and CVC5 do not allow changing the option after full initialization of the solver. Hence I will remove one of my regression tests that does that. Hence I do not see any possible problems with the current implementation - except that it should exit with an error on an attempt to change the option after init.

Tomaqa:
Z3 says `(error "line 11 column 33: error setting ':global-decls', option value cannot be modified after initialization")` and CVC5 says `(error "invalid call to 'setOption' for option 'global-declarations', solver is already fully initialized")`. I guess it means to try to change the option after the solver is already outside of the "Start mode":
screenshot

Handling of the invalid setting should be handled here #666 (uhh it is indeed going to be evil)

Tomaqa:
BTW I do not fully understand why in the case of the benchmarks with `get-assignment` the solver returns `unknown` for some variables. But it was there even before, e.g. in benchmark `test/regression/base/QF_LRA/constraints-cooking01-edit.smt2`.

blishko:
I don't know what example you are referring to, but it is possible that once you start asserting, they do not allow to change anymore...

blishko:
We should actually properly implement that diagram.

Tomaqa:
Yeah sorry I already deleted the benchmark. I tried to call `(set-option :global-declarations false)` after some assertions already took place.

src/common/TermNames.h
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    SMTConfig const & config;
+
+    ScopedVector<TermName> names;
+    NameToTermMap nameToTerm;
+    TermToNamesMap termToNames;
+};
+
+class TermNames::Terms {
```
blishko, line 89: This whole class is now not used, no?

Tomaqa: No it is not. I thought I would use it in Interpret but in the end, it did not make sense from performance reasons, because in Interpret you are always interested not only in the name but in the names as well. However, in other use cases, such as using unsat cores via public API, it will be useful because we may be interested just in the terms and not the names.

src/api/Interpret.cc
```diff
@@ -1302,9 +1303,11 @@ void Interpret::getInterpolants(const ASTNode& n)
     vec<PTRef> grouping; // Consists of PTRefs that we want to group
     LetRecords letRecords;
     letRecords.pushFrame();
-    termNames.forEachNamedTerm([&](auto const & name, PTRef term) {
+    auto & termNames = main_solver->getTermNames();
```
blishko, line 1306: Also, why not keep the `forEachNamedTerm` method?
It would operate directly over the map.
This is a weird pattern, because you obtain all the keys and then you iterate over the map by doing a lookup for each key. Looks unnecessarily inefficient.

src/common/TermNames.h
blishko: OK, although I am not a fan of adding functionality just because "it might be useful in the future".

Tomaqa: We can leave the last commit as an open pull request until it is needed.

blishko: I am fine with including it in this PR as well.

src/api/Interpret.cc
Tomaqa: Well, `forEachNamedTerm` always enforces to lookup the term for a name: `action(name, termByName(name));`, which is inefficient when we are only interested in names.
My code is actually equivalent to the previous version. I did not realize that it is inefficient.
Hence it would be better to iterate directly over the map. I will first change the iterators to the map iterators, then I will think of other options.
Using iterators is more transparent and compatible with STL algorithms. Also `forEachNamedTerm` does not allow `TAction` to have more arguments.

blishko:
You are right, it was inefficient before as well!
I thought we were iterating directly over the map entries, but that was not the case.

If you want to improve it, feel free to do it, but I think we can merge as it is now.

blishko (dismissed):
LGTM!

Tomaqa:
I corrected the `const` qualifiers and excluded the last commit.

blishko (approved):
LGTM!

PR #765, opened 2024-09-08
Implement subset minimal unsat core

Tomaqa:


src/api/Interpret.cc
```diff
@@ -1285,14 +1285,24 @@ void Interpret::getProof()
...
+    } else {
+        for (PTRef fla : unsatCore->getTerms()) {
+            if (termNames.contains(fla)) {
+                std::cout << termNames.nameForTerm(fla);
+            }
+            else {
+                std::cout << logic->printTerm(fla);
+            }
+            std::cout << ' ';
+        }
```
blishko, line 1305: Why is this needed?

Tomaqa: Just (slight) performance reasons. In the first branch, we use named terms so all of them must be in `termNames`. In second branch, we iterate over all terms so not all of them are necessarily in `termNames`. If it is not, we just print out string representation of the term, because we do not have any name for it.

Tomaqa: This commit implements option `:print-cores-full` from https://cvc5.github.io/blog/2024/04/15/interfaces-for-understanding-cvc5.html

blishko: I see, can you please document the meaning of this option? And maybe also the `minimal-unsat-cores`, to make it clear that these are subset minimal, and not global minimum.

test/regression/base/generic/trivial_unsat_core2_full.smt2.expected.out
```diff
@@ -0,0 +1,2 @@
+unsat
+( b1 a3 )
```
blishko, line 2: This is not correct. We should not mix terms and name together.
`cvc5` returns 
```
unsat
(
b1
(not b1)
)
```

blishko (changes requested):

test/regression/base/QF_UF/php_3_3_40_min_unsat_core.smt2
blishko, line 1: Do we need these large files? I would prefer to minimize as much as possible the regression tests that we add.

test/regression/base/generic/trivial_unsat_core2_full.smt2.expected.out
Tomaqa: Sounds reasonable.

test/regression/base/QF_UF/php_3_3_40_min_unsat_core.smt2
Tomaqa: I removed them from this PR. However, I think some of them will be needed to track the minimization process more thoroughly. They could be simplified, though.

Tomaqa: -> #770

src/api/MainSolver.cc
```diff
@@ -231,9 +234,15 @@ std::unique_ptr<UnsatCore> MainSolver::getUnsatCore() const {
     if (not config.produce_unsat_cores()) { throw ApiException("Producing unsat cores is not enabled"); }
     if (status != s_False) { throw ApiException("Unsat core cannot be extracted if solver is not in UNSAT state"); }
 
-    UnsatCoreBuilder unsatCoreBuilder{&config, &smt_solver->getResolutionProof(), &pmanager};
+    auto & proof = smt_solver->getResolutionProof();
+    auto unsatCoreBuilder = [&]() -> std::variant<UnsatCoreBuilder, MinUnsatCoreBuilder> {
+        if (not config.minimal_unsat_cores()) { return UnsatCoreBuilder{config, proof, pmanager, termNames}; }
+
+        auto auxSMTSolver = std::make_unique<MainSolver>(logic, config, "unsat core solver");
```
blishko, line 241: It might be more efficient to give this new solver a fresh config. Specifically, the new solver does not need to compute unsat cores anymore.

src/unsatcores/MinUnsatCoreBuilder.cc
```diff
@@ -0,0 +1,82 @@
...
+    minimizeAlg();
+    minimizeFinish();
+}
+
+void MinUnsatCoreBuilder::minimizeInit() {
+    assert(terms.size() >= namedTerms.size());
+    assert(size_t(namedTerms.size()) == namedTermsIdxs.size());
+}
+
+void MinUnsatCoreBuilder::minimizeAlgNaive() {
```
blishko, line 29: This algorithm only minimizes the named terms, no?
Thus this does not compute minimal explanation if used together with the option `print-cores-full`, no?

blishko (commented):
Looks good!
But it seems the two options that you are introducing does not work together correctly.
Maybe the full cores should be introduced in a separate PR? Since it is orthogonal functionality.

Another note is that we should investigate using the [Quickxplain algorithm](https://cdn.aaai.org/AAAI/2004/AAAI04-027.pdf) to compute minimal unsat cores.

src/api/MainSolver.cc
Tomaqa: That is a good point.

src/unsatcores/MinUnsatCoreBuilder.cc
blishko: My intuition is that if `print-cores-full` is set then the names should be ignored and, if minimal cores are required then the minimization algorithm should operate on all terms, not only the named ones.

src/api/Interpret.cc
Tomaqa: Moved to #771

src/api/MainSolver.cc
Tomaqa: Based on this, I actually reconsidered the design and I think both builder classes can be merged. Since we use `MainSolver` and not `SimpSMTSolver`, it is sufficient to extend `UnsatCoreBuilder` constructor of just `Logic &` parameter, which is fine. The internal SMT solver can be constructed inside only if minimization is desired. Do you agree?

blishko: That would be very good, I think!

Tomaqa: Done.

src/api/MainSolver.cc
```diff
@@ -231,7 +231,8 @@ std::unique_ptr<UnsatCore> MainSolver::getUnsatCore() const {
     if (not config.produce_unsat_cores()) { throw ApiException("Producing unsat cores is not enabled"); }
     if (status != s_False) { throw ApiException("Unsat core cannot be extracted if solver is not in UNSAT state"); }
 
-    UnsatCoreBuilder unsatCoreBuilder{&config, &smt_solver->getResolutionProof(), &pmanager};
+    auto & proof = smt_solver->getResolutionProof();
+    UnsatCoreBuilder unsatCoreBuilder{config, logic, proof, pmanager, termNames};
```
blishko, line 235: No need for the temporary variable anymore, no?

src/unsatcores/MinUnsatCoreBuilder.cc
Tomaqa: -> #770

src/unsatcores/UnsatCoreBuilder.cc
```diff
@@ -53,11 +101,94 @@ void UnsatCoreBuilder::computeTerms() {
...
+
+void UnsatCoreBuilder::minimizeInit() {
+    assert(terms.size() >= namedTerms.size());
+    assert(size_t(namedTerms.size()) == namedTermsIdxs.size());
+
+    setSmtSolverConfig();
+    setSmtSolver();
+}
+
+void UnsatCoreBuilder::minimizeAlgNaive() {
```
blishko, line 141: To me it seems like it would be better if the new config and new solver were local to this method. Then, there would be no need for the `set` methods.

blishko (approved):
LGTM!
I would make the solver and config local to the method where they are needed. Feels like the class would simpler that way.
But I am also fine with how it is now.

src/unsatcores/UnsatCoreBuilder.cc
Tomaqa: But then it would not be usable in functions other than `minimizeAlgNaive`, and I plan to add more functions and select one of them based on the configuration.

src/api/MainSolver.cc
Tomaqa: Not necessarily but I still find it more readable.

src/unsatcores/UnsatCoreBuilder.cc
blishko: Then each method can have its own local solver.
The solver is not supposed to be shared between different methods, no?

blishko: My concerns are readability and simplicity.

Tomaqa: Another thought: minimization stuff can be enclosed in a nested class (e.g. `Minimize`) where common functionalities among various strategies can be shared using inheritance, and member variables within the class can be "local" in a sense - within the scope of minimization. Also `minimizeInit` may be replaced by constructor, `minimizeFinish` by destructor, etc. It would still not use the variables as local variables within a function, but the scope of the variables would still be limited.

blishko: How about we merge this now and you can rethink the design when you want to add another minimization strategy?

Tomaqa: I got a similar idea. I will make it local for now and we will see what to do next.

blishko (approved):
LGTM!

PR #768, opened 2024-09-13
Efficient iterating and views on `TermNames`

Tomaqa:
It is often necessary to iterate over all pairs of names and terms of `TermNames`. Previously this was done by looking up the term for each name, which is inefficient. The proposed method has just linear complexity of such iterations.

Furthermore, the PR offers to also iterate only over names or only over terms, which, if the pairs are not desired, is even more efficient.

### DIFF
--- a/src/common/TermNames.h
+++ b/src/common/TermNames.h
@@ -76,11 +76,17 @@ public:
     bool empty() const noexcept { return scopedNamesAndTerms.empty(); }
     std::size_t size() const noexcept { return scopedNamesAndTerms.size(); }
 
+    // A const view to just the terms
+    inline auto terms() const;
+
 protected:
     friend class MainSolver;
 
+    using ScopedNamesAndTerms = ScopedVector<pair<TermName, PTRef>>;
     using NameToTermMap = std::unordered_map<TermName, PTRef>;
     using TermToNamesMap = std::unordered_map<PTRef, std::vector<TermName>, PTRefHash>;
+
+    class TermsView;
 
     // avoid undesired overload resolution with the public `namesForTerm`
     std::vector<TermName> & _namesForTerm(PTRef term) const {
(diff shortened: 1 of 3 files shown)


src/common/TermNames.h
```diff
@@ -38,49 +43,138 @@ class TermNames {
...
         auto & vec = namesForTerm(term);
-        assert(not vec.empty());
-        return vec.front();
+        return pickName(vec);
+    }
+
+    std::optional<PTRef> tryGetTermByName(TermName const & name) const {
+        if (auto it = nameToTerm.find(name); it != nameToTerm.end()) { return it->second; }
+
+        return {};
```
blishko, line 61:
```suggestion
        return std::nullopt;
```

src/common/TermNames.h
```diff
@@ -38,49 +43,138 @@ class TermNames {
...
-        names.popScope([this](TermName const & name) {
+        assert(scopedNames.size() == _terms.size());
+        scopedNames.popScope([this](TermName const & name) {
             auto it = nameToTerm.find(name);
             if (it == nameToTerm.end()) { return; }
             PTRef term = it->second;
-            assert(termToNames.find(term) != termToNames.end());
-            auto & names = termToNames.at(term);
-            names.erase(std::find(names.begin(), names.end(), name));
+            auto & names_ = const_cast<std::vector<TermName> &>(namesForTerm(term));
```
blishko, line 136: Let's not use `const_cast` please.
We can have a (private?) method to return non-const ref to the names, no?

src/common/TermNames.h
```diff
@@ -38,49 +43,138 @@ class TermNames {
...
     }
 
     SMTConfig const & config;
 
-    ScopedVector<TermName> names;
+    ScopedVector<TermName> scopedNames;
     NameToTermMap nameToTerm;
     TermToNamesMap termToNames;
+
+    std::vector<PTRef> _terms;
```
blishko, line 150: Couldn't we have just a `ScopedVector<pair<TermName, PTRef>>`? Wouldn't it make things simpler?

src/common/TermNames.h
Tomaqa: This is a perfectly fine use-case of `const_cast` proposed by Scott Meyer:
https://stackoverflow.com/questions/856542/elegant-solution-to-duplicate-const-and-non-const-getters
- it is ok to cast away const within non-const member function

Private method would cause the following name clash:
1) In most cases, objects are treated as non-const - `namesForTerm` is const method but the object itself is often not const
2) Hence the overload resolution would choose the non-const variant of `namesForTerm` which is however private. It would require the user to cast the object to const in order to call the public version.

Hence we can introduce the function as private but with a different name (e.g. `_namesForTerm`) to avoid the undesired overload resolution.

src/common/TermNames.h
Tomaqa: It makes it easier for the first commit, not easier for the rest commits. However, the first commit is more important, so I will use it this way.

blishko (approved):
LGTM!

PR #770, opened 2024-09-17
Some more complex regression tests on unsat cores

Tomaqa:
The tests can be simplified, though, to reduce their size.


blishko (approved):
LGTM!

PR #771, opened 2024-09-17
Implemented smt2 option ":print-cores-full" (from CVC5)

Tomaqa:


Tomaqa:
See the discussion in #765 on the combination with minimal unsat cores

Tomaqa:
If ":minimal-unsat-cores" is set as well, we have to minimize all terms, not only the named ones.

Tomaqa:
I verified that the current outputs are consistent with CVC5.

test/regression/base/generic/redundant_min_unsat_core2_full.smt2.expected.out
```diff
@@ -0,0 +1,2 @@
+unsat
+( (and b1 b2) (xor b2 b1) )
```
blishko, line 2: There seems to be some problem here, the arguments of `xor` are swapped.
Maybe some bug in the printing?

blishko (commented):
Looks good, but it seems we have some problem with handling/printing `xor`.

test/regression/base/generic/redundant_min_unsat_core2_full.smt2.expected.out
Tomaqa: I noticed that but did not care too much - I thought it is related to some preprocessing.
Nonetheless, I think this should be resolved in a separate issue.
-> #776

Tomaqa: OK!

PR #774, opened 2024-10-03
Getter of the count of inserted flas

Tomaqa:


src/api/MainSolver.cc
```diff
@@ -93,6 +93,7 @@ bool MainSolver::pop() {
         }
         pmanager.invalidatePartitions(mask);
     }
+    insertedFormulasCount -= frames.last().size();
```
blishko, line 96: I fear this has a lot of consequences because now you will be reusing indices when tracking partitions.
I am not sure `PartitionManager` is ready for that.

I will need to spend some time to assess if this would be OK.
I am worried this may break interpolation.

Tomaqa: -> #775

blishko (approved):
LGTM!

PR #775, opened 2024-10-04
Not considering popped flas in the number of assertions

Tomaqa:


Tomaqa:
We should keep `insertedFormulasCount` as it is - it returns the number of inserted formulas, not the number of currently valid assertions. For that, one can use `getCurrentAssertionsView().size()`.

PR #779, opened 2024-10-21
Sorting xor-args in ascending order (same as others)

Tomaqa:
I really do not know why it was previously sorted in descending order.
It seems to be an old line of code from the times before versioning using GitHub.

I only found that xors are handled in `Tseitin.cc` and `Logic.cc:getNewFacts`, but it does not seem that the order has any special meaning...

I believe that the fix should be just fine.


blishko (approved):
LGTM!

PR #780, opened 2024-10-21
Upd. changelog

Tomaqa:
We should probably go for a new major release because the API changed.


blishko (approved):
LGTM!

blishko:
Maybe you can be a bit more verbose in the commit message? :)
No need to shorten the word "update" :)

PR #781, opened 2024-10-22
Printing unsat cores line-wise

Tomaqa:
Inspired by CVC5, I believe that we should print unsat cores line-wise as well.
For example, we also do not print models in one line.
This should increase readability and importantly, it allows to count the size of the core easily.


src/api/Interpret.cc
```diff
@@ -1284,22 +1284,22 @@ void Interpret::getProof()
...
-    std::cout << "( ";
+    std::cout << "(\n";
     auto const & termNames = main_solver->getTermNames();
     if (not config.print_cores_full()) {
         // this is the default: we care only about ':named' terms and their names
         for (PTRef fla : unsatCore->getNamedTerms()) {
             assert(termNames.contains(fla));
             auto const & name = termNames.nameForTerm(fla);
-            std::cout << name << ' ';
+            std::cout << name << "\n";
```
blishko, line 1294:
```suggestion
            std::cout << name << '\n';
```

src/api/Interpret.cc
```diff
@@ -1284,22 +1284,22 @@ void Interpret::getProof()
...
         }
     } else {
         // we explicitly asked to include all terms and to ignore ':named' terms at all
         for (PTRef fla : unsatCore->getTerms()) {
-            std::cout << logic->printTerm(fla) << ' ';
+            std::cout << logic->printTerm(fla) << "\n";
```
blishko, line 1299:
```suggestion
            std::cout << logic->printTerm(fla) << '\n';
```

src/api/Interpret.cc
```diff
@@ -1284,22 +1284,22 @@ void Interpret::getProof()
...
         }
     }
-    std::cout << ')' << std::endl;
+    std::cout << ")" << std::endl;
```
blishko, line 1302:
```suggestion
    std::cout << ')' << std::endl;
```

blishko (commented):
I see you changed chars to strings on purpose, why?
When you want to print a character, just print a character and not one-character string, no?

blishko (approved):
LGTM!

PR #782, opened 2024-10-23
Rewriters: Fix typo in a comment and add assertion

blishko:
The comment incorrectly used `dividend` instead of `divisor` when formulating the condition on modulo.
Also, we should assert not only that the divisor is a constant, but it is an integer constant.


Tomaqa (approved):
OK!

PR #783, opened 2024-10-24
Add Github action for release

blishko:
This adds a GitHub action that we can use to automatically build and upload binaries once a release is published.
The drafting of the release and its publication is still manual, but that is OK since we want to have control over that process.


Tomaqa (approved):
LGTM!

PR #784, opened 2024-10-25
CI: check that the static Linux build is indeed static

Tomaqa:
We require that the linkage of the produced binary contains no dynamic libraries. For that, we use the exit status of `ldd` (since it is Linux), which fails if no dynamic libraries are present. Hence we negate the exit status with `!`.


blishko (approved):
LGTM!

PR #787, opened 2024-10-25
Fixed minimal unsat cores

Tomaqa:
Fixed minimal unsat cores s.t. the removed hidden terms do not block removal of named terms.


src/unsatcores/UnsatCoreBuilder.cc
```diff
@@ -113,6 +113,14 @@ void UnsatCoreBuilder::minimize() {
         return;
     }
 
+    // We need to minimize `namedTerms` which could have been masked by already eliminated non-named terms
+    auto const & termNames = solver.getTermNames();
+    // TODO: make more efficient
+    for (PTRef term : solver.getCurrentAssertionsView()) {
+        if (termNames.contains(term)) { continue; }
+        hiddenTerms.push(term);
```
blishko, line 121: Should we `clear` `hiddenTerms` beforehand?

blishko (commented):
Can you update the name of the PR? I think it is not correct, we do not need to consider eliminating named terms that did not participate in the proof. Those stay eliminated.
The difference is that we consider all unnamed assertions as `hiddenTerms` and not only those extracted from the proof, no?

Also the name is rather long :) The explanation can be moved to the description.

Tomaqa:
It seems that I missed your previous comments, I will take them into account as well.

src/unsatcores/UnsatCoreBuilder.cc
Tomaqa: Good point!

blishko (approved):
LGTM!

PR #790, opened 2024-10-29
Nonlin LA preds

BritikovKI:
### WIP
Removed all of the constraints for the creation of nonlin predicates inside of the functions.
Only the assertions are checked for nonlinearity


src/itehandler/IteToSwitch.cc
```diff
@@ -233,6 +233,12 @@ ite::Dag IteToSwitch::constructIteDag(PTRef root, Logic const & logic) {
 
         } else { // not Ite
             for (PTRef child : t) {
+                if (logic.hasIntegers() || logic.hasReals()) {
+                    if (logic.isNonlin(child)) {
+                        auto termStr = logic.pp(child);
+                        throw ApiException("Nonlinear operations in the formula: " + termStr);
+                    }
+                }
```
blishko, line 241: I don't think we need a check here.
ITE manipulation should be agnostic of arithmetic operations.

src/logics/ArithLogic.cc
```diff
@@ -194,6 +194,18 @@ bool ArithLogic::isLinearTerm(PTRef tr) const {
     return false;
 }
 
+bool ArithLogic::isNonlin(PTRef tr) const {
+    if (isTimes(tr)) {
+        Pterm const & term = getPterm(tr);
+        return (!isConstant(term[0]) && !isConstant(term[1]));
```
blishko, line 200:
```suggestion
        return (not isConstant(term[0]) and not isConstant(term[1]));
```

src/logics/ArithLogic.cc
```diff
@@ -194,6 +194,18 @@ bool ArithLogic::isLinearTerm(PTRef tr) const {
...
+    }
+    if (isRealDiv(tr) || isIntDiv(tr)) {
```
blishko, line 202: Add also check for `mod`.

aehyvari:
The code fails with this input ```
(set-logic QF_LIA)
(define-fun uninterp_mul ((a Int) (b Int)) Int (mod a b))
(assert (= (uninterp_mul 1 2) 0))
(check-sat)
(exit)
```
Here's the output:
```
(error "Divisor must be constant in linear logic")

(error "define-fun returns an unknown sort")

(error "Unknown symbol `uninterp_mul Int Int'")

(error "assertion returns an unknown sort")

sat
```

File: src/logics/ArithLogic.cc [hunk 4]
```diff
@@ -227,9 +238,15 @@ pair<PTRef, PTRef> ArithLogic::splitTermToVarAndConst(PTRef term) const {
...
+        if (not isConstant(fac)) {
+            std::swap(fac, var);
+            if (not isConstant(fac)) {
+                PTRef fac = yieldsSortInt(term) ? getTerm_IntOne() : getTerm_RealOne();
+                return {term, fac};
+            }
+        }
         assert(isConstant(fac));
-        assert(isNumVarLike(var));
+        assert(isNumVarLike(var) || isTimes(var));
```
blishko, line 249: I think this case is going to be more difficult.
We have to think about how terms like `2*x*y` or `3*x*x` should be represented.

src/logics/ArithLogic.cc
```diff
@@ -674,12 +691,7 @@ PTRef ArithLogic::mkTimes(vec<PTRef> && args) {
     PTRef tr = mkFun(s_new, std::move(args));
     // Either a real term or, if we constructed a multiplication of a
     // constant and a sum, a real sum.
-    if (isNumTerm(tr) || isPlus(tr) || isUF(tr) || isIte(tr))
-        return tr;
-    else {
-        auto termStr = pp(tr);
-        throw LANonLinearException(termStr.c_str());
-    }
+    return tr;
```
blishko, line 694: The comment above is outdated at this point.

src/logics/ArithLogic.h
```diff
@@ -8,16 +8,6 @@
...
-class LANonLinearException : public std::runtime_error {
-public:
-    LANonLinearException(char const * reason_) : runtime_error(reason_) {
-        msg = "Term " + std::string(reason_) + " is non-linear";
-    }
-    virtual char const * what() const noexcept override { return msg.c_str(); }
-
-private:
-    std::string msg;
-};
```
blishko, line 20: I think we need to keep it for now, but we need to do the check elsewhere, when we are trying to process the terms inside LASolver.

src/logics/Logic.h
```diff
@@ -251,6 +251,7 @@ class Logic {
     bool isDisequality(PTRef tr) const;            // { return disequalities.has(term_store[tr].symb()); }
     bool isIte(SymRef tr) const;                   // { return ites.has(tr);          }
     bool isIte(PTRef tr) const;                    // { return ites.has(term_store[tr].symb()); }
+    virtual bool isNonlin(PTRef) const { return false; }
```
blishko, line 254: Does `Logic` really need to know about this?

test/regression/base/arithmetic/miniexample2.smt2
```diff
@@ -0,0 +1,8 @@
+(set-logic QF_LIA)
+(declare-fun x () Int)
+(define-fun uninterp_mul ((a Int) (b Int)) Int (* a b))
+(assert (= x x))
```
blishko, line 4: Why do you have this `assert`?

test/unit/test_ArithLogicApi.cc
```diff
@@ -21,25 +23,31 @@ class ArithLogicApiTest: public ::testing::Test {
 };
 
 TEST_F(ArithLogicApiTest, test_LRA) {
+    SMTConfig config;
+    MainSolver solver(lraLogic, config, "test");
```
blishko, line 27: We should not be using `MainSolver` in this file.

test/unit/test_Rewriting.cc
```diff
@@ -84,7 +84,6 @@ TEST(Rewriting_test, test_RewriteDivMod) {
     PTRef mod = logic.mkMod(x,two);
     PTRef fla = logic.mkAnd(logic.mkEq(div, two), logic.mkEq(mod, logic.getTerm_IntZero()));
     PTRef rewritten = rewriteDivMod(logic, fla);
-//    std::cout << logic.printTerm(rewritten) << std::endl;
```
blishko, line 87: Try to avoid making unrelated changes.

src/itehandler/IteToSwitch.cc
BritikovKI: But ideally it would be a check in single place
My idea is to put it in the code where we insert assertions into the code somewhere...
Do you have some ideas where better to put it?

src/logics/ArithLogic.cc
BritikovKI: Won't it work as Times(Times) in the current encoding?
Basically (* 3 (* x y))?
Take a look at the miniexample10.smt2 in the regression, it seems to work at least for some inputs like this


I'll think of it, ideally I would be able to come up with some example to see if what we have rn is not sufficiemt

aehyvari:
For our limited form of axioms this might actually be just enough.  One thing is missing before I can put this to our production: To be sure that we're running the correct version, I'd need to have the version information from opensmt.

src/logics/Logic.h
BritikovKI: It depends on what and how we want to check.

To check nonlin propperly from the outside I need to have some interface through the logic to interact with ArithLogic, because I cannot call ArithLogic directly from the code of MainSolver for example...
Or can I?

src/logics/ArithLogic.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Fixed

test/regression/base/arithmetic/miniexample2.smt2
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Fixed

test/unit/test_Rewriting.cc
BritikovKI: Fixed

src/itehandler/IteToSwitch.cc
blishko: Yes, it should be in one place.
We can put the check when we get the assertions, but I think better place would be to put the check in LASolver.
If we start supporting nonlinear arithmetic, that would be the place to change.

src/logics/ArithLogic.cc
blishko: Then I think the encoding of Times needs to change. The terms should be normalized, same as for other operations.

src/logics/Logic.h
blishko: I guess we need to do it this way for now.

src/itehandler/IteToSwitch.cc
BritikovKI: Moved, no checks in IteToSwitch anymore

src/logics/ArithLogic.h
BritikovKI: Done

test/unit/test_ArithLogicApi.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Should it be done in this PR though?
Seems like a separate issue tbh...

BritikovKI:
Ok, I think generally PR is over

Though, we definitely need some sort of error handling, LANonLinearException produces errors with different format under different complilers rn

Tomaqa:
Kostia probably refers to the fact that we do not catch exceptions on top of the application (such as in `bin/opensmt.cc`), which results in [UB](https://stackoverflow.com/questions/368184/does-it-make-sense-to-catch-exceptions-in-the-main).

blishko:
> Kostia probably refers to the fact that we do not catch exceptions at the top level of the application (such as in `bin/opensmt.cc`), which results in [UB](https://stackoverflow.com/questions/368184/does-it-make-sense-to-catch-exceptions-in-the-main).

I don't see any mention of undefined behaviour in the link you provided. Unspecified, yes. Undefined, no.

blishko:
> Ok, I think generally PR is over
> 
> Though, we definitely need some sort of error handling, LANonLinearException produces errors with different format under different complilers rn. Should it be a different issue though?

I agree that we can catch the exception, probably at the level of a call to `solve` of `MainSolver` and return `unknown`.
SMT-LIB has a concept of "reason for unknown", which we should probably implement.

src/logics/ArithLogic.cc
```diff
@@ -194,6 +194,30 @@ bool ArithLogic::isLinearTerm(PTRef tr) const {
     return false;
 }
 
+bool ArithLogic::isNonlin(PTRef tr) const {
```
blishko, line 197: Shouldn't this answer the question only for the current term?
If we need recursion into subterms, it should be implemented using a visitor.

src/logics/ArithLogic.cc
```diff
@@ -227,9 +250,15 @@ pair<PTRef, PTRef> ArithLogic::splitTermToVarAndConst(PTRef term) const {
         assert(getPterm(term).size() == 2);
         PTRef fac = getPterm(term)[0];
         PTRef var = getPterm(term)[1];
-        if (not isConstant(fac)) { std::swap(fac, var); }
+        if (not isConstant(fac)) {
+            std::swap(fac, var);
+            if (not isConstant(fac)) {
+                PTRef fac = yieldsSortInt(term) ? getTerm_IntOne() : getTerm_RealOne();
+                return {term, fac};
```
blishko, line 257: I am pretty sure this is not correct.
For a term `x*y` this would return a pair `(x,1)`.

src/logics/ArithLogic.cc
blishko: This is a single issue of supporting nonlinear expression in`Logic`, so yes, should be done properly in one PR.

test/unit/test_LRALogicMkTerms.cc
```diff
@@ -173,11 +173,9 @@ TEST_F(LRALogicMkTermsTest, test_SumToZero)
     ASSERT_EQ(sum, logic.getTerm_RealZero());
 }
 
-TEST_F(LRALogicMkTermsTest, test_NonLinearException)
+TEST_F(LRALogicMkTermsTest, test_NoNonLinearException)
 {
-    EXPECT_THROW(logic.mkTimes(x,y), LANonLinearException);
-    PTRef two = logic.mkConst("2");
-    EXPECT_NO_THROW(logic.mkTimes(x,two));
```
blishko, line 180: This should be kept.

src/logics/ArithLogic.cc
BritikovKI: nope, it would not
We return `{term, fac}`, and `term` is already a multiplication of its own
So it would return `{x*y, 1}`

blishko: You are right, I overlooked that you return `term` and not `var`.

src/logics/ArithLogic.cc
BritikovKI: Hmm, I looked into it again.

So in current PR we have Times on multiple arguments as well.

So (* 3 (* x y)) is actually (* x y 3)

But the thing is, do we want to return 3 as a constant here, or is it better to return 
{(* x y 3), 1} ?

BritikovKI: The second way would allow to preserve current code, while change would be more logical but would require much changes

test/unit/test_LRALogicMkTerms.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Initially I wanted it to be this way, but unfortunately in LASolver.cc we register bigger terms first.

So instead of registering `(/ x y)` we register `(* -1 (/ x y))` which LASolver assertions have problem with

BritikovKI: I'm not sure what you mean by using a `visitor`
I looked through the TermVisitor class, but it seems rather complicated, it means that we would need to implement the whole child class to check if LRA/LIA formula has nonlinearity which seems a bit as an overkill
Or did I got the idea wrong a bit?

src/logics/ArithLogic.cc
blishko: You definitely need to return {x*y, 3} in that case.

src/logics/ArithLogic.cc
blishko: The visitor are exactly for that, to collect information of you need to traverse the whole term.

But I think we should not need to traverse whole terms to detect if somewhere inside is a nonlinear operation.

There should be one place, in LASolver, where you know you are dealing with a polynomial, so there you can check if it has degree more than one.

blishko: > Initially I wanted it to be this way, but unfortunately in LASolver.cc we register bigger terms first.
> 
> So instead of registering `(/ x y)` we register `(* -1 (/ x y))` which LASolver assertions have problem with
> 
> And I think if LASolver already starts working it should not accept any Nonlin terms for the reason of separating LA and NLA

What do you mean?
I think in `LASolver::initSolver` before calling `registerArithmeticTerm` you can check if the polynomial has any nonlinear monomials.

src/logics/ArithLogic.cc
BritikovKI: It means that we need to create another multiplication inside of the function, but `splitTermToVarAndConst` is a constant one, and is expected to be const in a lot of functions

I can change it but not sure if it will be Ok

src/logics/ArithLogic.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Just to elaborate on it a bit, it will require to make `termSort` into not const for example

src/api/MainSolver.cc
```diff
@@ -340,7 +340,12 @@ sstat MainSolver::check() {
     if (rval == s_Undef) {
         try {
             rval = solve();
-        } catch (std::overflow_error const & error) { rval = s_Error; }
+        } catch (std::overflow_error const & error) {
+            rval = s_Error;
+        } catch (opensmt::LANonLinearException const & error) {
+            printf("%s\n", error.what());
```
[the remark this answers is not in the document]
BritikovKI: Fixed, By the way, what do you think about `splitToVarAndConst`?
Should we make it into the non-const function? Or maybe create a separate function to split multiplication specifically?

src/logics/ArithLogic.cc
BritikovKI: I've added a special function to split multiplication, pls check it out...
Don't like the solution tbh, but it works...

blishko:
I suggest the following:
First, let's fix the terminology. Our arithmetic sums are basically (multivariate) polynomials.
Polynomial is composed of terms. This is unfortunately an overloaded word in our domain, so I suggest to refer to it as poly-term.
poly-term is a product of a coefficient and zero or more variables (We can restrict our case to variables with degree at most one for now). The coefficient is implicit if it is 1.

With this terminology I suggest to rename the method `splitTermToVarAndConst` to `splitPolyTerm`, but it would be doing the same thing: it would return a pair where the first component is the coefficient of the poly-term and the second coefficient is the rest of the poly-term. In case there is more than one variable, this will create a new term, so the method will have to be non-const. I think that should be OK, I hope.

BritikovKI:
Ok, I unfortunately had to strip out `const` out of the `termSort`, but otherwise it should be Ok...

Also to comment on:

> poly-term is a product of a coefficient and zero or more variables (We can restrict our case to variables with degree at most one for now). The coefficient is implicit if it is 1.

Poly-term might be also a product of a coefficient and other poly-terms (in case if we have int divisions, and those now might be divisions of variables as well)

aehyvari:
> New sym_Int_TIMES_NONLIN/sym_Real_TIMES_NONLIN was added and a draft implementation to handle it
Introducing new operator for nonlin multiplication might've been not the best choice, I'm not sure now how to separate lin and nonlin multiplication otherwise (bc if they have the same operators then sym_Int_TIMES_NONLIN = sym_Int_TIMES)

The new symbols make somehow a lot of sense to me at this point.  But is it really necessary to use the string representation `mul` for nonlinear stuff?  Couldn't it use the same string `*`?  I'm worried that if we print these out (say, for learned clauses, or models in UF, or something similar) nobody will be able to parse them.

blishko:
I agree that we should use the same symbol.
For that, we need to change `SymStore::newSymb` to allow it to skip the check if the same symbol already exist, and force it to create a new one.

src/logics/ArithLogic.h
```diff
@@ -95,6 +95,8 @@ class ArithLogic : public Logic {
 
     SymRef get_sym_Int_TIMES() const { return sym_Int_TIMES; }
     SymRef get_sym_Real_TIMES() const { return sym_Real_TIMES; }
+    SymRef get_sym_Int_TIMES_NONLIN() const { return sym_Int_TIMES_NONLIN; }
+    SymRef get_sym_Real_TIMES_NONLIN() const { return sym_Real_TIMES_NONLIN; }
```
Tomaqa, line 99: Personally I would use the terminology the other way around: 
```suggestion
    SymRef get_sym_Int_TIMES_LIN() const { return sym_Int_TIMES_LIN; }
    SymRef get_sym_Real_TIMES_LIN() const { return sym_Real_TIMES_LIN; }
```
because it the linear multiplication that is actually a special case, accepting only some arguments.

src/logics/ArithLogic.cc
```diff
@@ -75,6 +75,7 @@ std::string const ArithLogic::tk_int_neg = "-";
 std::string const ArithLogic::tk_int_minus = "-";
 std::string const ArithLogic::tk_int_plus = "+";
 std::string const ArithLogic::tk_int_times = "*";
+std::string const ArithLogic::tk_int_times_nonlin = "mul";
```
Tomaqa, line 78:
```suggestion
```

src/logics/ArithLogic.cc
```diff
@@ -90,6 +91,7 @@ std::string const ArithLogic::tk_real_neg = "-";
 std::string const ArithLogic::tk_real_minus = "-";
 std::string const ArithLogic::tk_real_plus = "+";
 std::string const ArithLogic::tk_real_times = "*";
+std::string const ArithLogic::tk_real_times_nonlin = "mul";
```
Tomaqa, line 94:
```suggestion
```

src/logics/ArithLogic.h
```diff
@@ -396,6 +407,8 @@ class ArithLogic : public Logic {
     static std::string const tk_int_plus;
     static std::string const tk_real_times;
     static std::string const tk_int_times;
+    static std::string const tk_real_times_nonlin;
+    static std::string const tk_int_times_nonlin;
```
Tomaqa, line 411:
```suggestion
```

Tomaqa (commented):
I would definitely use the same symbol `*`, this change should only affect the internal representation but not the interface.
But I think that explicitly distinguishing these two cases internally is a good idea!

src/logics/ArithLogic.h
BritikovKI: Removed

src/logics/ArithLogic.cc
BritikovKI: Removed

src/logics/ArithLogic.h
BritikovKI: It is a bit different

Both linear and nonlinear multiplication are special cases:

Nonlin is only of the view (* var_1, var_2, ..., var_n) - necessarily without any constants!

Lin is (* const var)

BritikovKI:
Ok, it is possible to have just a single symbol but it creates a block of problems:

1. We still want to detect same symbols and reuse same symrefs for them, as an example: `1`,`-1`,`0`. Because of them we can't just remove the check for same symrefs
2. If we have two '*' for `INT_MULT` and `INT_MULT_NONLIN` it creates ambiguity for multiplication during `lookupSymbol` operation, problem here as well that two `*` are also somewhat synonimous and separation only happens inside as we split `MULT` into linear multiplication and nonlinear multiplication...

My idea would be to introduce some sort of identifier as an additional parameter for a Symbol, which would separate `NONLIN_MULT` from normal `MULT` and we would be able to distinguish it in the comparison, but all of the "*" would be still parsed into `MULT` initially...

src/logics/ArithLogic.h
Tomaqa: What is e.g. `(* var const)` or `(* var const var)` then?

Tomaqa:
I think that we should go for a more systematic solution. Both multiplication symbols can be clearly distinguished based on their arguments: assuming only binary multiplication, linear mult. must contain a constant. Whether an argument is a constant or a variable should be handled in `lookupSymbol` - it is not enough to just handle sorts. I do not have a deeper knowledge within this area, so currently I cannot give you a hint, but I guess there must be a way how OpenSMT distinguishes variables and constants.

BritikovKI:
I've introduced notion of "subSymbol", it is basically a new SymRef, but it is purely internal as it is not stored in the `symbolTable` so it won't get parsed

src/logics/ArithLogic.h
```diff
@@ -160,12 +162,20 @@ class ArithLogic : public Logic {
...
+    bool isTimesLin(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
+    bool isTimesNonlin(SymRef sr) const { return isIntTimesNonlin(sr) or isRealTimesNonlin(sr); }
     bool isTimes(PTRef tr) const { return isTimes(getPterm(tr).symb()); }
+    bool isTimesLin(PTRef tr) const { return isTimesLin(getPterm(tr).symb()); }
+    bool isTimesNonlin(PTRef tr) const { return isTimesNonlin(getPterm(tr).symb()); }
     bool isIntTimes(PTRef tr) const { return isIntTimes(getPterm(tr).symb()); }
+    bool isIntTimesNonlin(PTRef tr) const { return isIntTimesNonlin(getPterm(tr).symb()); }
     bool isRealTimes(PTRef tr) const { return isRealTimes(getPterm(tr).symb()); }
+    bool isRealTimesNonlin(PTRef tr) const { return isRealTimesNonlin(getPterm(tr).symb()); }
     bool isIntTimes(SymRef sr) const { return sr == sym_Int_TIMES; }
```
Tomaqa, line 175:
```suggestion
    bool isIntTimesLin(SymRef sr) const { return sr == sym_Int_TIMES_LIN; }
```
I think that the symbol `sym_Int_TIMES` no longer makes sense since it is internal. Predicates like `isTimes` can stay e.g. in the form
```
    bool isTimes(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); }
```

src/logics/Logic.h
```diff
@@ -164,7 +164,8 @@ class Logic {
     virtual PTRef mkConst(char const *);
     virtual PTRef mkConst(SRef, char const *);
 
-    SymRef declareFun(std::string const & fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symbolConfig);
+    SymRef declareFun(std::string const & fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symbolConfig,
+                      bool duplicate = false);
```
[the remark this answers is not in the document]
BritikovKI: 1. I don't necessarily see why it is a problem, it gives more possibilities to the outside users potentially, but also for now it wouldn't allow to do anything new with the API I think...
2. Can do this as well, but it also will be function in `Logic.h` in the end, don't have a strong opinion on this though. @aehyvari  @blishko  guys what do you think?
3. I'm not exactly getting your idea here... You're suggesting we should return a vector of symbols? But now we are not storing duplicate functions SymRefs anywhere I think... `NONLIN_MUL` is an exception because it is stored as a part of the `ArithLogic` SymRefs, but it is only created and used withing `ArithLogic` at the moment...

src/logics/ArithLogic.h
BritikovKI: I think `sym_Int_TIMES` is external imho, only `sym_Int_TIMES_NONLIN` is internal...
And if we make `sym_Int_TIMES_LIN` we will have two internal multiplications

BritikovKI: Changed

src/logics/ArithLogic.h
```diff
@@ -160,12 +162,20 @@ class ArithLogic : public Logic {
     bool isIntNeg(SymRef sr) const { return sr == sym_Int_NEG; }
     bool isRealNeg(SymRef sr) const { return sr == sym_Real_NEG; }
 
-    bool isTimes(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
+    bool isTimes(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); };
+    bool isTimesLin(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
```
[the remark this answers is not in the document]
BritikovKI: Changed

src/logics/Logic.h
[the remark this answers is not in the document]
Tomaqa: After a discussion with @BritikovKI, I think we should leave the previous API unchanged and add a protected special function that would allow to create an internal `SymRef` that is derived from an already existing, publicly accessible one. So, in the case of integers, we would first create `sym_Int_TIMES` and then internally two auxiliary symbols `sym_Int_TIMES_LIN` and `sym_Int_TIMES_NONLIN` using a special function that would use `sym_Int_TIMES` as an argument. Although the implementation may not make an explicit connection between the auxiliary symbols and the symbol `sym_Int_TIMES`, I would still use it as a parameter to clarify what is going on. At the moment, it may be sufficient to represent the auxiliary symbols as `SymRef` while not including them in the "public" symbol table.

src/symbols/SymStore.h
```diff
@@ -40,10 +40,15 @@ class SymStore {
     SymStore(SymStore &&) = default;
     SymStore & operator=(SymStore &&) = default;
     // Constructs a new symbol.
-    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig);
+
+    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
```
blishko, line 44: The public interface should stay the same.
The method with a boolean flag should be private.

src/symbols/SymStore.h
```diff
@@ -40,10 +40,15 @@ class SymStore {
...
+                   bool subSymb = false);
     SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args) {
         return newSymb(fname, rsort, args, SymConf::Default);
     }
+    SymRef newUnparsableSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
```
blishko, line 49: We should come up with a better name.
`SymStore` does not care about parsing.

src/tsolvers/lasolver/LASolver.h
```diff
@@ -24,6 +24,17 @@
 #include <unordered_set>
 
 namespace opensmt {
+class LANonLinearException : public std::runtime_error {
+public:
+    LANonLinearException(char const * reason_) : runtime_error(reason_) {
```
blishko, line 29: Let's modernize this and use `std::string_view`.

src/rewriters/DivModRewriter.h
```diff
@@ -36,6 +36,7 @@ class DivModConfig : public DefaultRewriterConfig {
             PTRef modVar = divMod.mod;
             PTRef rewritten = logic.isIntDiv(symRef) ? divVar : modVar;
             if (not inCache) {
+                if (logic.isNonlin(term)) { return term; }
```
blishko, line 39: This is weird. We should not keep terms like `x div y` or `x mod y` in the formula.

src/logics/Logic.h
```diff
@@ -186,6 +186,10 @@ class Logic {
     SymRef declareFun_Commutative_NoScoping_LeftAssoc(std::string const & s, SRef rsort, vec<SRef> const & args) {
         return declareFun(s, rsort, args, SymConf::CommutativeNoScopingLeftAssoc);
     }
+    SymRef declareFun_Multiplication_Duplicate(std::string const & s, SRef rsort, vec<SRef> const & args) {
```
blishko, line 189: This should not be public.

blishko: It should probably not even be here, but in `ArithLogic`.

src/rewriters/DivModRewriter.h
BritikovKI: This terms are being detected by `LASolver` later anyway, but we might use `DivModRewriter` before the nonlin solver potentially, so why not?

src/symbols/SymStore.h
BritikovKI: Though it's also not that good, because this is the symbol which would be seen by the users, using the tool API 🤔

src/logics/Logic.h
BritikovKI: Moved to ArithLogic, isn't public anymore

src/rewriters/DivModRewriter.h
BritikovKI: Throwing exception for now, added issue to handle it propperly #815

src/tsolvers/lasolver/LASolver.h
BritikovKI: Updated constructor to receive `std::string_view`

src/symbols/SymStore.h
BritikovKI: Public interface now is the same, added private method with boolean flag

src/logics/ArithLogic.cc
```diff
@@ -114,10 +114,8 @@ ArithLogic::ArithLogic(Logic_t type)
       sym_Real_MINUS(declareFun_NoScoping_LeftAssoc(tk_real_minus, sort_REAL, {sort_REAL, sort_REAL})),
       sym_Real_PLUS(declareFun_Commutative_NoScoping_LeftAssoc(tk_real_plus, sort_REAL, {sort_REAL, sort_REAL})),
       sym_Real_TIMES(declareFun_Commutative_NoScoping_LeftAssoc(tk_real_times, sort_REAL, {sort_REAL, sort_REAL})),
-      sym_Real_TIMES_LIN(
-          declareFun_Commutative_NoScoping_LeftAssoc(tk_real_times, sort_REAL, {sort_REAL, sort_REAL}, true)),
-      sym_Real_TIMES_NONLIN(
-          declareFun_Commutative_NoScoping_LeftAssoc(tk_real_times, sort_REAL, {sort_REAL, sort_REAL}, true)),
+      sym_Real_TIMES_LIN(declareFun_Multiplication_Duplicate(tk_real_times, sort_REAL, {sort_REAL, sort_REAL})),
+      sym_Real_TIMES_NONLIN(declareFun_Multiplication_Duplicate(tk_real_times, sort_REAL, {sort_REAL, sort_REAL})),
```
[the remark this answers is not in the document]
BritikovKI: Renamed

BritikovKI:
`isNonlin` is temporarily needed, we don't do `RealDiv` refinement in the `DivModRefinement` because we expect originally that by the time of refinement `RealDiv` no longer exist in the instance(removed during parsing). So we need to check if there is nonlin `RealDiv` in the formula as well as multiplication...

src/api/MainSolver.cc
```diff
@@ -333,14 +334,24 @@ sstat MainSolver::check() {
...
     }
     if (isLastFrameUnsat()) { return s_False; }
-    sstat rval = simplifyFormulas();
+    sstat rval;
+    try {
+        rval = simplifyFormulas();
+    } catch (opensmt::NonLinException const & error) {
+        reasonUnknown = error.what();
+        rval = s_Undef;
+        return rval;
```
blishko, line 343:
```suggestion
        return s_Undef;
```

src/api/MainSolver.cc
```diff
@@ -333,14 +334,24 @@ sstat MainSolver::check() {
...
     if (config.dump_query()) printCurrentAssertionsAsQuery();
 
     if (rval == s_Undef) {
         try {
             rval = solve();
-        } catch (std::overflow_error const & error) { rval = s_Error; }
+        } catch (std::overflow_error const & error) { rval = s_Error; } catch (opensmt::NonLinException const & error) {
+            reasonUnknown = error.what();
+            rval = s_Undef;
+        }
```
blishko, line 354:
```suggestion
        } catch (std::overflow_error const & error) { rval = s_Error; } catch (opensmt::NonLinException const & error) {
            reasonUnknown = error.what();
            return  s_Undef;
        }
```

And please format the catch statements.

src/common/CMakeLists.txt
```diff
@@ -24,5 +24,5 @@ include(numbers/CMakeLists.txt)
 install(FILES
         StringMap.h Timer.h inttypes.h IColor.h
         TreeOps.h FlaPartitionMap.h PartitionInfo.h Partitions.h ApiException.h TypeUtils.h
-        NatSet.h ScopedVector.h TermNames.h
+        NatSet.h ScopedVector.h TermNames.h NonLinException.h
```
blishko, line 27: Why do we need to install it?

src/logics/ArithLogic.h
```diff
@@ -170,10 +164,23 @@ class ArithLogic : public Logic {
     bool isIntNeg(SymRef sr) const { return sr == sym_Int_NEG; }
     bool isRealNeg(SymRef sr) const { return sr == sym_Real_NEG; }
 
-    bool isTimes(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
+    bool isTimes(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr) or isTimesUnparsed(sr); };
+    bool isTimesLinNonlin(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); };
+    bool isTimesLin(SymRef sr) const { return isIntTimesLin(sr) or isRealTimesLin(sr); }
+    bool isTimesUnparsed(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
```
blishko, line 170: What does it mean that the times is "unparsed"?

src/logics/ArithLogic.h
```diff
@@ -222,10 +229,10 @@ class ArithLogic : public Logic {
 
     bool isNumVar(SymRef sr) const { return isVar(sr) and (yieldsSortInt(sr) or yieldsSortReal(sr)); }
     bool isNumVar(PTRef tr) const { return isNumVar(getPterm(tr).symb()); }
-    bool isNumVarLike(SymRef sr) const {
-        return yieldsSortNum(sr) and not isPlus(sr) and not isTimes(sr) and not isNumConst(sr);
+    bool isMonomial(SymRef sr) const {
```
blishko, line 232: Do we need to have a method that takes a symbol?
I think we should have it only for `PTRef`.

src/logics/ArithLogic.h
```diff
@@ -346,7 +355,7 @@ class ArithLogic : public Logic {
     bool isLinearTerm(PTRef tr) const;
     bool isLinearFactor(PTRef tr) const;
     pair<Number, vec<PTRef>> getConstantAndFactors(PTRef sum) const;
-    pair<PTRef, PTRef> splitTermToVarAndConst(PTRef term) const;
+    pair<PTRef, PTRef> splitPolyTerm(PTRef term) const;
```
blishko, line 358: Can you add a docstring here to document that it splits the term into a monomial and its coefficient?

src/simplifiers/LA.cc
```diff
@@ -57,14 +57,14 @@ void LAExpression::initialize(PTRef e, bool do_canonize) {
...
-            auto [var, constant] = logic.splitTermToVarAndConst(t);
+            auto [var, constant] = logic.splitPolyTerm(t);
             Real new_c = logic.getNumConst(constant);
             new_c *= c;
             curr_term.emplace_back(var);
             curr_const.emplace_back(std::move(new_c));
         } else {
             // Otherwise it is a variable, Ite, UF or constant
-            assert(logic.isNumVarLike(t) || logic.isConstant(t) || logic.isUF(t));
+            assert(logic.isMonomial(t) || logic.isConstant(t) || logic.isUF(t));
```
blishko, line 67: `isUF` should no longer be needed, no? It should be covered by the monomial check.
Also, please update the comment. For example, ITE is no longer possible here.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -91,9 +92,8 @@ void LASolver::isProperLeq(PTRef tr)
     assert(logic.isLeq(tr));
     auto [cons, sum] = logic.leqToConstantAndTerm(tr);
     assert(logic.isConstant(cons));
-    assert(logic.isNumVar(sum) || logic.isPlus(sum) || logic.isTimes(sum));
-    assert(!logic.isTimes(sum) || ((logic.isNumVar(logic.getPterm(sum)[0]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[1]))) ||
-                                   (logic.isNumVar(logic.getPterm(sum)[1]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[0])))));
+    assert(logic.isNumVar(sum) || logic.isPlus(sum) || logic.isTimesLinNonlin(sum) || logic.isMod(logic.getPterm(sum).symb()) ||
```
blishko, line 95: `mod` and `div` should not be possible here.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -288,6 +288,13 @@ std::unique_ptr<Tableau::Polynomial> LASolver::expressionToLVarPoly(PTRef term)
 //
 // Returns internalized reference for the term
 LVRef LASolver::registerArithmeticTerm(PTRef expr) {
+    if (logic.isTimesNonlin(expr)) throw NonLinException(logic.pp(expr));
```
blishko, line 291: I don't think this is the right place to check.
It seems to me that a better place would be in `getLAVar_single` which now `asserts` it gets linear term.
There we can check if we are getting nonlin times and throw the exception.

src/tsolvers/lasolver/LASolver.h
```diff
@@ -24,6 +24,7 @@
 #include <unordered_set>
 
 namespace opensmt {
+
```
blishko, line 27: Why?

test/regression/base/arithmetic/miniexample3.smt2
```diff
@@ -0,0 +1,8 @@
+(set-logic QF_LIA)
+(declare-fun x () Int)
+(declare-fun y () Int)
+(define-fun uninterp_mul ((a Int) (b Int)) Int (+ (* a b) 10))
+(assert (= x x))
```
blishko, line 5: This seems useless. Please, keep the tests focused.

blishko (changes requested):

src/logics/ArithLogic.h
BritikovKI: We have such methods for all other operations though...
Both `SymRef` and `PTRef`, for example `isNumVar`

But Ok, can remove

src/common/CMakeLists.txt
BritikovKI: We would still need it in LASolver though
Or do you mean to we'll move it purely to LASolver then?

BritikovKI: Added it to the issue #815

src/logics/ArithLogic.h
BritikovKI: There is also a check with `SymRef` at the line 557
I would leave it for now, we have two functions for everything in Arith

src/api/MainSolver.cc
BritikovKI: Fixed

src/api/MainSolver.cc
BritikovKI: Fixed

BritikovKI: I'm not sure what to do about it, this is how Formatter requires us to do it, it complaints otherwise

test/regression/base/arithmetic/miniexample3.smt2
BritikovKI: Fixed

src/tsolvers/lasolver/LASolver.h
BritikovKI: Fixed

src/tsolvers/lasolver/LASolver.cc
BritikovKI: Fixed

src/tsolvers/lasolver/LASolver.cc
BritikovKI: Fixed

src/simplifiers/LA.cc
BritikovKI: Fixed

src/common/CMakeLists.txt
blishko: I meant, only to remove it from this place in CMakeLists.txt. Not the actual file from the `common` directory.

src/common/NonLinException.h
```diff
@@ -0,0 +1,21 @@
+//
+// Created by Konstantin Britikov on 27.11.2024.
+//
```
blishko, line 3: Please, use proper heading.

src/logics/ArithLogic.h
blishko: That I understand, but the name does not convey this meaning.
Is this even used anywhere? I think the existence of this symbol should not be public information, no?

src/logics/ArithLogic.cc
```diff
@@ -193,6 +202,15 @@ bool ArithLogic::isLinearTerm(PTRef tr) const {
     return false;
 }
 
+bool ArithLogic::isNonlin(PTRef tr) const {
```
blishko, line 205: This method should not be needed anymore, no?

src/logics/ArithLogic.h
```diff
@@ -170,10 +164,23 @@ class ArithLogic : public Logic {
     bool isIntNeg(SymRef sr) const { return sr == sym_Int_NEG; }
     bool isRealNeg(SymRef sr) const { return sr == sym_Real_NEG; }
 
-    bool isTimes(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
+    bool isTimes(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr) or isTimesUnparsed(sr); };
+    bool isTimesLinNonlin(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); };
```
blishko, line 168:
```suggestion
    bool isTimesLinOrNonlin(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); };
```

src/logics/ArithLogic.cc
```diff
@@ -529,14 +545,18 @@ PTRef ArithLogic::mkNeg(PTRef tr) {
         PTRef tr_n = mkFun(symref, std::move(args));
         return tr_n;
     }
-    if (isTimes(symref)) { // constant * var-like
+    if (isTimesLin(symref)) { // constant * (var-like \/ times-nonlin)
```
blishko, line 548:
```suggestion
    if (isTimesLin(symref)) { // constant * monomial
```

src/logics/ArithLogic.cc
```diff
@@ -529,14 +545,18 @@ PTRef ArithLogic::mkNeg(PTRef tr) {
...
-        auto [var, constant] = splitTermToVarAndConst(tr);
+        auto [var, constant] = splitPolyTerm(tr);
         return constant == getMinusOneForSort(getSortRef(symref)) ? var : mkFun(symref, {var, mkNeg(constant)});
     }
-    if (isNumVarLike(symref)) {
+    if (isTimesNonlin(symref)) {
+        SRef returnSort = getSortRef(tr);
+        return mkFun(getTimesLinForSort(returnSort), {tr, getMinusOneForSort(returnSort)});
+    }
+    if (isMonomial(symref)) {
```
blishko, line 557: These two cases are identical, no?
`isMonomial` is sufficient.

src/logics/ArithLogic.cc
```diff
@@ -1200,8 +1229,8 @@ pair<Number, PTRef> ArithLogic::sumToNormalizedIntPair(PTRef sum) {
     std::vector<Number> coeffs;
     coeffs.reserve(varFactors.size());
     for (PTRef factor : varFactors) {
-        auto [var, coeff] = splitTermToVarAndConst(factor);
-        assert(ArithLogic::isNumVarLike(var) and isNumConst(coeff));
+        auto [var, coeff] = splitPolyTerm(factor);
+        assert((ArithLogic::isMonomial(var) || ArithLogic::isTimesLin(var)) and isNumConst(coeff));
```
blishko, line 1233:
```suggestion
        assert(ArithLogic::isMonomial(var) and isNumConst(coeff));
```

src/logics/ArithLogic.h
blishko: We have such methods where it makes sense, but here being a monomial is not a property of the symbol, it is a property of the `Pterm`.
You can easily fix the place you mention by passing `PTRef` instead of `SymRef`.

blishko (changes requested):

src/logics/ArithLogic.h
BritikovKI: You are right, this function is used in `isTimes` only

The problem here is that we don't have `private` in the `ArithLogic.h` at all, so I didn't want to brake this "singular style"...
This maybe should be another issue to work on the methods in `ArithLogic` in general to move part of them into the private zone?

blishko: There is `protected` section.

src/logics/ArithLogic.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Fixed

src/logics/ArithLogic.h
BritikovKI: Fixed

src/logics/ArithLogic.cc
BritikovKI: Removed

src/common/NonLinException.h
BritikovKI: Fixed

src/logics/ArithLogic.h
BritikovKI: Removed

src/logics/ArithLogic.h
BritikovKI: I removed the function...

src/common/CMakeLists.txt
BritikovKI: Updated the PR

src/logics/ArithLogic.h
```diff
@@ -170,10 +164,22 @@ class ArithLogic : public Logic {
...
 
-    bool isTimes(SymRef sr) const { return isIntTimes(sr) or isRealTimes(sr); }
+    bool isTimes(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr) or isIntTimes(sr) or isRealTimes(sr); };
+    bool isTimesLinOrNonlin(SymRef sr) const { return isTimesLin(sr) or isTimesNonlin(sr); };
+    bool isTimesLin(SymRef sr) const { return isIntTimesLin(sr) or isRealTimesLin(sr); }
+    bool isTimesNonlin(SymRef sr) const { return isIntTimesNonlin(sr) or isRealTimesNonlin(sr); }
     bool isTimes(PTRef tr) const { return isTimes(getPterm(tr).symb()); }
-    bool isIntTimes(PTRef tr) const { return isIntTimes(getPterm(tr).symb()); }
-    bool isRealTimes(PTRef tr) const { return isRealTimes(getPterm(tr).symb()); }
+    bool isTimesLinNonlin(PTRef tr) const { return isTimesLinOrNonlin(getPterm(tr).symb()); };
```
blishko, line 172:
```suggestion
    bool isTimesLinOrNonlin(PTRef tr) const { return isTimesLinOrNonlin(getPterm(tr).symb()); };
```

src/logics/ArithLogic.cc
```diff
@@ -511,11 +518,11 @@ bool ArithLogic::isNumTerm(PTRef tr) const {
 PTRef ArithLogic::mkNeg(PTRef tr) {
     assert(!isNeg(tr)); // MB: The invariant now is that there is no "Minus" node
     SymRef symref = getSymRef(tr);
```
blishko, line 520: As now all the if conditions are asking questions about `PTRef`, we no longer need to fetch `SymRef` here.

src/logics/ArithLogic.cc
```diff
@@ -669,16 +676,32 @@ PTRef ArithLogic::mkTimes(vec<PTRef> && args) {
...
+    if (vars.size() > 1) {
+        if (coef == PTRef_Undef) {
+            tr = mkFun(getTimesNonlinForSort(returnSort), vars);
+        } else {
+            tr = mkFun(s_new, {coef, mkFun(getTimesNonlinForSort(returnSort), vars)});
+        }
+    } else {
+        tr = mkFun(s_new, {coef, vars[0]});
+    }
+    return tr;
```
blishko, line 704: Note to myself. It would be good if we can avoid this complicated piece of code. It feels like the code in `simplify` should take care of most of this.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -244,7 +243,7 @@ LVRef LASolver::getVarForLeq(PTRef ref) const {
 }
 
 LVRef LASolver::getLAVar_single(PTRef expr_in) {
-
+    if (logic.isTimesNonlin(expr_in) || logic.isRealDiv(expr_in)) throw NonLinException(logic.pp(expr_in));
```
blishko, line 246: Why do check for real div here? I thought we agreed that div should never occur at this point.

blishko (changes requested):

src/logics/ArithLogic.cc
BritikovKI: Symref is used within some of the `if's` though...

BritikovKI: I'll move it inside the branches, but it will add 3 lines of code

blishko: OK I see. then do not change the other if conditions. They can take keep taking the symref.

src/tsolvers/lasolver/LASolver.cc
BritikovKI: The problem here is that `mkRealDiv` may be called for the function declaration, and then this function may be used from outside while assigning the variables in the divisor with  constants, so it could be still handled with LASolver...

src/logics/ArithLogic.h
BritikovKI: Fixed

src/tsolvers/lasolver/LASolver.cc
BritikovKI: Ok, Real division is no longer supported

src/logics/ArithLogic.cc
BritikovKI: Ideally `simplify` would order args in some way
(For example, if there is a constant, it would be always in `new_args[0]`)

Then, indeed, pretty much everything can be handled with `simplify`...

src/api/MainSolver.cc
```diff
@@ -333,14 +334,23 @@ sstat MainSolver::check() {
...
+        return s_Undef;
+    }
 
     if (config.dump_query()) printCurrentAssertionsAsQuery();
 
     if (rval == s_Undef) {
         try {
             rval = solve();
-        } catch (std::overflow_error const & error) { rval = s_Error; }
+        } catch (std::overflow_error const & error) { rval = s_Error; } catch (NonLinException const & error) {
```
blishko, line 350:
```suggestion
        } catch (std::overflow_error const & error) { 
            rval = s_Error;
        } catch (NonLinException const & error) {
```
Is this some formatter thing? The two catch clauses should not be on the same line.

src/logics/ArithLogic.h
```diff
@@ -380,6 +388,9 @@ class ArithLogic : public Logic {
     PTRef mkBinaryGeq(PTRef lhs, PTRef rhs) { return mkBinaryLeq(rhs, lhs); }
     PTRef mkBinaryLt(PTRef lhs, PTRef rhs) { return mkNot(mkBinaryGeq(lhs, rhs)); }
     PTRef mkBinaryGt(PTRef lhs, PTRef rhs) { return mkNot(mkBinaryLeq(lhs, rhs)); }
+    SymRef declareFun_Multiplication_LinNonlin(std::string const & s, SRef rsort, vec<SRef> const & args) {
```
blishko, line 391: There should be a better place for this method than between `mkBinaryGt` and `mkBinaryEq`.

src/logics/ArithLogic.cc
```diff
@@ -401,15 +410,14 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
...
-            auto [var, c] = logic.splitTermToVarAndConst(polyTerm);
+            auto [var, c] = logic.splitPolyTerm(polyTerm);
             auto coeff = logic.getNumConst(c);
             poly.addTerm(var, std::move(coeff));
         } else {
-            assert(logic.isPlus(polyTerm));
+            assert(logic.isPlus(polyTerm) || logic.isTimesNonlin(polyTerm));
             for (PTRef factor : logic.getPterm(polyTerm)) {
-                auto [var, c] = logic.splitTermToVarAndConst(factor);
+                auto [var, c] = logic.splitPolyTerm(factor);
```
blishko, line 420: Nonlinear multiplication should be handled as a factor, not as a sum, no? I think this is not correct.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -91,9 +92,7 @@ void LASolver::isProperLeq(PTRef tr)
     assert(logic.isLeq(tr));
     auto [cons, sum] = logic.leqToConstantAndTerm(tr);
     assert(logic.isConstant(cons));
-    assert(logic.isNumVar(sum) || logic.isPlus(sum) || logic.isTimes(sum));
-    assert(!logic.isTimes(sum) || ((logic.isNumVar(logic.getPterm(sum)[0]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[1]))) ||
-                                   (logic.isNumVar(logic.getPterm(sum)[1]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[0])))));
+    assert(logic.isNumVar(sum) || logic.isPlus(sum) || logic.isTimesLinOrNonlin(sum));
```
blishko, line 95: Would this express the cases better?

```suggestion
    assert(logic.isPlus(sum) or logic.isTimesLin(sum) or logic.isMonomial(sum);
```

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -296,11 +295,11 @@ LVRef LASolver::registerArithmeticTerm(PTRef expr) {
...
     }
 
-    if (logic.isNumVar(expr) || logic.isTimes(expr)) {
+    if (logic.isNumVar(expr) || logic.isTimesLin(expr)) {
         // Case (1), (2a), and (2b)
-        auto [v,c] = logic.splitTermToVarAndConst(expr);
-        assert(logic.isNumVar(v) || (laVarMapper.isNegated(v) && logic.isNumVar(logic.mkNeg(v))));
+        auto [v,c] = logic.splitPolyTerm(expr);
         x = getLAVar_single(v);
+        assert(logic.isNumVar(v) || (laVarMapper.isNegated(v) && logic.isNumVar(logic.mkNeg(v))));
```
blishko, line 302: Why has the assert moved? Just keep it in the original place, no?

blishko (commented):
Some more comments.

src/api/MainSolver.cc
Tomaqa: It seems like a bug in `clang-format`. We use `LLVM` style which seems to imply `BreakBeforeBraces: Attach` (see [here](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html)). All other styles correctly wrap each `catch`, but not in the case of `Attach` style.

src/common/NonLinException.h
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+#ifndef OPENSMT_NONLINEXCEPTION_H
+#define OPENSMT_NONLINEXCEPTION_H
+
+namespace opensmt {
+class NonLinException : public std::runtime_error {
+public:
+    NonLinException(std::string_view const reason_) : runtime_error(std::string(reason_)) {
+        msg = "Term " + std::string(reason_) + " is non-linear";
```
Tomaqa, line 14:
```suggestion
```

src/common/NonLinException.h
```diff
@@ -0,0 +1,23 @@
...
+    }
+    virtual char const * what() const noexcept override { return msg.c_str(); }
+
+private:
+    std::string msg;
```
Tomaqa, line 19:
```suggestion
```

src/common/NonLinException.h
```diff
@@ -0,0 +1,23 @@
...
+*  SPDX-License-Identifier: MIT
+*/
+
+#ifndef OPENSMT_NONLINEXCEPTION_H
+#define OPENSMT_NONLINEXCEPTION_H
+
+namespace opensmt {
+class NonLinException : public std::runtime_error {
+public:
+    NonLinException(std::string_view const reason_) : runtime_error(std::string(reason_)) {
```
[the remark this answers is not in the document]
Tomaqa: Do not use non-owning objects (such as `string_view`) within exceptions.

src/logics/ArithLogic.h
```diff
@@ -222,10 +228,10 @@ class ArithLogic : public Logic {
 
     bool isNumVar(SymRef sr) const { return isVar(sr) and (yieldsSortInt(sr) or yieldsSortReal(sr)); }
     bool isNumVar(PTRef tr) const { return isNumVar(getPterm(tr).symb()); }
-    bool isNumVarLike(SymRef sr) const {
-        return yieldsSortNum(sr) and not isPlus(sr) and not isTimes(sr) and not isNumConst(sr);
+    bool isMonomial(PTRef tr) const {
```
Tomaqa, line 231: Provide the variant for `SymRef` as well.

src/rewriters/DivModRewriter.h
```diff
@@ -36,6 +37,7 @@ class DivModConfig : public DefaultRewriterConfig {
             PTRef modVar = divMod.mod;
             PTRef rewritten = logic.isIntDiv(symRef) ? divVar : modVar;
             if (not inCache) {
+                if (!logic.isConstant(divisor)) throw NonLinException(logic.pp(term));
```
Tomaqa, line 40:
```suggestion
                if (not logic.isConstant(divisor)) { throw NonLinException(logic.pp(term)); }
```

src/symbols/SymStore.h
```diff
@@ -73,6 +79,9 @@ class SymStore {
     vec<SymRef> symbols;
     SymbolAllocator ta{1024};
     vec<char *> idToName;
+
+    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
+                   bool subSymb);
```
Tomaqa, line 84:
```suggestion
                   bool isInternal);
```

src/symbols/SymStore.cc
```diff
@@ -39,11 +39,12 @@ SymStore::~SymStore() {
         free(idToName[i]);
 }
 
-SymRef SymStore::newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
+SymRef SymStore::newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
+                         bool subSymb) {
```
Tomaqa, line 43:
```suggestion
                         bool isInternal) {
```

src/symbols/SymStore.h
Tomaqa: Or maybe just `internal`

src/symbols/SymStore.h
```diff
@@ -73,6 +79,9 @@ class SymStore {
     vec<SymRef> symbols;
     SymbolAllocator ta{1024};
     vec<char *> idToName;
+
+    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
```
Tomaqa, line 83:
```suggestion
    SymRef newSymbImpl(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
```

src/symbols/SymStore.cc
```diff
@@ -39,11 +39,12 @@ SymStore::~SymStore() {
         free(idToName[i]);
 }
 
-SymRef SymStore::newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
+SymRef SymStore::newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
```
Tomaqa, line 42:
```suggestion
SymRef SymStore::newSymbImpl(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
```

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -244,7 +243,7 @@ LVRef LASolver::getVarForLeq(PTRef ref) const {
 }
 
 LVRef LASolver::getLAVar_single(PTRef expr_in) {
-
+    if (logic.isTimesNonlin(expr_in)) throw NonLinException(logic.pp(expr_in));
```
Tomaqa, line 246:
```suggestion
    if (logic.isTimesNonlin(expr_in)) { throw NonLinException(logic.pp(expr_in)); }
```

src/common/NonLinException.h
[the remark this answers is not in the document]
Tomaqa: I take that back. It should not be an issue to construct the exception like this, it would only be an issue if the message was *stored* as a non-owning object.

src/logics/ArithLogic.h
Tomaqa: .. OK don't, as already discussed above. Maybe adding a comment that the property is related to `PTRef` only would help.

src/logics/ArithLogic.cc
BritikovKI: In this case it is the same though.. Operations like
x1 * x2 * x3
and
x1 + x2 + x3 
Are split in the same manner, elements of the arithmetic operation extracted one by one and analyzed separately

BritikovKI: Or maybe I don't fully understand what `toPoly` function does, but from what I get it splits operation into terms

src/tsolvers/lasolver/LASolver.cc
BritikovKI: It was moved because `getLAVar_single` contains check for nonlin multiplication and will throw exception if it is detected. Otherwise assertion will fail because v is nonlin multiplication and is not in the scope of assertion...

I can either change assertion disjunction or add handling of the nonlin multiplication before the assertion, but those are both non-perfect solutions either... Tell me which one you prefer!

Tomaqa: I do not understand. Since `expr` must be either `NumVar` or `TimesLin`, how can `splitPolyTerm(expr)` yield `TimesNonLin`? I would expect that `getLAVar_single` never throws here. I do not even consider it necessary to add `assert(not isTimesNonLin(v))` because I really do not see a rationale for how it could appear here unless I am missing something.

BritikovKI: Problem is that at this point `expr` can be potentially `timesNonlin`, since we don't have `NonLinArith` and send potentially `Nonlin` queries into the LASolver (we can truly understand if it is Lin or Nonlin only after simplifications).
`splitPolyTerm` is general call, that works both for linear and nonlinear terms...

Tomaqa: We are talking about an `if` block guarded by the condition `if (logic.isNumVar(expr) || logic.isTimesLin(expr))`, hence why I assume that it is impossible that `expr` is `TimesNonlin`.

BritikovKI: It is due to our representation of formulas

basically 2*x*y is (TimesLin 2 (TimesNonlin x y))
Therefore this check doesn't check that there is no nonlin multiplication in the formula

Tomaqa: OK. Then I would add an additional assert before `getLAVar_single` that considers nonlin as well, to 1) make it more under control and 2) to document that this is the expected behavior.

src/logics/ArithLogic.cc
Tomaqa: Can @blishko comment on this? I am not following up on the issue at the moment.

blishko: I am also confused right now. How is the term like`5 * x1 * x2` gonna be represented?
One nonlinear multiplication with three children, or like `5 * (x1*x2)` so the top-level multiplication is linear and the second one is nonlinear?

src/symbols/SymStore.cc
```diff
@@ -39,11 +39,12 @@ SymStore::~SymStore() {
         free(idToName[i]);
 }
 
-SymRef SymStore::newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
+SymRef SymStore::newSymbImpl(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig,
+                             bool isInternal) {
```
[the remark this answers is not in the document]
aehyvari: That part of the code is very much historical, to put it nicely.  `noScoping` would probably mean that the symbol is a reserved word.  that could well be the reason why it's `isInterpreted`.   I'm not convinced that the choices I made when introducing those fields were very informed.  I encourage you to look at them with a critical eye.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -91,9 +92,7 @@ void LASolver::isProperLeq(PTRef tr)
     assert(logic.isLeq(tr));
     auto [cons, sum] = logic.leqToConstantAndTerm(tr);
     assert(logic.isConstant(cons));
-    assert(logic.isNumVar(sum) || logic.isPlus(sum) || logic.isTimes(sum));
-    assert(!logic.isTimes(sum) || ((logic.isNumVar(logic.getPterm(sum)[0]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[1]))) ||
-                                   (logic.isNumVar(logic.getPterm(sum)[1]) && logic.isOne(logic.mkNeg(logic.getPterm(sum)[0])))));
+    assert(logic.isPlus(sum) or logic.isTimesLin(sum) or logic.isMonomial(sum));
     (void) cons; (void)sum;
```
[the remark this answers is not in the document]
BritikovKI: Same question as above, I can do it in separate PR

src/api/MainSolver.cc
BritikovKI: It is a formatter thing unfortunately, and clang-format throws an error otherwise :((

src/symbols/SymStore.h
```diff
@@ -40,9 +40,15 @@ class SymStore {
...
-    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig);
+
+    SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
+        return newSymbImpl(fname, rsort, args, symConfig, false);
+    };
     SymRef newSymb(char const * fname, SRef rsort, vec<SRef> const & args) {
-        return newSymb(fname, rsort, args, SymConf::Default);
+        return newSymbImpl(fname, rsort, args, SymConf::Default, false);
+    }
+    SymRef newInternalSymb(char const * fname, SRef rsort, vec<SRef> const & args, SymbolConfig const & symConfig) {
```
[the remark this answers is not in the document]
BritikovKI: Fixed

Issue #791, opened 2024-10-30
Unify regression test scripts


Tomaqa:
There is a lot of redundancy in the run scripts `test/regression/*/run*.sh`.
It should be convenient to have just one reliable script and use it with different parameters.

Tomaqa:
Also, some run scripts contain command `hostname` which does not exist on every platform.

PR #792, opened 2024-10-30
Separated regression tests of unsat cores

Tomaqa:
Similarly to e.g. `interpolation`, this PR moves all regression tests that use the command `(get-unsat-core)` into a separate directory `unsatcores`.
The redundancy between the run scripts should be resolved in the future ->#791.

Git messed up the tracked information about renaming some `*.smt2.expected.*` files because it apparently does not care about the name of the files but only about the contents - which are in these cases the same.


blishko (approved):
LGTM!

PR #795, opened 2024-11-01
CI: Update macos environment

blishko:
It is important to keep updating the environment, otherwise, homebrew may not have ready-to-pour packages any more (e.g. CMake).


Tomaqa (approved):
LGTM!

PR #797, opened 2024-11-01
Removed problematic use of `move_iterator`s

Tomaqa:
`std::move` does not require that the argument iterators are `move_iterator`s.
(It would be necessary only if `std::copy` was used: [example](https://stackoverflow.com/questions/26432990/are-stdmove-and-stdcopy-identical)).
Furthermore, `move_iterator` should not be dereferenced more than once: [cppreference](https://en.cppreference.com/w/cpp/iterator/move_iterator/operator*).


blishko (approved):
LGTM!

PR #802, opened 2024-11-04
Refactored and extended "redundant" unsat-core regression tests

Tomaqa:
These tests quite extensively cover corner cases of the options `:minimal-unsat-cores` and `:print-cores-full`.


blishko (dismissed):
Seems a bit excessive, no?
Are you sure we need all these tests?

Tomaqa:
I removed some of the tests (especially full cores). Some of them may still be redundant but can be useful in the future for verifying optimized algorithms.

PR #805, opened 2024-11-06
Moved number-related files into separate directory

Tomaqa:
I am making changes in the handling of numbers according to PR #751.
As a first step, it will be useful to separate this kind of files into a separate directory.
I do not want to make these renaming changes in one PR because keeping track of the actual modifications would be difficult then.


src/logics/ArithLogic.cc
```diff
@@ -1,10 +1,11 @@
 #include "ArithLogic.h"
 
 #include <common/ApiException.h>
-#include <common/FastRational.h>
 #include <common/InternalException.h>
 #include <common/StringConv.h>
 #include <common/TreeOps.h>
+#include <common/numbers/FastRational.h>
+#include <common/numbers/Number.h>
```
blishko, line 8: Do you really need to include the same file both in `.h` and the `.cc` file?
If it is included in the header, I would prefer to not include it again the the implementation file.

src/logics/BVLogic.h
```diff
@@ -10,7 +10,9 @@ Module: New Logic for BitVector
 
 #include "Logic.h"
 
-#include <common/NumberUtils.h>
+#include <common/numbers/NumberUtils.h>
+
+#include <gmpxx.h>
```
blishko, line 15: Why this new header?

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -7,12 +7,13 @@
 
 #include "LASolver.h"
 #include "FarkasInterpolator.h"
-#include <simplifiers/LA.h>
 #include "LIAInterpolator.h"
 #include "CutCreator.h"
 
-#include <models/ModelBuilder.h>
+#include <common/numbers/Real.h>
```
blishko, line 13: Isn't this unnecessary?

src/tsolvers/lasolver/Simplex.cc
```diff
@@ -5,6 +5,7 @@
 #include "Simplex.h"
 
 #include <common/InternalException.h>
+#include <common/numbers/Number.h>
```
blishko, line 8: Also seems unnecessary.

src/logics/ArithLogic.cc
Tomaqa: I will remove `Number.h`.
`FastRational.h` is not necessarily included by `ArithLogic.h` if `FAST_RATIONALS` is not defined, but since it is hardcoded this way it would not work in the other case anyway.
So I will remove it as well ...

src/logics/BVLogic.h
Tomaqa: I do not remember. Seems redundant. Removing.

src/tsolvers/lasolver/LASolver.cc
Tomaqa: I think not because it is not clear where `Real` comes from. I already experienced a few times that if not included explicitly, and if the order of the other included headers changes e.g. during a refactoring, the compilation may fail and even only on some platforms. Furthermore, it was not trivial to fix such compilation errors.
Hence I consider a better practice to not always rely on indirect inclusions by other headers. Here it would be probably better to move the inclusion into `LASolver.h`, though.

blishko (approved):
OK!

PR #807, opened 2024-11-08
Another suggestion for implementing versioning

aehyvari:


src/options/SMTConfig.cc
```diff
@@ -639,7 +639,7 @@ namespace opensmt {
   void SMTConfig::printVersion( )
   {
       char version_string[]
-          = OPENSMT_FULL_VERSION " - " GIT_COMMIT "\n";
+          = OPENSMT_GIT_DESCRIPTION "\n";
```
BritikovKI, line 642: But this would lose the description of the current opensmt version, no?
Wouldn't it make sense to print both version and current git description?

aehyvari: They probably serve a different purpose and it gets confusing if we print both.  For example in this commit the above will print the version as 2.8.0 but the version encoded in cmake files is 2.8.1.

BritikovKI: Hmm, ok, you are right here 🤔

PR #809, opened 2024-11-08
Model: Add a simple method to extend a model with value for a new variable

blishko:
This functionality is helpful for Golem and could be interesting for others as well. By adding the possibility of extension, we do not need to expose ModelBuilder to the outside world.

### DIFF
--- a/src/models/Model.h
+++ b/src/models/Model.h
@@ -1,16 +1,20 @@
-//
-// Created by Martin Blicha on 12.06.20.
-//
+/*
+ * Copyright (c) 2020-2024, Martin Blicha <martin.blicha@gmail.com>
+ *
+ *  SPDX-License-Identifier: MIT
+ *
+ */
+
+#ifndef OPENSMT_MODEL_H
+#define OPENSMT_MODEL_H
 
 #include <logics/Logic.h>
 #include <pterms/PTRef.h>
 
-#include <algorithm>
 #include <cassert>
+#include <memory>
+#include <span>
 #include <unordered_map>
-
-#ifndef OPENSMT_MODEL_H
-#define OPENSMT_MODEL_H
 
 namespace opensmt {
 class Model {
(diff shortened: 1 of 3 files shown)


src/models/Model.h
```diff
@@ -26,6 +30,9 @@ class Model {
         Logic const & logic, SymRef sr,
         std::string const & formalArgDefaultPrefix); // Return a string that is not equal to the argument
 
+    static std::unique_ptr<Model> extend(Model const & base, std::span<std::pair<PTRef, PTRef>> extension);
```
Tomaqa, line 33: Why `static`?

Tomaqa: .. and we may even consider a non-const (or `&&`) variant.

blishko: You mean `static` compared to member function?
Just because that this is an operation _on_ the object rather than an operation _of_ the object.
If you prefer member function, I could live with that.

Variants on what? The model?

Tomaqa: In my view, it is an operation of the object. It makes a copy of itself and adds extra variables.
I think that static functions are often used as factory functions where the object is not available yet. Personally, it just confused me why it is `static` and I needed to take a look whether there is an extra reason fo that.

I think that the case where the object is also non-const is useful too, if we do not need the previous model.
For example:
```
Model & Model::extend(PTRef var, PTRef val);
```
However, to avoid name clash, it should be named differently than the variant `std::unique_ptr<Model> Model::extend(PTRef var, PTRef val) const;`.
I propose to get inspired by [algorithm](https://en.cppreference.com/w/cpp/algorithm) library and name the const variant `extendCopy`, or `copyAndExtend`. Or, to name the non-const variant `inplaceExtend`.

blishko: I changed it to be a member function, though I disagree with it.
The reason is related to why we cannot have `inplaceExtend`: the models were designed to be immutable, once you get a model you can only use it to evaluate terms, but you cannot change it. The members related to the evaluation of the base terms are `const` and I think they should stay that way.

That's why I also see the extension as operation on the model, not of the model. The model itself is immutable (from logical perspective).

Tomaqa (approved):
LGTM!

src/models/Model.h
Tomaqa: I think that in the future, it should be still perfectly fine to at least allow to move-construct the new extended model.

PR #811, opened 2024-11-11
Helper script to create and possibly push annotated release tag

### LINKED ISSUE #808: Use annotated tags for releases
At the moment in opensmt's git, the following happens
```
% git describe
OpenSMT-2.0.0-1533-g319deeed
```
That is, the version that git reports for opensmt is 2.0.0.  It look like [we should use annotated tags when doing releases](https://git-scm.com/docs/git-tag).

Tomaqa:
Resolves #808 
Extracts full version from `CMakeLists.txt` and creates the corresponding release tag.
Optionally, it also pushes the tag to `origin`.

### DIFF
--- a/scripts/release-tag.sh
+++ b/scripts/release-tag.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+ROOT=$(cd $(dirname "$0")/..; pwd)
+
+## Extract the full version from the top-level CMakeLists.txt
+function current_version {
+    local str
+    for v in MAJOR MINOR PATCH; do
+        ## Concatenate particular version parts
+        str+=$(sed -n "s/set(OPENSMT_VERSION_${v} \([^ )]*\))/\1/p" <"$ROOT"/CMakeLists.txt).
+    done
+    str=${str%.}
+
+    printf '%s\n' $str
+}
+
+VERSION=$(current_version)
+TAG=v${VERSION}
+git tag -a $TAG -m "Release $VERSION" || exit $?
+
+printf 'Created release tag "%s"\n' $TAG
+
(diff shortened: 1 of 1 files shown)


blishko:
Can you please add a description of how to script work?
Where it is looking for the information and how it extracts it?
I think I figured it out but it took me a minute.
Probably it would be good to put it all as a comment in the script and also as into the description of this PR.

Also, the script needs to be run in the  same directory as the top-level CMakeLists file, correct?

scripts/release-tag.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+function current_version {
+    local str
+    for v in MAJOR MINOR PATCH; do
+        str+=$(sed -rn "s/set\(OPENSMT_VERSION_${v} ([^ )]*)\)/\1/p" CMakeLists.txt).
```
aehyvari, line 6: `sed -r` doesn't work on Mac's standard `sed`.  It could be good to avoid using the extension.

aehyvari: Also, does this write to `CMakeLists.txt`?  Isn't that what `-r` does?

Tomaqa: `-r` stands for extended regular expressions. I will remove it ... The script does not overwrite `CMakeLists.txt`, but I will make it more clear.

Tomaqa:
> Also, the script needs to be run in the same directory as the top-level CMakeLists file, correct?

Yes. But since it is supposed to be used by developers and not users, I do not consider it an issue ...

blishko (approved):
LGMT!

PR #813, opened 2024-11-16
Checking that every `Number` of sort Int in `mkConst` is indeed integer

Tomaqa:
This assertion fails with substitutions and also in `FarkasInterpolator::getFlexibleInterpolant` in `TEST_F(LIAInterpolationTest, test_InterpolationLRASat)`. We should probably forbid such substitutions. In the interpolation, I am not sure if the test makes sense or if the situation must be handled more properly.


Tomaqa:
As a consequence of the check, I removed substitutions for integers when it introduces rational coefficients.
I also disable flexible Farkas interpolants where the current implementation IMO does not make sense for integers.

blishko:
I think flexible interpolation still makes sense even for interger arithmetic. But the interpolant probably needs some normalization to avoid rational coefficients.

Tomaqa:
I ran a performance check on QF_LRA and QF_LIA SMT-LIB benchmarks. The PR does not affect QF_LRA too much:
2024-QF_LRA-v2 8 0-10-g161b6756 scatter
where v2.8.0-10 is the version before the PR and v2.8.0-13 is after applying the 3 commits.

In the case of QF_LIA, satisfiable instances are often affected:
2024-QF_LIA-v2 8 0-10-g161b6756 scatter
but on average, the runtime remains very similar (with some tiny speedup). In the case of sat instances, I would guess that the differences are just caused by the randomized search.

Hence, I think that the change is OK.

src/logics/ArithLogic.cc
```diff
@@ -351,6 +351,45 @@ namespace {
...
+                polyRef.divideBy(coeff);
+                assert(hasInts or not logic.yieldsSortInt(var));
+                if (hasInts and logic.yieldsSortInt(var)) {
+                    if (not std::ranges::all_of(polyRef, [](auto const & term) { return term.coeff.isInteger(); })) {
+                        return PTRef_Undef;
+                    }
+                }
+            }
+
+            if (hasInts) { poly = std::move(polyCp); }
```
blishko, line 385: I believe this whole thing is unnecessary, because equalities in integer logic are normalized (see `ArithLogic::mkBinaryEq` -> `return sumToNormalizedEquality(diff);`) Although, this may not be true if UFs or arrays are present (we should have a look at that).

Normalized means that the coefficients of different terms in the polynomial are coprime.
Thus, I believe we can just check if the coefficient of the variable is one or not, and bail out if it is not.
No need to copy the polynomial.

Tomaqa:
I re-measured the comparison, because in the case of version v2.8.0-10 the benchmarks were run for all logics, while in the case of v2.8.0-13 only for QF_LRA and QF_LIA. It turns out this may have some effect on the runtimes. Long story short, now the comparison should be consistent.

QF_LRA:
2024-QF_LRA-v2 8 0-10-g161b6756 scatter

QF_LIA:
2024-QF_LIA-v2 8 0-10-g161b6756 scatter

I think that the message remains similar: sometimes the performance drops, sometimes rises. v10 solves slightly more instances, but the total avg. time is greater.

I will try to implement your comment above to see what difference it makes.

blishko (approved):
LGTM!

Tomaqa:
There is a twist: the last commit influenced integers way more than the previous commits. Importantly, unsat instances are now influenced as well.
2024-QF_LIA-v2 8 0-10-g161b6756 scatter

Now, there is a trade-off: there is an almost 20% average speedup of solved instances, but there are also 12 more time-outs, including 6 unsat instances (while no unsat instances time-out only in the old version):
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20210219-Dartagnan/ReachSafety-Loops/sum_array-1-O0
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20210219-Dartagnan/ReachSafety-Loops/sum_array-2-O0
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20210219-Dartagnan/ConcurrencySafety-Main/stack_longest-2-O0
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20210219-Dartagnan/ConcurrencySafety-Main/stack_longer-2-O0
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20220307-SMPT/SharedMemory-PT-000200/RC-01
Indet y-only (unsat): results/2024/non-incremental/QF_LIA/20220307-SMPT/SharedMemory-PT-000200/RC-09

Tomaqa:
The issue was that I bailed out from the substitution if the coefficient of the variable was not 1, but I should have also considered if it was -1. Now it seems to be effectively more-or-less the same as before.
2024-QF_LIA-v2 8 0-10-g161b6756 scatter

Three more instances timed out but they seem to be in the area where v10 was also close to the timeout.

src/logics/ArithLogic.cc
```diff
@@ -351,6 +351,33 @@ namespace {
...
+                    return term.var != PTRef_Undef and not logic.isVar(term.var);
+                })) {
+                return PTRef_Undef;
+            }
+        }
+
+        Real coeff = poly.getCoeff(var);
+        bool const isMinusOne = coeff.isOne();
+        coeff.negate();
+        bool const isOne = coeff.isOne();
```
blishko, line 366: While this is the most efficient way to determine these flags, it is also very confusing.

On the first read, the line `bool const isMinusOne = coeff.isOne()` immediately raised a red flag in my mind.
Only after thinking about it for a while, I realized what is actually going on.

Please also add a note that these boolean flags refer to the state when the variable is put on the other side of equality (and therefore these are properties of the coefficient only **after** it has been negated).

src/logics/ArithLogic.cc
```diff
@@ -351,6 +351,36 @@ namespace {
...
+            }
+        }
+
+        Real coeff = poly.getCoeff(var);
+        bool const wasOne = coeff.isOne();
+        coeff.negate();
+
+        // properties of the variable coefficient after we moved it to the rhs
+        bool const isOne = coeff.isOne();
+        bool const isMinusOne = wasOne;
```
blishko, line 369: Nice! Thanks!

blishko (approved):
LGMT!

PR #817, opened 2024-12-27
Tests: Remove outdated test

blishko:
The file test_LIACutSolver contains only one test, which is currently not testing any useful functionality.
It seems like this test originated in 2019 as part of preparation for a new type of LIA solver. However, there has been a lot of changes in the architecture of the arithmetic solver since then and this "LIACutSolver" was removed a long time ago.

The test attempts to create an instance of simplex, however, there are no bounds asserted, nor any basic variables are marked as out of bounds, so the check does not actually do any work.
Simplex is not prepared to be tested in this way anyway, so this test should be removed.


Tomaqa (approved):
LGTM!

Would this be detected by a static analysis such as code coverage?
Would it be too painful to include sth. like that in the pipeline?

blishko:
I don't think it would, the code that is being run is valid, it's just that the test got outdated.

PR #818, opened 2025-01-07
Fixed CMake definition of the executable regarding the version

Tomaqa:
The build fails with `cmake -DBUILD_EXECUTABLES=OFF ..` because the definition of `OPENSMT_GIT_DESCRIPTION` is not conditioned by `BUILD_EXECUTABLES` and rule `OpenSMT-bin` does not exist without the parameter.
Hence I moved the definition into the `CMakeLists.txt` that concerns the binary specifically and is already conditioned.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,12 +125,6 @@ endif()
 
 add_subdirectory(${CMAKE_SOURCE_DIR})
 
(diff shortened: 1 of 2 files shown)


blishko (approved):
LGTM!

PR #819, opened 2025-01-07
Included `cassert` in standalone headers

Tomaqa:
I failed to build OpenSMT as an external library when I swapped the order of `StringConv.h` and `NumberUtils.h`. The reason is that some headers use assertions but do not include `<cassert>`, even indirectly via other headers.


blishko (approved):
LGTM!

PR #820, opened 2025-01-07
Replaced non-const `getTermNames` by `tryAddNamedAssertion`/`tryAddTermNameFor`

Tomaqa:
Non-const getter `getTermNames` seems necessary only for consecutive `insert` after `insertFormula`.
I consider it clearer to use a function such as `insertNamedFormula` directly, similar to what is done in the SMT-LIB2 format.
This way it is also clearer that the user is *not* responsible for removing the names in incremental queries, this is what `pop` implicitly does.

Instead of directly removing `getTermNames`, I used a deprecated attribute since we changed the API.
However, it causes a lot of warnings because the compiler still favors the non-const getter in many situations, even when the const is sufficient - one would have to use e.g. `std::as_const` to avoid the warning.
Here I am not sure, maybe we should remove the non-const getter right away. It may break existing code after the update but only if someone used it for `insert`.

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -531,14 +531,14 @@ PTRef Interpret::parseTerm(const ASTNode& term, LetRecords& letRecords) {
         if (tr == PTRef_Undef) return tr;
 
         if (strcmp(name_attr.getValue(), ":named") == 0) {
-            auto & termNames = main_solver->getTermNames();
             ASTNode& sym = **(name_attr.children->begin());
             assert(sym.getType() == SYM_T or sym.getType() == QSYM_T);
-            if (termNames.contains(sym.getValue())) {
-                notify_formatted(true, "name %s already exists", sym.getValue());
+            char const * str = sym.getValue();
+            bool const success = main_solver->tryAddTermNameFor(tr, str);
+            if (not success) {
+                notify_formatted(true, "name %s already exists", str);
                 return PTRef_Undef;
             }
-            termNames.insert(sym.getValue(), tr);
         }
         return tr;
     }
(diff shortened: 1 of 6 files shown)


Tomaqa:
I did not realize that we should indeed get rid of the warnings. Hence I propose to remove the non-const getter entirely.

Tomaqa:
While there still does not seem to be a reason for a user why `insertNamedFormula` would not be sufficient, we need it in `Interpret.cc`. In the end, I believe there may be some corner cases where it would be beneficial to do something in between the insertion of the formula and of the term name.
Hence, I still keep the non-const getter but add a note that `insertNamedFormula` should be preferred if possible.

blishko (approved):
LGTM!

Do we still need the non-const getter?

blishko:
Can we mark the non-const getter deprecated? If none of us sees a good use case for it, we should probably remove it.

blishko:
The only use case in our codebase is in Interpreter, right?
```
if (strcmp(name_attr.getValue(), ":named") == 0) {
            auto & termNames = main_solver->getTermNames();
            ASTNode& sym = **(name_attr.children->begin());
            assert(sym.getType() == SYM_T or sym.getType() == QSYM_T);
            if (termNames.contains(sym.getValue())) {
                notify_formatted(true, "name %s already exists", sym.getValue());
                return PTRef_Undef;
            }
            termNames.insert(sym.getValue(), tr);
        }
```

I think this just needs an API at the level of `MainSolver`. Like `tryNameTerm`?

blishko (approved):
LGTM!

blishko (approved):

PR #823, opened 2025-02-26
LASolver: Simplify handling of bounds

blishko:
This commit removes a redundant member `in_bounds` of `LABoundStore` and updates the code accordingly.
This member was used only for initialization, and we don't need it since we can store the bounds directly in the `bounds` member. The initialization also becomes simpler.

### DIFF
--- a/src/tsolvers/lasolver/LABounds.cc
+++ b/src/tsolvers/lasolver/LABounds.cc
@@ -84,7 +84,6 @@ bool LABoundStore::isUnbounded(LVRef v) const { return getBounds(v).size() == 0;
 
 void LABoundStore::clear() {
     this->ba.clear();
-    this->in_bounds.clear();
     this->bounds.clear();
 }
 
@@ -94,21 +93,33 @@ void LABoundStore::ensureReadyFor(LVRef v) {
     }
 }
 
-void LABoundStore::updateBound(BoundInfo bi) {
-    auto & varBounds = getBounds(bi.v);
-    for (LABoundRef bound : {bi.ub, bi.lb}) {
-        unsigned idx = varBounds.size();
-        varBounds.push(bound);
+LABoundStore::BoundInfo LABoundStore::allocBoundPair(LVRef v, BoundValuePair boundPair) {
+    ensureReadyFor(v);
+    LABoundRef ub = ba.alloc(bound_u, v, std::move(boundPair.upper));
+    LABoundRef lb = ba.alloc(bound_l, v, std::move(boundPair.lower));
+    auto & varBounds = bounds.at(getVarId(v));
+    varBounds.push(ub);
+    varBounds.push(lb);
+    return BoundInfo{v, ub, lb};
+}
+
+LABoundStore::BoundInfo LABoundStore::allocBoundPairAndSort(LVRef v, BoundValuePair boundPair) {
+    auto res = allocBoundPair(v, std::move(boundPair));
+
+    // Get the two newly allocated bounds in the right position
+    auto & varBounds = getBounds(v);
+    assert(varBounds.size_() >= 2);
+    unsigned indexToSort = varBounds.size_() - 2;
+    for (unsigned index : {indexToSort, indexToSort + 1}) {
+        [[maybe_unused]] const LABoundRef bound = varBounds[index];
         bound_lessthan lessthan(ba);
-        for (; idx > 0; --idx) {
-            if (lessthan(varBounds[idx - 1], varBounds[idx])) {
-                break;
-            }
-            std::swap(varBounds[idx - 1], varBounds[idx]);
-            ba[varBounds[idx]].setIdx(LABound::BLIdx{idx});
+        while (index > 0 and not lessthan(varBounds[index - 1], varBounds[index])) {
+            std::swap(varBounds[index - 1], varBounds[index]);
+            ba[varBounds[index]].setIdx(LABound::BLIdx{index});
+            --index;
         }
-        assert(varBounds[idx] == bound);
-        ba[varBounds[idx]].setIdx(LABound::BLIdx{idx});
+        assert(varBounds[index] == bound);
+        ba[varBounds[index]].setIdx(LABound::BLIdx{index});
     }
     // Post-condition; indices must correspond to the place in the bound list
     assert([&](){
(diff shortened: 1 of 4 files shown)


src/tsolvers/lasolver/LABounds.cc
```diff
@@ -94,21 +93,21 @@ void LABoundStore::ensureReadyFor(LVRef v) {
...
-    for (LABoundRef bound : {bi.ub, bi.lb}) {
-        unsigned idx = varBounds.size();
-        varBounds.push(bound);
+void LABoundStore::placeLastTwoBounds(LVRef v) {
+    auto & varBounds = getBounds(v);
+    assert(varBounds.size_() >= 2);
+    unsigned indexToSort = varBounds.size_() - 2;
+    for (unsigned index : {indexToSort, indexToSort + 1}) {
+        LABoundRef bound = varBounds[index];
+        (void)bound;
```
Tomaqa, line 102:
```suggestion
        [[maybe_unused]] LABoundRef const bound = varBounds[index];
```

src/tsolvers/lasolver/LABounds.cc
```diff
@@ -94,21 +93,21 @@ void LABoundStore::ensureReadyFor(LVRef v) {
...
         bound_lessthan lessthan(ba);
-        for (; idx > 0; --idx) {
-            if (lessthan(varBounds[idx - 1], varBounds[idx])) {
-                break;
-            }
-            std::swap(varBounds[idx - 1], varBounds[idx]);
-            ba[varBounds[idx]].setIdx(LABound::BLIdx{idx});
+        while (index > 0 and not lessthan(varBounds[index - 1], varBounds[index])) {
```
Tomaqa, line 104:
```suggestion
        for (; index > 0 and not lessthan(varBounds[index - 1], varBounds[index]); --index) {
```
Not too important but I would still use `for` since we use the loop variable in a standard way.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -25,30 +25,33 @@ PtAsgn LASolver::getAsgnByBound(LABoundRef br) const {
...
     else {
         const Real& constr = logic.getNumConst(const_tr);
-        bi = boundStore.allocBoundPair(v, this->getBoundsValue(v, constr, true));
+        LABoundStore::BoundInfo bi = boundStore.allocBoundPair(v, this->getBoundsValue(v, constr, true));
         br_pos = bi.lb;
         br_neg = bi.ub;
     }
+    assert(isInitializationPhase == (status == INIT));
+    if (not isInitializationPhase) {
+        boundStore.placeLastTwoBounds(v);
```
Tomaqa, line 53: This is the only place where `placeLastTwoBounds` is used. Hence I would try to embed it into `allocBoundPair`, which could also allow the use of the last two elements of `varBounds` more transparently. Currently, it is a bit cryptic and fragile. Maybe function `placeLastTwoBounds` can be used only in the implementation file and not cause confusion in the header file.

src/tsolvers/lasolver/LASolver.h
```diff
@@ -88,12 +88,11 @@ class LASolver : public TSolver {
     // Possible internal states of the solver
     typedef enum { INIT, INCREMENT, SAT, UNSAT, NEWSPLIT, UNKNOWN, ERROR } LASolverStatus;
 
-    void setBound(PTRef leq);
     bool assertBound(LABoundRef boundRef);
 
     PTRef getVarPTRef(LVRef v) const { return laVarMapper.getVarPTRef(v); }
 
-    LABoundStore::BoundInfo addBound(PTRef leq_tr);
+    void addBound(PTRef leq_tr, bool isInitializationPhase);
```
Tomaqa, line 95: I would be happier if we could get rid of this flag or use a clearer name. In any case, it is quite unclear how `addBound(ref, true)` vs. `addBound(ref, false)` differ. Could we just internally use `status == INIT`?

Tomaqa (changes requested):

src/tsolvers/lasolver/LABounds.cc
blishko: Updated! Thanks!

src/tsolvers/lasolver/LABounds.cc
blishko: I prefer `while` variant. Seems more readable to me, but that's personal taste.
I can change it if you insist.

src/tsolvers/lasolver/LASolver.h
blishko: Yes, deciding based on the internal flag is better! Updated!

src/tsolvers/lasolver/LASolver.cc
blishko: I am thinking now of having the original `allocBoundPair` and `allocBoundPairAndSort` that would call `allocBoundPair` and then put the newly allocated bounds in the right place. In `LASolver` we would call one or the other depending on the initialization flag.

Does that sound reasonable?

src/tsolvers/lasolver/LABounds.cc
Tomaqa: I do not.

src/tsolvers/lasolver/LASolver.cc
blishko: Can you check the method now? I tried to make it more concise.

src/tsolvers/lasolver/LABounds.cc
```diff
@@ -94,21 +93,20 @@ void LABoundStore::ensureReadyFor(LVRef v) {
...
-    auto & varBounds = getBounds(bi.v);
-    for (LABoundRef bound : {bi.ub, bi.lb}) {
-        unsigned idx = varBounds.size();
-        varBounds.push(bound);
+void LABoundStore::placeLastTwoBounds(LVRef v) {
+    auto & varBounds = getBounds(v);
+    assert(varBounds.size_() >= 2);
+    unsigned indexToSort = varBounds.size_() - 2;
+    for (unsigned index : {indexToSort, indexToSort + 1}) {
+        [[maybe_unused]] LABoundRef bound = varBounds[index];
```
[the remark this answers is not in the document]
Tomaqa: It reads easier to me, that we just want to check the value and modify it.

Tomaqa (commented):
Looks good! I pointed just a few more details.

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -25,45 +25,31 @@ PtAsgn LASolver::getAsgnByBound(LABoundRef br) const {
...
 
     int tid = Idx(logic.getPterm(leq_tr).getId());
     if (LeqToLABoundRefPair.size() <= tid) {
         LeqToLABoundRefPair.growTo(tid + 1);
     }
-    LeqToLABoundRefPair[tid] = LABoundRefPair{br_pos, br_neg};
+    LeqToLABoundRefPair[tid] = boundRefs;
 
+    int br_pos_idx = boundStore[boundRefs.pos].getId();
+    int br_neg_idx = boundStore[boundRefs.neg].getId();
```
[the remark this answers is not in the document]
blishko: `LABoundRefPair` is quite a trivial struct. Do you seen any benefit in moving it? To me it's a bit weird, like trying to move an `int`?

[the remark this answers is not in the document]
Tomaqa: It's a struct of ints, but it will indeed be very similar to copying. You can keep it ..

Tomaqa (dismissed):
LGTM!

Tomaqa (approved):
Still good!

PR #824, opened 2025-03-10
Always run CMake from root Makefile (to upd. `OPENSMT_GIT_DESCRIPTION`)

### LINKED ISSUE #821: `OPENSMT_GIT_DESCRIPTION` is not updated on new builds automatically
In my experience, the OpenSMT version printed by the binary (implemented in PR #801) is not up-to-date after adding new commits and building them. The issue seems to be that in the build dir, it is not sufficient to run `make` but it is necessary to explicitly run `cmake ..`. Only then, the binary `opensmt.cc.o` is built as well.

It would be good if the Makefile itself was sensitive to the change of the version and was able to trigger the build of `opensmt.cc.o` as well.

Tomaqa:
It should be generally more robust if we always run `cmake` from our provided main `Makefile` (and not just the build).
Runtime overhead is minimal.

Resolves #821

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,9 @@ all: release debug parallel-release parallel-debug
 
 .PHONY: release debug
 .PHONY: _dir-release _dir-debug _build-release _build-debug
+## Run `cmake` unconditionally of the existence of the build dirs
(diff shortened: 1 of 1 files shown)


aehyvari (approved):
LGTM!

PR #827, opened 2025-03-21
LASolver: Small refactoring

blishko:
Small cleanup in LASolver.

Include the removal of method `updateBound`.
This method was checking if the given bound (PTRef) has already been processed or not.
However, it was only called in a situation where it is guaranteed that the bound has not been seen before (the theory solver keeps track of processed inequalities with `isInformed` and `setInformed` methods in `LASolver::declareAtom`.

### DIFF
--- a/src/tsolvers/lasolver/LASolver.cc
+++ b/src/tsolvers/lasolver/LASolver.cc
@@ -38,6 +38,8 @@ void LASolver::addBound(PTRef leq_tr) {
     }();
 
     int const tid = Idx(logic.getPterm(leq_tr).getId());
+    // Must be a new inequality
+    assert(tid >= LeqToLABoundRefPair.size() or (LeqToLABoundRefPair[tid] == LABoundRefPair{LABoundRef_Undef, LABoundRef_Undef}));
     if (LeqToLABoundRefPair.size() <= tid) {
         LeqToLABoundRefPair.growTo(tid + 1);
     }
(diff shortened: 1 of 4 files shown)


src/tsolvers/lasolver/LAVarMapper.cc
```diff
@@ -31,7 +31,6 @@ void LAVarMapper::registerNewMapping(LVRef lv, PTRef e_orig) {
 
     PTId id_pos = logic.getPterm(e_orig).getId();
     PTId id_neg = logic.getPterm(logic.mkNeg(e_orig)).getId();
-    assert(!hasVar(id_pos));
```
Tomaqa, line 34: Why?

Tomaqa: And how often does that `laVarToPTRef.growTo(lv.x+1, PTRef_Undef);` happen?

src/tsolvers/lasolver/LASolver.cc
```diff
@@ -38,6 +38,8 @@ void LASolver::addBound(PTRef leq_tr) {
     }();
 
     int const tid = Idx(logic.getPterm(leq_tr).getId());
+    // Must be a new inequality
+    assert(tid >= LeqToLABoundRefPair.size() or (LeqToLABoundRefPair[tid] == LABoundRefPair{LABoundRef_Undef, LABoundRef_Undef}));
     if (LeqToLABoundRefPair.size() <= tid) {
         LeqToLABoundRefPair.growTo(tid + 1);
```
Tomaqa, line 44: I guess we already discussed this. Is it not performance bottleneck?

Tomaqa (commented):
Looks good! I think we can also format `LAVARMapper` files, in a separate commit.

src/tsolvers/lasolver/LASolver.cc
blishko: It's not for OpenSMT.
I am hitting this in Golem on a few occasions, but I think I can try some workaround on the model checker level.

Note also that we tried using maps in the past (https://github.com/usi-verification-and-security/opensmt/pull/500), but there was a few percent performance hit for OpenSMT on SMT-LIB benchmarks.

src/tsolvers/lasolver/LAVarMapper.cc
blishko: > Why?

This is already checked at the start of the method (line 25).

> And how often does that laVarToPTRef.growTo(lv.x+1, PTRef_Undef); happen?

This is not the problem, LVRefs are computed only for arithmetic terms in current formula. There are usually not that many.
Only mappings *from* PTRefs are (sometimes) problem (no mappings *to* PTRef).

Tomaqa (approved):
LGTM!

PR #828, opened 2025-03-25
Added missing const modifiers in `Logic` and `ArithLogic` functions

Tomaqa:
I believe that in none of the cases it is reasonable to have the functions as non-const.


blishko (approved):
LGTM!

PR #835, opened 2025-04-20
Core: Fix split clause handling

### LINKED ISSUE #834: Assertion failure in ALIA instance
For the following SMT-LIB2 script, `OpenSMT` hits assertion failure (in debug mode):

```
(set-logic QF_ALIA)
(declare-fun t () Int)
(declare-fun a () (Array Int Int))
(declare-fun s () Int)
(declare-fun i () Int)
(assert (and (= t t) (= (store a 1 i) (store a 1 1)) (= (store a 1 1) (store (store a 1 0) s 1))))
(check-sat)
(assert (= s t))
(check-sat)
```

Output:
```
sat
Assertion failed: (value(c[0]) != l_False or value(c[1]) != l_False), function attachClause, file CoreSMTSolver.cc, line 317.
```

blishko:
So-called split clauses from array solver can contain many literals. It can happen that a split clause is satisfied,
but the first two literals are falsified.
The code for processing new split clauses would previosuly take the clause as is and try to attach it to the SAT solver. However, the method `attachClause` has a check that at the time of attaching, the first two literals cannot be both falsified.
We already had a check in place for the situation where the two first literals are falsified, but there is at least one undecided literal. However, it may happen that there is no undecided literal (all others are satisfied). In this case, we also need to swap one of the satisfied literals with the first, falsified, literal.

Fixes #834.


src/smtsolvers/TheoryIF.cc
```diff
@@ -88,16 +89,18 @@ TPropRes CoreSMTSolver::handleNewSplitClauses(SplitClauses & splitClauses) {
         auto & splitClause = splitClauses[index];
         unsigned satisfied = 0;
         unsigned unknown = 0;
+        std::optional<int> notFalsifiedIndex = std::nullopt;
```
BritikovKI, line 92: Making `impliedIndex` optional wasn't necessary tbh, but I see why you did it...
Overall PR is good!

BritikovKI (approved):
Lgtmt!

src/smtsolvers/TheoryIF.cc
blishko: True. But I prefer to be explicit in the code in saying "Here I don't have any value yet!" and "Here I have a value!"

PR #837, opened 2025-04-22
Substitutions: Add one more transitive closure computation at the end

### LINKED ISSUE #836: Assertion failure in incremental solving
For the following instance, OpenSMT encounters assertion violation.
```
(set-logic QF_LIA)
(declare-fun b () Bool)
(declare-fun x () Int)
(declare-fun y () Int)

(assert (and (= x y) b (or (not b) (= x 1))))
(push 1)
(assert (> x 0))
(check-sat)
(push 1)
(assert (= y 1))
(check-sat)
(push 1)
(assert (> y 0))
(check-sat)
```

Output:
```
sat
sat
Assertion failed: (not allSubst.has(key)), function current, file MainSolver.h, line 240.
```

blishko:
The computation of substitutions is a fixed point operation: we keep extracting new substitutions until there are no new one. In each iteration we compute a transitive closure of the newly uncovered substitution. For example, we can discover simultaneously that `x -> y` and `y -> 1`. In this case we can actually derive `x -> 1`, and we prefer that over `x -> y`. That's why we compute transitive closure of the substitutions.
At the end, we previously simply collected all substitutions from the iterations and returned them.
However, this can lead to the situation described above where we miss the simplification of `x -> y` to `x -> 1` (if `x -> y` is discovered in one iteration and `y -> 1` is discovered some time later).

Therefore, we should make one last transitive closure at the end.

Fixes #836.


BritikovKI (approved):
Lgtmt

PR #839, opened 2025-05-09
Substitutions: Add detection of conflict in arithmetic substitutions

blishko:
Previously, the test I am adding here would trigger an assertion failure, because it would try to create a substitution `x -> 1/3` in the integer logic.
To prevent that, we can detect such cases and abort the computation of substitutions early. We have to pass along the fact that the formula from which we are computing substitutions is unsatisfiable.

### DIFF
--- a/src/logics/ArithLogic.cc
+++ b/src/logics/ArithLogic.cc
@@ -261,7 +261,29 @@ namespace {
         }
     }
 
-    Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<LAPoly> & zeroPolynomials) {
+    class SubstitutionsOrConflict {
+    public:
+        static SubstitutionsOrConflict conflict() { return SubstitutionsOrConflict(); }
+        SubstitutionsOrConflict(Logic::SubstMap && subst) : maybeSubstitutions{std::move(subst)} {}
+
+        [[nodiscard]] bool isConflict() const { return not maybeSubstitutions.has_value(); }
+
+        [[nodiscard]] Logic::SubstMap const & getSubstitutions() const {
+            assert(not isConflict());
+            return maybeSubstitutions.value();
+        }
+
+    private:
+        SubstitutionsOrConflict() : maybeSubstitutions{std::nullopt} {}
+
+        std::optional<Logic::SubstMap> maybeSubstitutions;
+    };
+
+    /// Given a system of arithmetic polynomials @p zeroPolynomials, collects substitutions of the form `v -> c`
+    /// where v is a variable and c is a constant. Applies the discovered substitutions to simplify the polynomials
+    /// in the process.
+    /// Returns early if a conflict has been detected.
+    SubstitutionsOrConflict collectConstantSubstitutions(ArithLogic & logic, std::vector<LAPoly> & zeroPolynomials) {
         Logic::SubstMap substitutions;
 
         while (true) {
@@ -282,7 +304,8 @@ namespace {
                 if (poly.size() == 1) {
                     auto const & term = *poly.begin();
                     if (term.var == PTRef_Undef) { // FALSE equality
-                        continue;
+                        assert(not term.coeff.isZero());
+                        return SubstitutionsOrConflict::conflict();
                     }
                     // poly is "x = 0"
                     PTRef var = term.var;
@@ -299,7 +322,11 @@ namespace {
                     auto const & [var, coeff] = *poly.begin();
                     if (not substitutions.has(var)) {
                         auto val = -((poly.begin() + 1)->coeff) / coeff;
-                        substitutions.insert(var, logic.mkConst(logic.getSortRef(var), val));
+                        auto sortRef = logic.getSortRef(var);
+                        if (sortRef == logic.getSort_int() and not val.isInteger()) {
+                            return SubstitutionsOrConflict::conflict();
+                        }
+                        substitutions.insert(var, logic.mkConst(sortRef, val));
                         new_keys.push(var);
                     }
                     processedIndices.push_back(i);
@@ -400,8 +427,10 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
     std::transform(top_level_arith.begin(), top_level_arith.end(), std::back_inserter(polynomials),
                    [&](auto const & eq) { return ptrefToPoly(eq, logic); });
 
-    auto constSubstitutions = collectConstantSubstitutions(logic, polynomials);
-    for (PTRef key : constSubstitutions.getKeys()) {
+    auto substitutionsOrConflict = collectConstantSubstitutions(logic, polynomials);
+    if (substitutionsOrConflict.isConflict()) { return l_False; }
+    for (SubstMap const & constSubstitutions = substitutionsOrConflict.getSubstitutions();
+         PTRef key : constSubstitutions.getKeys()) {
         assert(not out_substitutions.has(key));
         out_substitutions.insert(key, constSubstitutions[key]);
     }
@@ -430,7 +459,7 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
         assert(not out_substitutions.has(var));
         out_substitutions.insert(var, sub);
     }
-    // To simplify this method, we do not try to detect a conflict here, so result is always l_Undef
+    // We could try to check the substitutions for a conflict here, but we leave that up to the caller.
     return l_Undef;
 }
 
--- a/test/unit/test_Substitutions.cc
+++ b/test/unit/test_Substitutions.cc
@@ -143,6 +143,19 @@ TEST_F(LIASubstitutionsRegression, test_TwoVarsEqual) {
     ASSERT_TRUE((key == x and value == y) or (key == y and value == x));
 }
 
+TEST_F(LIASubstitutionsRegression, test_Divisibility) {
+    auto const osmt = getLIAOsmt();
+    auto & lialogic = osmt->getLIALogic();
+    // 3x + y = 0 and y = -1
+    PTRef x = lialogic.mkIntVar("x");
+    PTRef y = lialogic.mkIntVar("y");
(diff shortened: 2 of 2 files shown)


Tomaqa:
What exactly do you mean by the first sentence, especially `run into an assertion`?

blishko:
> What exactly do you mean by the first sentence, especially `run into an assertion`?

I mean it would trigger assertion failure. I'll rephrase the description.

src/logics/ArithLogic.cc
```diff
@@ -400,11 +410,12 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
     std::transform(top_level_arith.begin(), top_level_arith.end(), std::back_inserter(polynomials),
                    [&](auto const & eq) { return ptrefToPoly(eq, logic); });
 
-    auto constSubstitutions = collectConstantSubstitutions(logic, polynomials);
+    auto [knownValue, constSubstitutions] = collectConstantSubstitutions(logic, polynomials);
     for (PTRef key : constSubstitutions.getKeys()) {
         assert(not out_substitutions.has(key));
         out_substitutions.insert(key, constSubstitutions[key]);
     }
+    if (knownValue == l_False) { return l_False; }
```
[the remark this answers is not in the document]
Tomaqa: What is the rationale behind using `lbool` here? Would `l_True` mean that we already know for sure that the instance is satisfiable? Can it ever happen here?
It seems that it may happen in `Logic::retrieveSubstitutions` but I am not sure if it still makes sense in `arithmeticElimination`.

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,12 @@ namespace {
         }
     }
 
-    Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<LAPoly> & zeroPolynomials) {
+    /// Given a system of arithmetic polynomials @p zeroPolynomials, collects substitutions of the form `v -> c`
+    /// where v is a variable and c is a constant. Applies the discovered substitutions to simplify the polynomials
+    /// in the process.
+    /// Returns early if a conflict has been detected.
+    pair<lbool, Logic::SubstMap> collectConstantSubstitutions(ArithLogic & logic,
```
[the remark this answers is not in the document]
blishko: These were good questions. The design has been change to use a dedicated type to better reflect the intended semantics.

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,31 @@ namespace {
         }
     }
 
-    Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<LAPoly> & zeroPolynomials) {
+    class SubstitutionsOrConflict {
+    public:
+        static SubstitutionsOrConflict conflict() { return SubstitutionsOrConflict(); }
```
Tomaqa, line 266:
```suggestion
        static inline const SubstitutionsOrConflict conflict{SubstitutionsOrConflict()};
```
?

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,31 @@ namespace {
...
+        static SubstitutionsOrConflict substitutions(Logic::SubstMap && subst) {
```
Tomaqa, line 267: I am not sure what is the advantage of this static function over directly using the constructor.

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,31 @@ namespace {
...
+
+        [[nodiscard]] bool isConflict() const { return not maybeSubstitutions.has_value(); }
+        [[nodiscard]] Logic::SubstMap && extractSubstitutions() {
+            assert(not isConflict());
+            return std::move(maybeSubstitutions.value());
+        }
+
+    private:
+        SubstitutionsOrConflict() : maybeSubstitutions(std::nullopt) {}
+        explicit SubstitutionsOrConflict(Logic::SubstMap && subst) : maybeSubstitutions(std::move(subst)) {}
```
Tomaqa, line 279: Why not make this one public and non-explicit?

src/logics/ArithLogic.cc
```diff
@@ -282,7 +306,8 @@ namespace {
                 if (poly.size() == 1) {
                     auto const & term = *poly.begin();
                     if (term.var == PTRef_Undef) { // FALSE equality
-                        continue;
+                        assert(not term.coeff.isZero());
+                        return SubstitutionsOrConflict::conflict();
```
Tomaqa, line 310:
```suggestion
                        return SubstitutionsOrConflict::conflict;
```

src/logics/ArithLogic.cc
```diff
@@ -327,7 +356,7 @@ namespace {
             eraseIndices(zeroPolynomials, processedIndices);
             if (not changed) { break; }
         }
-        return substitutions;
+        return SubstitutionsOrConflict::substitutions(std::move(substitutions));
```
Tomaqa, line 359:
```suggestion
        return substitutions;
```

Tomaqa: Or if we keep the constructor explicit: `return SubstitutionsOrConflict{std::move(substitutions)};`

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,31 @@ namespace {
...
+        }
+
+        [[nodiscard]] bool isConflict() const { return not maybeSubstitutions.has_value(); }
+        [[nodiscard]] Logic::SubstMap && extractSubstitutions() {
+            assert(not isConflict());
+            return std::move(maybeSubstitutions.value());
+        }
+
+    private:
+        SubstitutionsOrConflict() : maybeSubstitutions(std::nullopt) {}
```
Tomaqa, line 278:
```suggestion
        SubstitutionsOrConflict() : maybeSubstitutions{std::nullopt} {}
```

src/logics/ArithLogic.cc
Tomaqa, line 279:
```suggestion
        explicit SubstitutionsOrConflict(Logic::SubstMap && subst) : maybeSubstitutions{std::move(subst)} {}
```

src/logics/ArithLogic.cc
```diff
@@ -400,8 +429,10 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
     std::transform(top_level_arith.begin(), top_level_arith.end(), std::back_inserter(polynomials),
                    [&](auto const & eq) { return ptrefToPoly(eq, logic); });
 
-    auto constSubstitutions = collectConstantSubstitutions(logic, polynomials);
-    for (PTRef key : constSubstitutions.getKeys()) {
+    auto substitutionsOrConflict = collectConstantSubstitutions(logic, polynomials);
+    if (substitutionsOrConflict.isConflict()) { return l_False; }
+    for (SubstMap const constSubstitutions = substitutionsOrConflict.extractSubstitutions();
```
Tomaqa, line 434:
```suggestion
    for (SubstMap const constSubstitutions = std::move(substitutionsOrConflict).extractSubstitutions();
```

src/logics/ArithLogic.cc
```diff
@@ -261,7 +261,31 @@ namespace {
...
-    Logic::SubstMap collectConstantSubstitutions(ArithLogic & logic, std::vector<LAPoly> & zeroPolynomials) {
+    class SubstitutionsOrConflict {
+    public:
+        static SubstitutionsOrConflict conflict() { return SubstitutionsOrConflict(); }
+        static SubstitutionsOrConflict substitutions(Logic::SubstMap && subst) {
+            return SubstitutionsOrConflict(std::move(subst));
+        }
+
+        [[nodiscard]] bool isConflict() const { return not maybeSubstitutions.has_value(); }
+        [[nodiscard]] Logic::SubstMap && extractSubstitutions() {
```
Tomaqa, line 272:
```suggestion
        [[nodiscard]] Logic::SubstMap && extractSubstitutions() && {
```

Tomaqa: I would maybe also add `[[nodiscard]] Logic::SubstMap const & getSubstitutions() const { ... }`

src/logics/ArithLogic.cc
```diff
@@ -400,8 +429,10 @@ lbool ArithLogic::arithmeticElimination(vec<PTRef> const & top_level_arith, Subs
     std::transform(top_level_arith.begin(), top_level_arith.end(), std::back_inserter(polynomials),
                    [&](auto const & eq) { return ptrefToPoly(eq, logic); });
 
-    auto constSubstitutions = collectConstantSubstitutions(logic, polynomials);
-    for (PTRef key : constSubstitutions.getKeys()) {
+    auto substitutionsOrConflict = collectConstantSubstitutions(logic, polynomials);
+    if (substitutionsOrConflict.isConflict()) { return l_False; }
+    for (SubstMap const constSubstitutions = substitutionsOrConflict.extractSubstitutions();
+         PTRef key : constSubstitutions.getKeys()) {
```
Tomaqa, line 435:
```suggestion
         PTRef key : substitutionsOrConflict.getSubstitutions().getKeys()) {
```
Another possibility. Probably no need to destroy it?

src/logics/ArithLogic.cc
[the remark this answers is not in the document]
Tomaqa: It is relevant, but `we do not try to detect a conflict here, so result is always l_Undef` sounds to me as a general statement for the entire function.

src/logics/ArithLogic.cc
blishko: Changed to constructor.

src/logics/ArithLogic.cc
blishko: For simplicity I removed the destructive method and only kept and used `getSubstitutions`.

src/logics/ArithLogic.cc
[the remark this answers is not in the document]
blishko: Comment updated.

Tomaqa (approved):
LGTM!

PR #840, opened 2025-05-29
Fix assertion violation in split clauses

### LINKED ISSUE #838: Assertion violation on QF_ALIA instance
OpenSMT reports assertion violation on the following instance when run with random seed `-r 75`.
```
(set-logic QF_ALIA)
(declare-const x (Array Int Int))
(declare-const x5 Int)
(declare-const x54 (Array Int Int))
(declare-fun i () Int)
(declare-fun a1 () (Array Int Int))
(declare-fun ux () Int)
(declare-fun j () Int)
(declare-fun k () Int)
(declare-fun l () Int)
(declare-fun m () Int)
(declare-fun n () Int)
(declare-fun u () Int)
(assert (and (= 0 (select (store (store (store (store a1 0 ux) (+ i 2) u) (+ i 1) 0) (+ 4 i) 0) (* 2 i)))  (= 0 u) (<= 0 n) (> 0 l) (> 0 (* ux (- 1))) (= 0 (select a1 (* 2 m))) (= 0 (select x (* 2 n))) (= 0 (select (store x54 0 i) 1)) (= 0 (select (store (store a1 1 l) 9 0) 4)) (= 0 (select (store (store (store a1 0 k) 0 u) 2 0) (* 2 k))) (= 0 (select (store (store (store (store (store (store a1 i j) 1 1) x5 0) j 0) k 0) 0 0) 1))))
(check-sat)
```

This requires a bit of investigation. The assertion violation indicates that the array solver generated a single lemma, but it was satisfiable, which is weird. Also, the number of new lemmas seems quite high on this relatively simple instance.

blishko:
Previously, it could have happened that both the arithmetic solver and the array solver would produce a new lemma for the SAT solver. In our current implementation we prefer arithmetic lemmas, so if there is a new arithmetic lemma, we ignore the ones from array solver. However, we would previously not clear the stored lemmas in the array solver. So if the solver would backtrack due to the arithmetic lemma and change the assignment, the array lemma might be outdated (e.g., already satisfied). If only a single satisfied lemma is given to the code handling newly introduced lemmas, an assertion violation is triggered. The proposed solution is to clear the lemmas in the array solver when the SAT solver backtracks.

Fixes #838.


Tomaqa (approved):
LGTM!

PR #844, opened 2025-06-11
Included SMT-LIB unsat-core options into command-line options

Tomaqa:
Similarly to how option `--produce-interpolants` works, we also include SMT-LIB options `:minimal-unsat-cores` and `:print-cores-full` into the command line options.
We do not include `:produce-unsat-cores` because it does not seem to be useful (as discussed below). An alternative is suggested in #845.

### DIFF
--- a/src/bin/opensmt.cc
+++ b/src/bin/opensmt.cc
@@ -219,10 +219,12 @@ void printHelp()
           "where OPTION can be\n"
           "  --help [-h]                     prints this help message and exits\n"
           "  --version                       prints opensmt version and exits\n"
+          "  --verbose [-v]                  verbose run of the solver\n"
           "  --dry-run [-d]                  executes dry run\n"
           "  --random-seed [-r] <seed>       sets random seed to specific number\n"
+          "  --minimal-unsat-cores           produced unsat cores must be irreducible\n"
+          "  --print-cores-full              produced unsat cores are agnostic to the smt2 attribute ':named'\n"
           "  --produce-interpolants [-i]     enables interpolant computation\n"
-          "  --verbose [-v]                  verbose run of the solver\n"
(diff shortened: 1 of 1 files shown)


Tomaqa:
When I thought about it again, is it ever useful to use command line options `--produce-interpolants` or `--produce-unsat-cores`? It still does not imply calling `(get-interpolant)` or `(get-unsat-core)`, and if the SMT-LIB file already contains it, then it should also contain the `produce` option. It would make sense to me to provide an option such as `--get-unsat-core` which would enable the production and simultaneously get the core after each `(check-sat)` command. In the case of interpolants, I think it is not applicable because it needs to partition the formula.

blishko:
Not sure what is the right approach here.
The command-line option was really just a substitute for `produce-interpolants` command.
For unsat cores, `get-unsat-core` might make more sense.
The meaning has to be clearly defined. What you suggest makes sense. Basically automatically follow each `(check-sat)` command with `(get-unsat-core)`. I would add that only the case that the solver's response was `unsat`.

blishko:
Looks like the option was introduced here: 6ca1a446a
Apparently at that time it was also setting the option to produce proof, but it looks like we have changed the mechanism since.
I guess we can remove the option, but I also don't see any harm in keeping it.
It depends if we want to force users to add `produce-interpolants` to every SMT-LIB script.
It would be good to check how they call OpenSMT in JavaSMT, in order to not break anything there.

Tomaqa:
If we keep it, I find it inconsistent not to add `produce-unsat-cores` on the command line, because the argument for keeping it is exactly the same as for interpolants (not forcing users to add the option into the scripts), only in the context of unsat cores.

blishko (approved):
LGTM!

PR #852, opened 2025-07-21
Wall-clock timeout

### LINKED ISSUE #766: Supporting timeouts
Hello there.

Z3 supports setting timeout as an option.
```smt2
; time in millis
(set-option :timeout 10000)
```

Would be cool if the winning solver for `QF_LRA` and `QF_LIA` supported this as well.

NB: Congrats on the great result in SMT-COMP 2024!

Tomaqa:
It is possible to set it via API, SMT-LIB, or as a command-line parameter.
I did not implement it within the `set-option` command, hence within `SMTConfig`, because the timeout also takes an immediate action by starting a thread, which may stop the evaluation. The configuration may even be supplied by a user externally. It is also possible to specify this command after the solver was already created, which typically happens in SMT-LIB scripts. Using `SMTConfig` would require special treatment compared to other options, that is, to run an action once the option is set.

Resolves #766 

We still do not support a timeout per query for incremental solving (as e.g. CVC5 does).

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -107,8 +107,16 @@ void Interpret::setOption(ASTNode& n) {
     SMTOption value(n);
     const char* msg = "ok";
     bool rval = config.setOption(name, value, msg);
-    if (rval == false)
+    if (rval == false) {
         notify_formatted(true, "set-option failed for %s: %s", name, msg);
+    } else if (main_solver) {
+        const SMTOption& stored_value = config.getOption(name);
+        assert(not stored_value.isEmpty());
+
+        if (strcmp(name, SMTConfig::o_time_limit) == 0) {
+            main_solver->setTimeLimit(std::chrono::milliseconds{stored_value.getValue().numval});
+        }
+    }
     free(name);
 }
 
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -23,7 +23,35 @@
 #include <tsolvers/RDLTHandler.h>
 #include <unsatcores/UnsatCoreBuilder.h>
 
+#include <condition_variable>
+#include <thread>
+
 namespace opensmt {
+
+class MainSolver::TimeLimitImpl {
+public:
+    TimeLimitImpl(MainSolver & s) : solver(s) {}
+    ~TimeLimitImpl();
+
+    void setLimit(std::chrono::milliseconds);
+    void setLimitIfNotRunning(std::chrono::milliseconds);
+
+protected:
+    bool isRunning() const noexcept;
+
+    void requestEnd();
+    void waitToEnd();
+
+private:
+    MainSolver & solver;
+
+    std::jthread thread{};
+
+    // See https://en.cppreference.com/w/cpp/thread/condition_variable.html
+    std::mutex mtx{};
+    std::condition_variable condVar{};
+    bool endReq{false};
+};
 
 MainSolver::MainSolver(Logic & logic, SMTConfig & conf, std::string name)
     : theory(createTheory(logic, conf)),
@@ -56,7 +84,11 @@ MainSolver::MainSolver(std::unique_ptr<Theory> th, std::unique_ptr<TermMapper> t
     initialize();
 }
 
+MainSolver::~MainSolver() = default;
+
 void MainSolver::initialize() {
+    timeLimitImplPtr = std::make_unique<TimeLimitImpl>(*this);
+
     frames.push();
     frameTerms.push(logic.getTerm_true());
     preprocessor.initialize();
@@ -67,6 +99,11 @@ void MainSolver::initialize() {
 
     smt_solver->addOriginalSMTClause({~term_mapper->getOrCreateLit(logic.getTerm_false())}, iorefs);
     if (iorefs.first != CRef_Undef) { pmanager.addClauseClassMask(iorefs.first, 1); }
+
+    if (auto option = config.getOption(SMTConfig::o_time_limit); not option.isEmpty()) {
+        assert(option.getValue().type == O_NUM);
+        timeLimitImplPtr->setLimit(std::chrono::milliseconds{option.getValue().numval});
+    }
 }
 
 void MainSolver::push() {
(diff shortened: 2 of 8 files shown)


Tomaqa:
In the end, I implemented it as a `set-option` and not as a separate command. If the timeout is set sooner than the existence of `MainSolver`, it takes effect once the solver is created.

Tomaqa:
I moved the regression tests to a separate branch and replaced them with unit tests.
They only test `MainSolver::setTimeLimit`, though.
They, for example, do not test using `SMTConfig::o_time_limit`.

Tomaqa:
> This works even for incremental solving, but if further checks are trigerred even after the timeout, the timeout is reset again. Do we want to, instead, continue returning `unknown`? If so, the solver must remember its internal state - that it is in timeout state.

Currently, it actually keeps returning `unknown` after the first timeout. The reason is that we never reset the stop flag. We should implement `resetStop` in any case (we already have `resetGlobalStop`), but I am not sure if we want or do not want to call it at the end of each `check()` - because returning `unknown` until the user explicitly resets the timeout state (probably using an additional function for that instead of `resetStop`, like `resetTimeLimit`) seems to me as a good behavior.

I checked that CVC5 nor z3 return `unknown`, they simply terminate the whole program and output sth. like `timeout`. The timeout in Yices2 is actually time limit per query (supported by CVC5 and z3 as well). In the case of time limits per query, all of them return `unknown` if the specific query did not finish within the limit, regardless of previous results.

Tomaqa:
[CVC5](https://cvc5.github.io/docs-ci/docs-main/resource-limits.html) only supports `tlimit-per` in `(set-option)`. Using `tlimit` in there has no effect. The option is only usable if CVC5 is executed as a binary, but not as a library. They even claim that it has no effect when set via API, if not executed as a binary.

It makes sense to me. I would not include the overall timeout in the options at all, only `time-limit-per-query`.
However, we can introduce an SMT-LIB command `(set-time-limit <ms>)` which would immediately start the timer. It would use the API call `MainSolver::setTimeLimit`, as well as the command line option `--time-limit` would.

This would ideally also require checking the stop flag not only within the solving, but also within adding assertions and preprocessing.

Tomaqa:
I reverted to the previous solution where `time-limit` is executed immediately or as soon as `MainSolver` is initialized. When set via `SMTConfig`, it only takes effect implicitly when set before the `MainSolver` is initialized; otherwise, the user must call `setTimeLimit` explicitly, which is not a limitation when using the API (where `SMTConfig` is actually not necessary). `Interpret`, on the other hand, makes the call to `setTimeLimit` implicitly at the moment when the `set-option` is parsed - otherwise the user could not trigger the time limit at the desired moment.

As mentioned above, it should ideally also stop processing formulas and simplifying them once the time limit is reached, i.e., not only in the case of solving.

src/options/SMTConfig.h
```diff
@@ -146,6 +146,8 @@ namespace opensmt {
       SMTOption(ASTNode const & n);
       SMTOption() {}
       SMTOption(int i)   : value(i) {}
+      //+ Should also support long representation
+      SMTOption(long i)  : SMTOption(static_cast<int>(i)) {}
```
BritikovKI, line 150: Hmm, when do you think it will be called with long? 🤔 
And why allow it even, if you anyway would cast it to int?

src/bin/opensmt.cc
```diff
@@ -329,6 +341,18 @@ SMTConfig parseCMDLineArgs( int argc, char * argv[ ] )
...
+            case 't': {
+                int64_t timeLimit;
+                try {
+                    timeLimit = std::stoll(optarg);
+                } catch (std::logic_error const & e) {
+                    throw std::invalid_argument{"Invalid argument of time-limit: "s + e.what()};
+                }
+
+                //+ SMTOption is not defined for long long
+                ok &= res.setOption(SMTConfig::o_time_limit, SMTOption(long(timeLimit)), msg);
```
BritikovKI, line 353: Why not to convert it to int at this line instead of supporting SMTOption(long)?

BritikovKI: Also why break if ok is false? 🤔 
This is the only place when Ok is used anyway, mb it is possible to omit it? Or do you want to use it in more places eventually?

src/bin/opensmt.cc
```diff
@@ -49,7 +50,11 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 namespace opensmt {
 
-inline bool pipeExecution = false;
+using namespace std::string_literals;
+
+namespace {
+    bool pipeExecution = false;
```
BritikovKI, line 56: What's an intuition behind moving it in its own namespace?

src/api/MainSolver.h
```diff
@@ -149,6 +150,16 @@ class MainSolver {
     // For stopping at the global scope, refer to GlobalStop.h
     void notifyStop() { smt_solver->notifyStop(); }
 
+    // Set wall-clock time limit for the solver in miliseconds
+    // When it expires, the solving is terminated gracefully and unknown is returned
+    // Overrides previously set and still running limit
+    void setTimeLimit(std::chrono::milliseconds limit) { setTimeLimit(limit, {}); }
+    struct TimeLimitConf {
```
BritikovKI, line 157: Why is separate struct needed for a single bool? 🤔  Do you think it will be extended?

src/api/Interpret.cc
```diff
@@ -107,8 +107,16 @@ void Interpret::setOption(ASTNode& n) {
     SMTOption value(n);
     const char* msg = "ok";
     bool rval = config.setOption(name, value, msg);
-    if (rval == false)
+    if (rval == false) {
         notify_formatted(true, "set-option failed for %s: %s", name, msg);
+    } else if (main_solver) {
```
BritikovKI, line 112: Should it be `else if` or `else { assert(main_solver)}`? Can it be that main_solver is null?

src/api/MainSolver.cc
```diff
@@ -23,8 +23,37 @@
...
+#include <thread>
+
 namespace opensmt {
 
+class MainSolver::TimeLimitImpl {
+public:
+    TimeLimitImpl(MainSolver & s) : solver(s) {}
+    ~TimeLimitImpl();
+
+    void setLimit(std::chrono::milliseconds);
```
BritikovKI, line 36: Should it be milliseconds? 🤔 
Why not seconds, do you think this much precision is needed?

src/api/MainSolver.cc
```diff
@@ -588,4 +638,56 @@ void MainSolver::Preprocessor::addPreprocessedFormula(PTRef fla) {
...
+
+    requestEnd();
+    waitToEnd();
+}
+
+void MainSolver::TimeLimitImpl::setLimit(std::chrono::milliseconds limit) {
+    assert(limit > std::chrono::milliseconds::zero());
+
+    // Override already running thread
+    if (isRunning()) {
```
BritikovKI, line 653: Hmm, does it mean you are not updating limit, but killing current solver and starting a new one basically?

PR #860, opened 2025-08-27
Getting names from unsat cores

Tomaqa:
Allowed getting term names out of `NamedUnsatCore`s directly without the need to manually use `TermNames` for that.


blishko (dismissed):
LGTM!

blishko (approved):

Issue #855, opened 2025-08-13
Do not produce models by default


Tomaqa:
If we do not want the model, it slows down the runtime.

PR #861, opened 2025-08-28
Introduced command line options for producing models and unsat cores

### LINKED ISSUE #846: Remove command line option `--produce-interpolants [-i]`
The option does not seem necessary and may confuse why `--produce-unsat-cores` is absent (which is addressed differently in #845).
Also discussed in #844.

Tomaqa:
Consistently to `--produce-interpolants`, introduce the commands for models and unsat cores as well. These do *not* trigger `(get-*)` command after `(check-sat)`, they serve only for the SMT-LIB options (in case they were omitted).

Option `--produce-models` is currently useless since it is enabled by default. We should change this (c.f. #855).

Resolves #846 (i.e. closes it; we do not remove the option).

### DIFF
--- a/src/bin/opensmt.cc
+++ b/src/bin/opensmt.cc
@@ -51,6 +51,12 @@ namespace opensmt {
 
 inline bool pipeExecution = false;
 
+// Replace this with std::to_underlying once we move to C++23
+template<typename Enum>
+constexpr std::underlying_type_t<Enum> to_underlying(Enum e) noexcept {
+    return static_cast<std::underlying_type_t<Enum>>(e);
+}
+
 SMTConfig parseCMDLineArgs(int argc, char * argv[ ]);
 
 void        catcher            ( int );
@@ -222,6 +228,8 @@ void printHelp()
           "  --verbose [-v]                  verbose run of the solver\n"
           "  --dry-run [-d]                  executes dry run\n"
           "  --random-seed [-r] <seed>       sets random seed to specific number\n"
+          "  --produce-models [-m]           enable producing models\n"
+          "  --produce-unsat-cores           enable producing unsat cores\n"
           "  --minimal-unsat-cores           produced unsat cores must be irreducible\n"
           "  --print-cores-full              produced unsat cores are agnostic to the smt2 attribute ':named'\n"
           "  --produce-interpolants [-i]     enables interpolant computation\n"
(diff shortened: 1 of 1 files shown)


src/bin/opensmt.cc
```diff
@@ -231,21 +233,27 @@ void printHelp()
...
-    constexpr int versionLongOpt = 1;
-    constexpr int minUcoreLongOpt = 2;
-    constexpr int fullUcoreLongOpt = 3;
+    int selectedLongOpt;
 
     struct option long_options[] =
         {
             {"help", no_argument, nullptr, 'h'},
-            {"version", no_argument, &selectedLongOpt, versionLongOpt},
+            {"version", no_argument, &selectedLongOpt, int(LongOpt::version)},
```
blishko, line 249:
```suggestion
            {"version", no_argument, &selectedLongOpt, std::to_underlying(LongOpt::version)},
```

src/bin/opensmt.cc
```diff
@@ -231,21 +233,27 @@ void printHelp()
...
             {"verbose", no_argument, nullptr, 'v'},
             {"dry-run", no_argument, nullptr, 'd'},
             {"random-seed", required_argument, nullptr, 'r'},
-            {"minimal-unsat-cores", no_argument, &selectedLongOpt, minUcoreLongOpt},
-            {"print-cores-full", no_argument, &selectedLongOpt, fullUcoreLongOpt},
+            {"produce-models", no_argument, nullptr, 'm'},
+            {"produce-unsat-cores", no_argument, &selectedLongOpt, int(LongOpt::produceUcore)},
+            {"minimal-unsat-cores", no_argument, &selectedLongOpt, int(LongOpt::minUcore)},
+            {"print-cores-full", no_argument, &selectedLongOpt, int(LongOpt::fullUcore)},
```
blishko, line 256:
```suggestion
            {"produce-unsat-cores", no_argument, &selectedLongOpt, std::underlying(LongOpt::produceUcore)},
            {"minimal-unsat-cores", no_argument, &selectedLongOpt, std::underlying(LongOpt::minUcore)},
            {"print-cores-full", no_argument, &selectedLongOpt, std::underlying(LongOpt::fullUcore)},
```

blishko:
Hmm, I guess `std::to_underlying` is C++23 and we are not there yet, are we?

Tomaqa:
We are not, hence we provide a temporary copy-paste implementation for that.

PR #862, opened 2025-08-28
Options that invoke `get-*` commands after each `check-sat` gracefully

### LINKED ISSUE #845: Implement command-line option `--get-unsat-cores`
Based on the discussion in #844, implement a command line option that would gracefully call `(get-unsat-core)` after every unsatisfiable `(check-sat)`, ensuring that the option `:produce-unsat-cores` is set.

We may call it `--get-unsat-cores`.
In CVC5, they use option `--dump-unsat-cores`.

Tomaqa:
As a follow-up to #861, adding also user-friendly options for models and unsat cores that automatically ensure that the corresponding `:produce-*` option is set, and that after each `check-sat` command with the desired status, the corresponding `get-*` command is invoked, while ignoring these commands if given in the script because they have already been invoked after the `check-sat` command.

Resolves #845

### DIFF
--- a/src/api/Interpret.cc
+++ b/src/api/Interpret.cc
@@ -279,15 +279,14 @@ void Interpret::interp(ASTNode& n) {
                 break;
             }
             case t_getunsatcore: {
-                if (config.produce_unsat_cores()) {
-                    if (isInitialized()) {
-                        getUnsatCore();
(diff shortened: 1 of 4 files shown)


blishko (approved):
LGTM!

PR #864, opened 2025-10-23
Not producing models by default

### LINKED ISSUE #855: Do not produce models by default

Tomaqa:
... and fixed all tests and examples so that they enable the production of models explicitly.

This may break some applications that use OpenSMT if they assume that the option is set to true by default.

Resolves #855

### DIFF
--- a/examples/test_lra_value.cc
+++ b/examples/test_lra_value.cc
@@ -7,6 +7,7 @@ Opensmt*
 pre()
 {
     auto config = std::unique_ptr<SMTConfig>(new SMTConfig());
+    config->setProduceModels();
     const char* msg;
     config->setOption(SMTConfig::o_produce_inter, SMTOption(true), msg);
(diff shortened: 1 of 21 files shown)


examples/test_lra_value.cc
```diff
@@ -8,6 +8,7 @@ pre()
 {
     auto config = std::unique_ptr<SMTConfig>(new SMTConfig());
     const char* msg;
+    config->setOption(SMTConfig::o_produce_models, SMTOption(true), msg);
```
blishko: Can we use `config->setProduceModels()` here as well?
It seems simpler than this.

test/regression/base/model/issue123.smt2
```diff
@@ -1,4 +1,4 @@
-(set-option :produce-model 1)
+(set-option :produce-models 1)
```
blishko:
```suggestion
(set-option :produce-models true)
```

For consistency.

blishko (approved):
LGMT!

PR #869, opened 2025-10-24
Refactored string (print) functions in `Logic` and clarified the use of SMT-LIB2 format

### LINKED ISSUE #789: Clarify the difference between `pp` and `printTerm`
`printTerm` seems to be SMT-LIB compliant, while `pp` does not.
May be connected to the way how `FastRational`s are printed out or parsed.
Investigate the reasons behind `pp` and consider removing it from the public API.

Tomaqa:
Clarified that the former `printTerm` refers to the SMT-LIB2 format in the first place.
In addition, I avoided to use the name `print*` and prefer `*ToString` instead, because the result is `std::string`.

I kept `Logic::pp` because it is used at several parts, and is not necessarily a bad thing to have - one does not always want an output in SMT-LIB format.

Resolves #789

### DIFF
--- a/examples/test_itp.cc
+++ b/examples/test_itp.cc
@@ -70,7 +70,7 @@ int main()
 
         std::vector<PTRef> itps;
         itp_context->getSingleInterpolant(itps, mask);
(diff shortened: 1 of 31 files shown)


blishko (approved):
LGTM!

PR #871, opened 2025-10-29
Extended `ScopedVector` of `clear` and of accessing particular scopes

Tomaqa:
Extended `ScopedVector` to allow more operations and look more std. container-like.

Importantly, I added functions that allow accessing particular scopes, not only the entire vector, and not even necessarily the top scope. This allows traversing the elements in batches in a lazy fashion, as for example done with frames in the preprocessing pipeline.

Btw., `AssertionStack` could probably also be rebased on top of a `ScopedVector`, but it would not be that straightforward.

### DIFF
--- a/src/common/ScopedVector.h
+++ b/src/common/ScopedVector.h
@@ -51,6 +51,28 @@ public:
     [[nodiscard]] T * data() { return elements.data(); }
 
     [[nodiscard]] std::size_t scopeCount() const { return limits.size() + 1; }
+
+    [[nodiscard]] std::size_t topScopeEmpty() const { return topScopeSize() == 0; }
+    [[nodiscard]] std::size_t topScopeSize() const { return size() - sizeBeforeTopScope(); }
+
+    [[nodiscard]] auto topScopeBegin() const { return begin() + sizeBeforeTopScope(); }
+    [[nodiscard]] auto topScopeEnd() const { return end(); }
+
+    [[nodiscard]] auto topScopeBegin() { return begin() + sizeBeforeTopScope(); }
+    [[nodiscard]] auto topScopeEnd() { return end(); }
+
+    [[nodiscard]] T const * topScopeData() const { return data() + sizeBeforeTopScope(); }
+    [[nodiscard]] T * topScopeData() { return data() + sizeBeforeTopScope(); }
+
+    // Views to just the top scope
+    inline auto topScope() const;
+    inline auto topScope();
+
+protected:
+    template<typename>
+    class TopScopeView;
+
+    inline std::size_t sizeBeforeTopScope() const;
 };
 
 template<typename T>
@@ -75,6 +97,44 @@ template<typename T>
 void ScopedVector<T>::clear() {
     elements.clear();
     limits.clear();
+}
+
+template<typename T>
+std::size_t ScopedVector<T>::sizeBeforeTopScope() const {
+    if (limits.empty()) { return 0; }
+    return limits.back();
+}
+
+template<typename T>
+template<typename VectorT>
+class ScopedVector<T>::TopScopeView {
+public:
+    explicit TopScopeView(VectorT & scopedVector_) : scopedVector{scopedVector_} {}
+
+    bool empty() const noexcept { return scopedVector.topScopeEmpty(); }
+    std::size_t size() const noexcept { return scopedVector.topScopeSize(); }
+
+    auto begin() const noexcept { return scopedVector.topScopeBegin(); }
+    auto end() const noexcept { return scopedVector.topScopeEnd(); }
+
+    auto begin() noexcept { return scopedVector.topScopeBegin(); }
+    auto end() noexcept { return scopedVector.topScopeEnd(); }
+
+    T const * data() const noexcept { return scopedVector.topScopeData(); }
+    T * data() noexcept { return scopedVector.topScopeData(); }
+
+protected:
+    VectorT & scopedVector;
+};
+
+template<typename T>
+auto ScopedVector<T>::topScope() const {
+    return TopScopeView<ScopedVector const>{*this};
+}
+
+template<typename T>
+auto ScopedVector<T>::topScope() {
+    return TopScopeView<ScopedVector>{*this};
 }
 } // namespace opensmt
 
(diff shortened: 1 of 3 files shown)


Tomaqa:
#873 currently requires this PR + #872

Tomaqa:
Update: #873 requires clearing the whole scope vector and a const-access of any individual scope.

src/common/ScopedVector.h
```diff
@@ -18,29 +20,80 @@ class ScopedVector {
     std::vector<unsigned> limits;
 
 public:
+    using value_type = T;
```
BritikovKI, line 23: You introduced value_type, ..., const_pointer, but they are never used 🤔

src/common/ScopedVector.h
```diff
@@ -18,29 +20,80 @@ class ScopedVector {
...
+    [[nodiscard]] auto scopeEnd(std::size_t idx) const;
+
+    [[nodiscard]] T const * scopeData(std::size_t idx) const { return data() + sizeBeforeScope(idx); }
+
+    // Views to just the top scope
+    inline auto topScope() const;
+    inline auto topScope();
+
+    // Views to just a scope
+    inline auto scope(std::size_t idx) const;
```
BritikovKI, line 84: Why only const here?

src/common/ScopedVector.h
```diff
@@ -60,6 +113,104 @@ void ScopedVector<T>::popScope(TFun callback) {
...
+}
+
+template<typename T>
+bool ScopedVector<T>::isValidScopeIdx(std::size_t idx) const {
+    return idx < scopeCount();
+}
+
+template<typename T>
+void ScopedVector<T>::checkScopeIdx(std::size_t idx) const {
+    if (isValidScopeIdx(idx)) { return; }
```
BritikovKI: Why this function is void? 🤔 
Usually `checkSmth` function would return bool
Because this function will just throw an error or do nothing...
Also why is it not used in all of the other funcs, like: `scopeSize`, `scopeEnd`, etc.

src/common/ScopedVector.h
```diff
@@ -60,6 +113,104 @@ void ScopedVector<T>::popScope(TFun callback) {
...
+    return begin() + sizeBeforeScope(idx + 1);
+}
+
+template<typename T>
+template<typename VectorT>
+class ScopedVector<T>::TopScopeView {
+public:
+    explicit TopScopeView(VectorT & scopedVector_) : scopedVector{scopedVector_} {}
+
+    bool empty() const noexcept { return scopedVector.topScopeEmpty(); }
```
BritikovKI, line 159: Why const?

test/unit/test_ScopedVector.cc
```diff
@@ -0,0 +1,218 @@
...
+            if (sve != ve) { return false; }
+        }
+
+        return true;
+    }
+
+    ScopedVector<int> svector;
+};
+
+TEST_F(Test, test_ScopeVector1) {
```
BritikovKI, line 55: Can you mb add some comments/documentation about what are those tests doing?

BritikovKI (commented):
Just a general question what are the calls doing?

src/common/ScopedVector.h
Tomaqa: These types are required by the standard if one wants to use the container in a range for loop. Without them, you cannot do e.g. `for (auto & elem : scopeVector)`.

src/common/ScopedVector.h
Tomaqa: I am currently being conservative. The main purpose of a scoped vector is to operate on it similarly to a stack, modifying only the top contents. Sometimes, one needs to read specific previous scopes, but I currently do not have a use case that would require modifying them. This way, the class guarantees that the contents of a scope that is not at the top level are immutable.

src/common/ScopedVector.h
Tomaqa: `isValidScopeIdx` is `bool`, `checkScopeIdx` additionally throws an exception.

> Because this function will just throw an error or do nothing...

This is exactly the intended behavior :) I am not sure what the issue is here, could you elaborate?

All such functions call `checkScopeIdx` at some point, sometimes via `sizeBeforeScope`.

src/common/ScopedVector.h
Tomaqa: It is just a getter, similarly to STL containers

Tomaqa:
> Just a general question what are the calls doing?

Which calls?

test/unit/test_ScopedVector.cc
Tomaqa: I added a short comment

src/common/ScopedVector.h
BritikovKI: Gotcha!

src/common/ScopedVector.h
BritikovKI: Fair enough,  my concerns are mainly adressed... 

The only place `isValidScopeIdx` is used is inside `checkScopeIdx`... Should both those checks be public? 🤔 Do you think they both would be used outside??

About `sizeBeforeScope` - fair enough, should not it be called in `ScopeEmpty` though? Because you access

test/unit/test_ScopedVector.cc
BritikovKI: Thanks, it helps to give a bit more context to the tests

src/common/ScopedVector.h
Tomaqa: 1) Both are `protected` and they should stay that way. I consider it a good practice to separate `isValid*` and `check*` which allows a potential use of `isValid*` in the future. However, in this case, I agree that `isValidScopeIdx` is droppable.

2) `scopeEmpty` -> `scopeSize` -> `sizeBeforeScope` -> `checkScopeIdx`

Tomaqa: I inlined `isValidScopeIdx` to `idx < scopeCount()`, it also makes it clearer that it is similar to other checks such as `idx == scopeCount() - 1`.

BritikovKI (dismissed):
Lgtmt!!!

As a small comment, can you add tests, when compare/compareScope returns false?
Because right now you only test cases when it works correctly - this is not too important however

Tomaqa:
Good point, I added a few `ASSERT_FALSE` cases.

BritikovKI (approved):
Lgtmt

PR #881, opened 2025-11-13
Partial fix of `get-assignment`: `produce-assignments` and tracking partitions

### LINKED ISSUE #876: `get-assignment` is not working properly
`get-assignment` sometimes returns `unknown`, and we even use such cases in regression tests, e.g. in `test/regression/base/incremental/assignment_global.smt2`.
The test does *not* output `unknown` if `trackPartitions()` evaluates to true (i.e., with `produce-proofs`).

It uses `MainSolver::getTermValue`, and it seems that `term_mapper->hasLit(tr)` is not working properly here - it claims that the xor term is not present in `TermMapper`, hence returning `unknown`.

Tomaqa:
Added `:produce-assignments` option, which implies tracking partitions, needed to provide values for named top-level formulas.

Only partially resolves #876.
In `QF_LRA/constraints-cooking01-edit.smt2`, one of the values remains `unknown`.

### DIFF
--- a/test/regression/base/QF_LRA/constraints-cooking01-edit.smt2
+++ b/test/regression/base/QF_LRA/constraints-cooking01-edit.smt2
@@ -1,4 +1,5 @@
 (set-logic QF_LRA)
+(set-option :produce-assignments true)
 (set-info :source |
 TLP-GP automated DTP to SMT-LIB encoding for planning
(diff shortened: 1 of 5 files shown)


Tomaqa:
> tracking partitions might be an overkill, when what we really want is tracking the named terms.

Agreed -> #885

PR #888, opened 2025-11-18
Split `simplifyFormulas` to sub-functions

Tomaqa:
This is just a refactoring. The behavior or performance should not change.

This makes the preprocessing more maintainable, allows overriding particular parts, and also allows preprocessing single formulas without necessarily giving them to the SMT solver with `giveToSolver`.

Also added `preprocessAssertions` as an alias to `simplifyFormulas`.

### DIFF
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -141,23 +141,11 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
 sstat MainSolver::preprocessAssertions() {
     status = s_Undef;
     for (std::size_t i = firstNotPreprocessedFrame; i < frames.frameCount(); ++i) {
-        auto & frame = frames[i];
-        FrameId const frameId = frame.getId();
-        PreprocessingContext context{.frameCount = i, .perPartition = trackPartitions()};
-        preprocessor.prepareForProcessingFrame(i);
-        firstNotPreprocessedFrame = i + 1;
-
-        if (not context.perPartition) {
-            PTRef frameFormula = preprocessFormulasConjoined(frame.formulas, context);
-            status = giveToSolver(frameFormula, frameId);
-        } else {
-            vec<PTRef> processedFormulas = preprocessFormulasPerPartition(frame.formulas, context);
-            for (PTRef fla : processedFormulas) {
-                status = giveToSolver(fla, frameId);
-                if (status == s_False) { break; }
-            }
+        assert(status != s_False);
+        if (not tryPreprocessFrame(i)) {
+            assert(status == s_False);
+            break;
         }
-        if (status == s_False) { break; }
     }
 
     if (status == s_False) {
@@ -166,6 +154,31 @@ sstat MainSolver::preprocessAssertions() {
     }
 
     return status;
+}
+
+bool MainSolver::tryPreprocessFrame(std::size_t i) {
+    auto & frame = frames[i];
+    FrameId const frameId = frame.getId();
+    PreprocessingContext context{.frameCount = i, .perPartition = trackPartitions()};
+    preprocessor.prepareForProcessingFrame(i);
+    firstNotPreprocessedFrame = i + 1;
+
+    assert(status != s_False);
+
+    if (not context.perPartition) {
+        PTRef frameFormula = preprocessFormulasConjoined(frame.formulas, context);
+        status = giveToSolver(frameFormula, frameId);
+        return status != s_False;
+    }
+
+    vec<PTRef> processedFormulas = preprocessFormulasPerPartition(frame.formulas, context);
+    for (PTRef fla : processedFormulas) {
+        status = giveToSolver(fla, frameId);
+        if (status == s_False) { return false; }
+    }
+
+    assert(status != s_False);
+    return true;
 }
 
 PTRef MainSolver::preprocessFormulasConjoined(vec<PTRef> const & flas, PreprocessingContext const & context) {
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -138,63 +138,124 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
     return termNames.tryInsert(name, fla);
 }
 
-sstat MainSolver::simplifyFormulas() {
+sstat MainSolver::preprocessAssertions() {
     status = s_Undef;
-    for (std::size_t i = firstNotSimplifiedFrame; i < frames.frameCount() && status != s_False; i++) {
+    for (std::size_t i = firstNotSimplifiedFrame; i < frames.frameCount(); ++i) {
+        auto & frame = frames[i];
+        FrameId const frameId = frame.getId();
         PreprocessingContext context{.frameCount = i, .perPartition = trackPartitions()};
         preprocessor.prepareForProcessingFrame(i);
         firstNotSimplifiedFrame = i + 1;
-        if (context.perPartition) {
-            vec<PTRef> frameFormulas;
-            for (PTRef fla : frames[i].formulas) {
-                PTRef processed = theory->preprocessAfterSubstitutions(fla, context);
-                pmanager.transferPartitionMembership(fla, processed);
-                frameFormulas.push(processed);
-                preprocessor.addPreprocessedFormula(processed);
-            }
-            if (frameFormulas.size() == 0 or std::all_of(frameFormulas.begin(), frameFormulas.end(),
-                                                         [&](PTRef fla) { return fla == logic.getTerm_true(); })) {
-                continue;
-            }
-            theory->afterPreprocessing(preprocessor.getPreprocessedFormulas());
-            for (PTRef fla : frameFormulas) {
-                if (fla == logic.getTerm_true()) { continue; }
-                assert(pmanager.getPartitionIndex(fla) != -1);
-                // Optimize the dag for cnfization
-                if (logic.isBooleanOperator(fla)) {
-                    PTRef old = fla;
-                    fla = rewriteMaxArity(fla);
-                    pmanager.transferPartitionMembership(old, fla);
-                }
-                assert(pmanager.getPartitionIndex(fla) != -1);
-                pmanager.propagatePartitionMask(fla);
-                status = giveToSolver(fla, frames[i].getId());
+
+        if (not context.perPartition) {
+            PTRef frameFormula = preprocessFormulasConjoined(frame.formulas, context);
+            status = giveToSolver(frameFormula, frameId);
+        } else {
+            vec<PTRef> processedFormulas = preprocessFormulasPerPartition(frame.formulas, context);
+            for (PTRef fla : processedFormulas) {
+                status = giveToSolver(fla, frameId);
                 if (status == s_False) { break; }
             }
-        } else {
-            PTRef frameFormula = logic.mkAnd(frames[i].formulas);
-            if (context.frameCount > 0) { frameFormula = applyLearntSubstitutions(frameFormula); }
-            frameFormula = theory->preprocessBeforeSubstitutions(frameFormula, context);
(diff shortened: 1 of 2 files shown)


src/api/MainSolver.cc
```diff
@@ -140,63 +140,138 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+    } else if (logic.isFalse(processed)) {
+        return logic.getTerm_false();
+    }
+
+    preprocessor.addPreprocessedFormula(processed);
+
+    return processed;
+}
+
+void MainSolver::preprocessFormulaDoFinalTheoryPreprocessing(PreprocessingContext const &) {
```
BritikovKI: I don't think this function makes much sense...
It is just a function which calls one function, why not use original function call???

Tomaqa: This way, it is clearer that this part must not be omitted. Without the auxiliary function, it could be missed. Besides, connecting the `theory` and `preprocessor` is not completely useless, prevents code duplication.

src/api/MainSolver.cc
```diff
@@ -140,63 +140,138 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+    for (PTRef fla : processedFormulas) {
+        status = giveToSolver(fla, frameId);
+        if (status == s_False) { return false; }
+    }
+
+    assert(status != s_False);
+    return true;
+}
+
+PTRef MainSolver::preprocessFormulasDefault(vec<PTRef> const & frameFormulas, PreprocessingContext const & context) {
```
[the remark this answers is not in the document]
Tomaqa: What is wrong with a 3-line function? It also helps readability. Before I did this, I struggled to see that `not context.perPartition` is the usual (default) case, while `context.perPartition` is a special case.
It also clearly and consistently separates the part of the transformation of the formula, and the `giveToSolver` part.

src/api/MainSolver.cc
```diff
@@ -140,63 +140,138 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+}
+
+PTRef MainSolver::preprocessFormula(PTRef fla, PreprocessingContext const & context) {
+    PTRef processed = preprocessFormulaBeforeFinalTheoryPreprocessing(fla, context);
+    preprocessFormulaDoFinalTheoryPreprocessing(context);
+    processed = preprocessFormulaAfterFinalTheoryPreprocessing(processed, context);
+    return processed;
+}
+
+PTRef MainSolver::preprocessFormulaBeforeFinalTheoryPreprocessing(PTRef fla, PreprocessingContext const & context) {
```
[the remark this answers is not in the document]
BritikovKI: I like your idea here! 
`Middle` looks a little bit off, but I don't have a better idea for the name
There is also an option to ask LLM for the recommendation))

BritikovKI:
> I need to use `preprocessFormula` later as a separate function in a different PR (#873). Maybe encapsulating all related stuff into a class (e.g. `PreprocessFormula`) would help readability?

I think it actually can improve readability, especially if separated and split into private/public, to see which functions will be used outside and which are purely internal...

src/api/MainSolver.cc
```diff
@@ -316,11 +391,15 @@ sstat MainSolver::giveToSolver(PTRef root, FrameId push_id) {
...
         void operator()(vec<Lit> && c) override { clauses.push_back(std::move(c)); }
     };
+
+    if (root == logic.getTerm_true()) { return s_Undef; }
+
     ClauseCallBack callBack;
     ts.setClauseCallBack(&callBack);
     ts.Cnfizer::cnfize(root, push_id);
     bool const keepPartitionsSeparate = trackPartitions();
-    Lit frameLit = push_id == 0 ? Lit{} : term_mapper->getOrCreateLit(frameTerms[push_id]);
```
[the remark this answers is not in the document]
Tomaqa: I tried it out but to conform to clang-format, it ended up with:
```
[[maybe_unused]]
Lit frameLit = [&] { return (push_id != 0) ? term_mapper->getOrCreateLit(frameTerms[push_id]) : Lit{}; }();
```
It does not look better to me.

I just added `[[maybe_unused]]`.

src/api/MainSolver.cc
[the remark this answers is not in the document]
Tomaqa: After discussing with Gemini, I ended up with the suffix `GlobalPhase` that suggests that the second step is different than step 1 and 3, which both may be applied to individual formulas.

src/api/MainSolver.cc
Tomaqa: See below. It tries to make a hopefully clear scheme where all three steps are necessary and virtual, so you are also allowed to override them.

Tomaqa:
> I need to use `preprocessFormula` later as a separate function in a different PR (#873). Maybe encapsulating all related stuff into a class (e.g. `PreprocessFormula`) would help readability?

I decided not to, because there is already a class `Preprocessor` that is being used under the hood. I am afraid that introducing another related class would make it confusing. Merging everything into `Preprocessor` also does not seem to be a good idea to me.

IMO, the naming scheme where the individual steps of `preprocessFormula` use just suffixes makes it clear that it is an implementation detail. It should still not be `private` to allow the use of them if one overrides the functions.

src/api/MainSolver.h
```diff
@@ -104,7 +104,9 @@ class MainSolver {
     sstat solve();
     // Simplify formulas until all are simplified or the instance is detected unsat
     // Skip assertion levels that have already been simplified
-    sstat simplifyFormulas();
+    sstat preprocessAssertions();
+    // Alias for `preprocessAssertions` for backward compatibility
+    sstat simplifyFormulas() { return preprocessAssertions(); }
```
BritikovKI, line 108: **Not important comment**
Maybe add a message that it is deprecated, and the users should switch to "preprocessAssertions"

src/api/MainSolver.h
```diff
@@ -89,7 +89,7 @@ class MainSolver {
     void insertFormula(PTRef fla);
     // Alias for `insertFormula`, reserved for future use
     void addAssertion(PTRef fla) { return insertFormula(fla); }
-    std::size_t getInsertedFormulasCount() const { return insertedFormulasCount; }
+    std::size_t getInsertedFormulasCount() const { return insertedAssertionsCount; }
```
BritikovKI: Mb rename to "getInsertedAssertionsCount"?
(While maintaining deprecated getter)

src/api/MainSolver.cc
```diff
@@ -138,65 +138,140 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+    auto & frame = frames[i];
+    FrameId const frameId = frame.getId();
+    PreprocessingContext context{.frameCount = i, .perPartition = trackPartitions()};
+    preprocessor.prepareForProcessingFrame(i);
+    firstNotPreprocessedFrame = i + 1;
+
+    assert(status != s_False);
+
+    if (not context.perPartition) {
+        PTRef frameFormula = preprocessFormulasDefault(frame.formulas, context);
```
BritikovKI: Should those functions be renamed to `preprocessAssertionsDefault`, `preprocessAssertionsPerPartition`?

src/api/MainSolver.cc
```diff
@@ -138,65 +138,140 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+        status = giveToSolver(frameFormula, frameId);
+        return status != s_False;
+    }
+
+    vec<PTRef> processedFormulas = preprocessFormulasPerPartition(frame.formulas, context);
```
BritikovKI, line 174: **Not important comment**
imho putting it into `else` block is nicer structurally, but not that important, it's the same code....

src/api/MainSolver.cc
```diff
@@ -138,65 +138,140 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+    for (PTRef fla : processedFormulas) {
+        status = giveToSolver(fla, frameId);
+        if (status == s_False) { return false; }
+    }
+
+    assert(status != s_False);
+    return true;
+}
+
+PTRef MainSolver::preprocessFormulasDefault(vec<PTRef> const & flas, PreprocessingContext const & context) {
```
BritikovKI: **Not important comment**
I think it's Ok to not make a separate function for it, because I cannot imagine other cases when this function can be useful, but it's ok

src/api/MainSolver.cc
```diff
@@ -138,65 +138,140 @@ bool MainSolver::tryAddTermNameFor(PTRef fla, std::string const & name) {
...
+    if (flas.size() == 0) { return {}; }
+
+    vec<PTRef> processedFormulas;
+    for (PTRef fla : flas) {
+        PTRef processed = preprocessFormulaBeforeGlobalPhase(fla, context);
+        processedFormulas.push(processed);
+    }
+
+    assert(processedFormulas.size() == flas.size());
+    assert(processedFormulas.size() > 0);
```
BritikovKI: ***Not important comment***
This assert is not needed, it is impossible because of the line 196 and automatically asserted by the line 204

src/api/MainSolver.cc
BritikovKI: And why have 2 functions, cannot you just have
`preprocessFormulasPerPartition({logic.mkAnd(frame.formulas)}, context)` instead of `preprocessFormulasByDefault`?

Because it seems that the logic is just duplicated
It will allow to remove `preprocessFormulasByDefault`, `preprocessFormula` functions

src/api/MainSolver.h
Tomaqa: I see that even `insertFormula` and `getInsertedFormulasCount` are still not deprecated, and instead the alternatives are "reserved for future use". I will consistently do the same here. In the future, the old methods should be marked deprecated together.

src/api/MainSolver.h
Tomaqa: I renamed `getAssertionsCount` to `getAddedAssertionsCount` and also added a description. It was misleading, it does not return the number of current assertions, but actually the number of calls to `addAssertion`.

I also renamed `insertedAssertionsCount` to `addedAssertionsCount` to make it consistent with the getter.

src/api/MainSolver.cc
Tomaqa: > Should those functions be renamed to `preprocessAssertionsDefault`, `preprocessAssertionsPerPartition`?

No, because it depends on a formulas argument, which are not necessarily the top-level assertions.

Tomaqa: > And why have 2 functions, cannot you just have `preprocessFormulasPerPartition({logic.mkAnd(frame.formulas)}, context)` instead of `preprocessFormulasByDefault`?
> 
> Because it seems that the logic is just duplicated It will allow to remove `preprocessFormulasByDefault`, `preprocessFormula` functions

I understand that it now seems very similar, but both `preprocessFormulasDefault` and `preprocessFormula` are necessary for the upcoming PRs :) To say the least, `preprocessFormulasDefault` saves a dynamic allocation by not using `vec`.

Tomaqa: Anyway, I decided to rename `preprocessFormulasDefault` to `preprocessFormulasConjoined` so that the name is more informative.

src/api/MainSolver.cc
Tomaqa: I think it is better that the else branch is less nested this way.

src/api/MainSolver.cc
Tomaqa: Addressed above

src/api/MainSolver.cc
Tomaqa: Removed

BritikovKI (approved):
Lgtmt

PR #889, opened 2025-11-24
Added counters of already preprocessed assertions

Tomaqa:
Only matters in incremental solving when it may happen that some assertions are preprocessed multiple times.

If `not trackPartitions()`, then we do not use the counters and always preprocess all frame formulas again in the hope that it will benefit from the new assertions.

If `trackPartitions()`, then we use the counters to ensure that every formula is preprocessed at most once.

Requires #888

### DIFF
--- a/src/api/MainSolver.cc
+++ b/src/api/MainSolver.cc
@@ -23,6 +23,8 @@
 #include <tsolvers/RDLTHandler.h>
 #include <unsatcores/UnsatCoreBuilder.h>
 
+#include <type_traits>
+
 namespace opensmt {
 
 MainSolver::MainSolver(Logic & logic, SMTConfig & conf, std::string name)
@@ -60,6 +62,7 @@ void MainSolver::initialize() {
     frames.push();
     frameTerms.push(logic.getTerm_true());
     preprocessor.initialize();
+    preprocessedAssertionsCountPerFrame.push_back(0);
     smt_solver->initialize();
     pair<CRef, CRef> iorefs{CRef_Undef, CRef_Undef};
     smt_solver->addOriginalSMTClause({term_mapper->getOrCreateLit(logic.getTerm_true())}, iorefs);
@@ -74,6 +77,7 @@ void MainSolver::push() {
     frames.push();
     preprocessor.push();
     frameTerms.push(newFrameTerm(frames.last().getId()));
+    preprocessedAssertionsCountPerFrame.push_back(0);
     termNames.pushScope();
     if (alreadyUnsat) { rememberLastFrameUnsat(); }
 }
@@ -92,7 +96,9 @@ bool MainSolver::pop() {
     }
     frames.pop();
     preprocessor.pop();
+    preprocessedAssertionsCountPerFrame.pop_back();
     termNames.popScope();
+    // goes back to frames.frameCount()-1 only if a formula is added via addAssertion
     firstNotPreprocessedFrame = std::min(firstNotPreprocessedFrame, frames.frameCount());
     if (not isLastFrameUnsat()) { getSMTSolver().restoreOK(); }
     return true;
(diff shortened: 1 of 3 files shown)


Tomaqa:
As done in #888, I add a performance comparison in `QF_LRA` with the 2-minute timeout. This time I always exclude unsat cores and only include `(non-)incremental`.

Comparison before and after applying this PR: 
[2025-v2.9.2-23-g883d2889_v2.9.2-24-ga646c8af.scatter.pdf](https://github.com/user-attachments/files/23713594/2025-v2.9.2-23-g883d2889_v2.9.2-24-ga646c8af.scatter.pdf)
.. confirms that when not tracking partitions, there is no effect.

Comparison before and after applying this PR when `trackPartitions()` is forced to true: 
[2025-v2.9.2-23-g883d2889-dirty_v2.9.2-24-ga646c8af-dirty.scatter.pdf](https://github.com/user-attachments/files/23713618/2025-v2.9.2-23-g883d2889-dirty_v2.9.2-24-ga646c8af-dirty.scatter.pdf)
.. shows a beneficial trend, especially in the case of unsat benchmarks.

Comparison of the new version with and without tracking partitions: 
[2025-v2.9.2-24-ga646c8af_v2.9.2-24-ga646c8af-dirty.scatter.pdf](https://github.com/user-attachments/files/23713657/2025-v2.9.2-24-ga646c8af_v2.9.2-24-ga646c8af-dirty.scatter.pdf)
.. shows that in the case of unsat benchmarks, the overhead of the tracking is quite low.

Tomaqa:
I did exhaustive SMT-LIB benchmarking and unfortunately observed some counter-intuitive results. I need to investigate.

Tomaqa:
I did my best to make the run of the experiments as stable as possible at the HPC, and now the results comparing the version before and after applying this PR seem fine.

Below, I only experimented with `non-incremental` benchmarks on the CSCS cluster with 20-min timeout.
For `incremental` benchmarks, I need to update the plot script to take into account the number of solved queries. Currently, it only distinguishes whether any has been solved within the time limit.
I think that the earlier experiments above include incremental benchmarks as well, which likely gave some misleading results.

If `not trackPartitions()`, the results are almost the same:
[2025-_non-incremental_QF_LRA_-master_preprocess-counters.scatter.pdf](https://github.com/user-attachments/files/31369207/2025-_non-incremental_QF_LRA_-master_preprocess-counters.scatter.pdf)
[2025-_non-incremental_QF_LIA_-master_preprocess-counters.scatter.pdf](https://github.com/user-attachments/files/31369214/2025-_non-incremental_QF_LIA_-master_preprocess-counters.scatter.pdf)
[2025-_non-incremental_QF_UF-master_preprocess-counters.scatter.pdf](https://github.com/user-attachments/files/31369221/2025-_non-incremental_QF_UF-master_preprocess-counters.scatter.pdf)

If `trackPartitions()` is forced to `true`, the results are also almost the same:
[2025-_non-incremental_QF_LRA_-master-track_preprocess-counters-track.scatter.pdf](https://github.com/user-attachments/files/31369257/2025-_non-incremental_QF_LRA_-master-track_preprocess-counters-track.scatter.pdf)
[2025-_non-incremental_QF_LIA_-master-track_preprocess-counters-track.scatter.pdf](https://github.com/user-attachments/files/31369386/2025-_non-incremental_QF_LIA_-master-track_preprocess-counters-track.scatter.pdf)
[2025-_non-incremental_QF_UF-master-track_preprocess-counters-track.scatter.pdf](https://github.com/user-attachments/files/31369391/2025-_non-incremental_QF_UF-master-track_preprocess-counters-track.scatter.pdf)
.. which is expected, because the feature mostly targets incremental situations when a frame is popped, more constraints are added, and then the preprocessor should only preprocess the newly introduced assertions.

So this sanity check passed and now we also need to make sure that it actually helps in incremental benchmarks.

BritikovKI:
Hey @Tomaqa, is this PR ready for the review or not yet?

PR #896, opened 2026-06-22
Preferring brace initialization of `SMTOption`

Tomaqa:
And avoiding narrowig conversions


Tomaqa:
Does *NOT* resolve #897, it only allows the use of `SMTOption` with `long long` but on platforms where `long` is 32-bit, it would misbehave with values that would not fit into 32 bits.

Issue #893, opened 2026-02-19
Invalid interpolant computed when a mod operation is present in both partitions


blishko:
For the following problem OpenSMT computes incorrect invariant:
```
(set-option :produce-interpolants 1)
;(set-option :certify-interpolants 1)
(set-logic QF_LIA)
(declare-fun x () Int)
(assert (!
(< (mod x 10) 5)
:named a))


(assert (!
(> (mod x 10) 6)
:named b))
(check-sat)
(get-interpolants a b)
(exit)
```

Interpolant computed:
```sh
((<= (- 4) (* (- 1) .mod_24_27)))
```

The problem is that OpenSMT keeps the internal auxiliary variable in the interpolant.

Suggested solution:
Add a post processing step to clean up the interpolant and replace the internal auxiliary variable with the term it represents.

PR #900, opened 2026-07-22
Interpol mod fix

BritikovKI:
This pull request is supposed to fix issue #893

### DIFF
--- a/src/proof/InterpolationContext.cc
+++ b/src/proof/InterpolationContext.cc
@@ -674,7 +674,6 @@ PTRef SingleInterpolationComputationContext::computePartialInterpolantForTheoryC
     std::vector<Lit> const & oldvec = n.getClause();
     for (Lit l : oldvec) {
         newvec.push(~l);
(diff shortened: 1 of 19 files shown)


BritikovKI:
This pull request parses the produced interpolant (Linear Integer Arithmetic particularly) and replaces introduced div/mod variables with exact div/mod representations

Tomaqa:
Could you add a unit/regression test such as the one in #893? Ideally, it should fail before the fix and pass after.

BritikovKI:
> Could you add a unit/regression test such as the one in #893? Ideally, it should fail before the fix and pass after.

Oh, sorry, my bad, I had a commit locally but forgot to push it...

Tomaqa:
The test does not fail before the fix. I also tried the Debug mode, but I assume the test is not supposed to only trigger or not trigger a debug assertion failure.

BritikovKI:
My bad, now it should work!

